PR #14, opened 2015-01-13
Changed integer suffix from 'u' to 'us'

jasonthompson:
I was really happy to find your work on cpal, as I'm interested in using rust for audio programming.

I changed the suffixes on ints in alsa from 'u' to 'us' as per compiler output in Rust alpha. 

Jason


tomaka:
Thanks!

PR #23, opened 2015-02-22
Update for rustc

tomaka:


src/alsa/mod.rs
```diff
@@ -19,7 +19,7 @@ impl Voice {
             let name = ffi::CString::from_slice(b"default");
 
             let mut playback_handle = mem::uninitialized();
-            check_errors(alsa::snd_pcm_open(&mut playback_handle, name.as_slice_with_nul().as_ptr(), alsa::SND_PCM_STREAM_PLAYBACK, alsa::SND_PCM_NONBLOCK)).unwrap();
+            check_errors(alsa::snd_pcm_open(&mut playback_handle, name.as_bytes_with_nul().as_ptr() as *const libc::c_char, alsa::SND_PCM_STREAM_PLAYBACK, alsa::SND_PCM_NONBLOCK)).unwrap();
```
mitchmindtree: I think this line should just be:

``` Rust
check_errors(alsa::snd_pcm_open(&mut playback_handle, name.as_ptr(), alsa::SND_PCM_STREAM_PLAYBACK, alsa::SND_PCM_NONBLOCK)).unwrap();
```

As `name` will deref to a `CStr` and call `CStr`'s `.as_ptr()` which returns a `*const libc::c_char`

tomaka: Oh ok, thanks. This FFI API changes some often that I just try whatever works.

tomaka:
I'll merge if `cargo build` passes. Tests are probably not broken, but don't work because dev-deps are broken.

src/alsa/mod.rs
mitchmindtree: No prob, I tried the exact same thing as you did here when I was doing the glutin fixes :stuck_out_tongue:

PR #24, opened 2015-02-24
Correct reported ALSA sample format

kmcallister:


tomaka:
Are you sure that's correct? Did you try the "music" example?

kmcallister:
Doesn't line 28 say `alsa::SND_PCM_FORMAT_S16_LE`? Am I misunderstanding the implications of that call?

Anyway, I needed this change to make the output of `play -e signed` sound right in my program with an `i16` buffer.  It's possible the bug is somewhere else, though.

> Did you try the "music" example?

No, I tried at one point but couldn't build/install vorbis.

tomaka:
The music example is totally buggy and the beep example doesn't work.

tomaka:
Closing in favor of #51 

Turns out the music example was itself buggy.

PR #27, opened 2015-02-27
OSX support via the Apple Core Audio, Audio Unit C API. Only supports f32 samples so far. closes #15

### LINKED ISSUE #15: OSX support
Are there plans for OSX support?

mitchmindtree:
@tomaka  :santa: 

Sorry about the massive commit! It's taken this long just to get it building successfully hah.

Only the f32 sample type works at the moment - it's the canonical osx/ios sample format and I haven't handled conversions from the other types just yet. i.e. the beep.rs example works if I change it to use f32 samples instead, but it panics with the original u16. I'll try get this working properly tomorrow.

Also sorry about the huge `audio_unit` bindings (10kloc)! I'll go through it and clean out the unnecessary stuff soon... I'm actually thinking I might do a proper set of Core Audio bindings under the RustAudio group - if I do get it going, then we can swap it in as a dependency instead of `core_audio-sys`.

I've got some other questions and ideas related to api stuff but I'll just post them as separate issues tomorrow... so sleepy!

:zzz: 

ps: travis seems to successfully build but he times out due to a missing a key of some sort or something


tomaka:
Yay

bvssvni:
:balloon: @mitchmindtree You are are awesome!

PR #31, opened 2015-03-03
Fixed coreaudio callback to send proper buffersize, removed code in lib where sampleformat affected buffersize

mitchmindtree:


src/lib.rs
```diff
@@ -192,8 +192,6 @@ impl Voice {
             let max_elements = max_elements * target_channels as usize / channels as usize;
             let max_elements = max_elements * target_samples_rate.0 as usize /
                                samples_rate.0 as usize;
-            let max_elements = max_elements * target_samples_format.get_sample_size() /
```
mitchmindtree, line 195: @tomaka I'm pretty sure these lines were unnecessary, but could you double check? I don't think the number of elements within the buffer should be affected by the size of the sample format itself. While this code was in I kept getting a runtime error at `buffer lengths mismatch` [here](https://github.com/tomaka/cpal/blob/master/src/lib.rs#L292), however after I removed it it seems to run ok.

tomaka, line 195: Yeah it looks like a mistake

PR #37, opened 2015-04-18
Remove integer suffixes in alsa-sys

ruud-v-a:
The `us` suffixes have long been replaced with `usize`, but in this case, no suffixes are required at all.

I did not test this because I am at a Windows box at the moment. It should work in theory, and I hope this triggers the Travis build, so we will know.

**Edit**: The Travis build passed.


mitchmindtree:
Looks fine, thanks!

PR #51, opened 2015-07-22
Fix signed/unsigned formats mismatch

tomaka:
cc @ruud-v-a


ruud-v-a:
:thumbsup: Thanks, this solves my issue!

PR #55, opened 2015-08-18
Add more detailed message to panic

mceier:
When samples format is not supported, cpal panics with "not yet implemented" message, which is not useful in bug reports.
This adds samples format to the message.

Signed-off-by: Mariusz Ceier <mceier+cpal@gmail.com>


tomaka:
Thanks

PR #67, opened 2015-09-12
Endpoints and formats enumeration for ALSA backend

mceier:
Includes changes from alsa-formats branch; default endpoint is not default, but the first found; and there's some memory leak according to valgrind.

### DIFF
--- a/examples/enumerate.rs
+++ b/examples/enumerate.rs
@@ -1,0 +1,18 @@
+
+extern crate cpal;
+
+use cpal::*;
+
+fn main() {
+    let endpoints = cpal::get_endpoints_list();
+    
+    println!("Endpoints: ");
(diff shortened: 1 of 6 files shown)


mceier:
Travis uses old alsa-library and it doesn't have snd_pcm_query_chmaps and snd_pcm_free_chmaps functions, used for enumerating channels :/
I think this should be rewritten based on amixer source then:

 http://git.alsa-project.org/?p=alsa-utils.git;a=blob;f=amixer/amixer.c;h=db1849333da3c09d8661dd0d4d7adb6c953230e5;hb=0e864e1a3aaf79d451556064eb09052530a0a085

^ this is the version from asound 1.0.25 which is on travis.

mitchmindtree:
Going to close this as it's pretty old and both device and format enumeration and defaults for input and output streams should be working on ALSA now.

PR #81, opened 2015-09-24
Restore CoreAudio support after API overhaul.

Furyhunter:
Since the API changes, compilation on OSX fails. This restores compilation and the features that the CoreAudio backend was originally capable of (that is to say, only f32 samples and a fixed buffer size).

~~The most recent commit adds support-ish for U16/I16. AUHAL is able to convert between PCM formats internally, but we are doing it Rust-side before sending the buffer to the unit.~~

### DIFF
--- a/src/coreaudio/mod.rs
+++ b/src/coreaudio/mod.rs
@@ -2,6 +2,8 @@ extern crate coreaudio_rs as coreaudio;
 extern crate libc;
 
 use std::sync::mpsc::{channel, Sender, Receiver};
+use std::sync::{Arc, Mutex};
+use std::cell::RefCell;
 use std::mem;
 
 use CreationError;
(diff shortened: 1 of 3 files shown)


tomaka:
It looks like the code reports `U16` and `I16` are supported and does the conversion to f32 itself. This is wrong. Just report that only `F32` is supported

src/coreaudio/mod.rs
```diff
@@ -39,6 +148,7 @@ impl Voice {
                     samples: vec![unsafe{ mem::uninitialized() }; buffer_size],
                     num_channels: channels as usize,
                     marker: ::std::marker::PhantomData,
+                    len: 64
```
tomaka: This 64 is totally arbitrary. The length of the buffer is supposed to be `samples.len()`.

src/coreaudio/mod.rs
```diff
@@ -46,76 +156,19 @@ impl Voice {
...
-        let Buffer { samples_sender, samples, num_channels, .. } = self;
-        // TODO: At the moment this assumes the Vec<T> is a Vec<f32>.
-        // Need to add T: Sample and use Sample::to_vec_f32.
-        let samples = unsafe { mem::transmute(samples) };
-        match samples_sender.send((samples, num_channels)) {
-            Err(_) => panic!("Failed to send samples to audio unit callback."),
-            Ok(()) => (),
-        }
+    pub fn get_pending_samples(&self) -> usize {
+        0
```
tomaka: I can't merge until this is supported. It's too important.

src/coreaudio/mod.rs
```diff
@@ -46,76 +156,19 @@ impl Voice {
...
-            audio_unit: audio_unit,
-            ready_receiver: ready_receiver,
-            samples_sender: samples_sender
-        }),
-        Err(err) => {
-            use ::std::error::Error;
-            Err(err.description().to_string())
-        },
+    pub fn underflowed(&self) -> bool {
+        false
```
tomaka: Please either put `unimplemented!()` or implement it, but returning a wrong dummy value leads to bugs that are very annoying to debug.

src/coreaudio/mod.rs
GuillaumeGomez: Dark magic !

Furyhunter:
The implementations of `underflow` and `get_pending_samples` are a little rough and may have performance issues since they use mutex refcells to communicate between the audiounit thread and the the owner of the Voice.

tomaka:
Closing in favor of #94

PR #86, opened 2015-10-13
Catch another 'device not found' error code with ALSA

mvdnes:
On a device with no sound (such as [Travis-CI][travis]) I could get a default endpoint but enumerating the formats would then panic. This patch adds the error code that my case returned, and which I believe should be caught instead of panic.

Looking at the code, I think a more rigorous change would be nice, but in the meantime I propose merging this fix.

[travis]: https://travis-ci.org/mvdnes/rboy/builds/85080893


tomaka:
Thanks

PR #96, opened 2016-01-14
allow for building for mipsel targets

psych0d0g:
mipsel is mostly used on embedded multimedia devices like settop bockes for eg.
This change will allow building cpal for those platforms aswell


tomaka:
Thanks

PR #100, opened 2016-01-28
Fix underflow bug on linux

nicokoch:
fixes tomaka/rodio#67


tomaka:
Does it really fix the problem? The change seems totally random.

nicokoch:
It definitely does. Let me elaborate a bit.
Before, when the underflow occured, `get_pending_samples` returned `self.buffer_len`.

This in turn caused [this](https://github.com/tomaka/rodio/blob/master/src/engine.rs#L293) line to always return false, which was why rodio stopped writing to the output. This in turn means that new samples were never written to the output.

tomaka:
Ok, that seems reasonable.
Thanks!

PR #101, opened 2016-01-30
Using ComPtr for COM

retep998:
I dunno, what do you think?


mitchmindtree:
This actually looks much nicer IMO, but seeing as this is so old and needs a rebase I'm going to close it for now. Feel free to re-open if feeling inspired to rebase though!

PR #107, opened 2016-04-02
alsa-sys: Add pollfd

diwic:
The pollfd struct got added to libc in v0.2.7, so now we can uncomment those commented functions, enabling us to use async I/O for ALSA.


tomaka:
Thanks!

PR #112, opened 2016-06-01
[osx] Add get_period to Voice

Furyhunter:
cpal fails to compile on OSX without the get_period method on Voice, this PR provides it as it appears in the wasapi backend. The examples work as intended.

### DIFF
--- a/src/coreaudio/mod.rs
+++ b/src/coreaudio/mod.rs
@@ -161,6 +161,11 @@ impl Voice {
     }
 
     #[inline]
+    pub fn get_period(&self) -> usize {
+        0
+    }
+
+    #[inline]
     pub fn get_pending_samples(&self) -> usize {
         if let Some(ready) = self.update_last_ready() {
             (ready.0 * ready.1) as usize
--- a/src/coreaudio/mod.rs
+++ b/src/coreaudio/mod.rs
@@ -3,6 +3,7 @@ extern crate libc;
 
 use std::sync::mpsc::{channel, Sender, Receiver};
 use std::sync::{Arc, Mutex};
+use std::sync::atomic::{AtomicUsize, Ordering};
 use std::cell::RefCell;
 use std::mem;
 use std::cmp;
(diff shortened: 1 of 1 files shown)


tomaka:
Is the value actually correct?
It's not because the example works that the value is good.

Furyhunter:
I believe the closest analogue is `num_frames` [in the callback args](https://github.com/RustAudio/coreaudio-rs/blob/18e607fd568399ab0e38a2794ca419506234f3a3/src/audio_unit/render_callback.rs#L43) but `get_period` is on the Voice itself, so there is an API discrepancy here.

I think the best way to go about this would be to store a Mutex or RwLock and mutate it whenever the callback is called, and return its current value in `get_period`. But I'm still not certain that `num_frames` is _exactly_ what this is calling for.

Furyhunter:
Did some more research into this. The AudioUnit callback specifies the number of frames per channel requested by the OS, and it is not possible to _set_ the callback interval/buffer size, only a hint can be provided (but that should be implemented in another PR).

For this, `get_period()` is semantically identical to the `get_pending_frames()` implementation, which waits for the ready event from the ready channel to indicate the number of frames and channels requested. However, the public `get_period()` interface does not specify if users should expect the period to be the same for any given Voice instance. I will push a commit implementing this, but that I think is worth putting in the documentation, because there's nothing that says the OS _has_ to return the same `inNumberFrames` on each render. In practice, though, my MacBook Air _did_ request exactly 512 frames (~11.6ms 44.1) every call for the beep example.

tomaka:
It's weird that `get_period()` would be the same as `get_pending_samples()`, but according to your explanation it would mean to me that `get_pending_samples()`'s implementation is wrong.

Furyhunter:
I am confused what the difference is supposed to be, or what "currently being processed" means. Is `get_pending_samples` supposed to be the number of samples submitted to the sample channel in this situation?

tomaka:
> Is get_pending_samples supposed to be the number of samples submitted to the sample channel in this situation?

Yes, the number of samples submitted minus the number that have been processed already.

Furyhunter:
Provided CI passes, the most recent commit should provide a correct implementation of `get_pending_samples` for coreaudio.

PR #114, opened 2016-07-05
coreaudio: update to coreaudio-rs 0.5.0

jansol:
- [x] typechecks
- [x] builds
- [x] beep example makes a sound
- [ ] I know what I'm doing

From what I could gather from coreaudio-rs docs with my limited Rust knowledge this seems okay but please do review...

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,7 +32,7 @@ version = "0"
 path = "alsa-sys"
 
 [target.x86_64-apple-darwin.dependencies]
-coreaudio-rs = "~0.4.0"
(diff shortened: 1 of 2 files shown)


jansol:
While the patch seems to work fine, I get massive stuttering in example code. Since I don't know if cpal ever actually worked on OSX I decided to poke around a bit. After some more testing with rodio it seems the stuttering is caused by samples_receiver.try_recv() on src/coreaudio/lib.rs:122 taking an undeterministic time to complete, which is a no-go in the audio unit render callback. (can that even happen?)

PR #121, opened 2016-08-02
Draft for switching to futures

tomaka:
Windows-only for now.

Changes:

- `Voice::new()` requires passing an `EventLoop` which can be created with `EventLoop::new()`.
- `Voice::new()` now returns a `(Voice, SamplesStream)`. The `SamplesStream` implements the `futures::stream::Stream` trait and produces `UnknownTypeBuffer` objects.
- The `UnknownTypeBuffer` and `Buffer` structs are now `'static`.
- `EventLoop::run()` needs to be called on a thread in order to be guarateed to have the stream produce buffers.
- All the old methods that are now irrelevant have been removed.

On Windows, the implementation now calls `SetEventHandle` on the audio client. Calling `SamplesStream::poll()` checks whether the `Event` is signalled and returns a buffer if so. Calling `SamplesStream::schedule()` adds the event to a list in the event loop. Calling `EventLoop::run()` repeatedly waits for any of the events that have been added to the list and signals the appropriate `TaskHandle`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,7 +34,7 @@ version = "0"
 path = "alsa-sys"
 
 [target.x86_64-apple-darwin.dependencies]
-coreaudio-rs = "~0.5.0"
+coreaudio-rs = "0.6"
 
 [dev-dependencies]
 vorbis = "0"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,16 +1,15 @@
 [package]
-
 name = "cpal"
-version = "0.2.12"
-authors = ["Pierre Krieger <pierre.krieger1708@gmail.com>"]
-description = "Cross-platform audio playing library in pure Rust."
+version = "0.3.0"
+authors = ["The CPAL contributors", "Pierre Krieger <pierre.krieger1708@gmail.com>"]
+description = "Low-level cross-platform audio playing library in pure Rust."
 repository = "https://github.com/tomaka/cpal"
 documentation = "http://tomaka.github.io/cpal/"
 license = "Apache-2.0"
 keywords = ["audio", "sound"]
 
 [dependencies]
-futures = { git = "https://github.com/alexcrichton/futures-rs" }
+futures = "0.1.0"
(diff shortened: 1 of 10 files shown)


mitchmindtree:
Interesting!

I only have a very basic understanding of futures and I've not personally used them before. Could you describe how this change will benefit CPAL, and perhaps how it differs from regular callbacks? Looking at the example, it looks almost identical to a regular callback, however I suspect the differences lie in that `SamplesStream` type?

tomaka:
It is very similar to callbacks. On OSX the callback registered to `SamplesStream` will be automatically called by coreaudio from an external thread.

On Windows and Linux, however, the engine doesn't magically spawn a new thread to call the callbacks. Instead the callback will be called one by one when you call `run()` on the events loop.

For example let's say you create 3 voices that each use the same event loop. When you call `event_loop.run()`, the thread will start sleeping. When one of the 3 voices is ready to get some data, the thread will wake up and call the corresponding callback. Then if one of the other voices is ready as well, we call its callback afterward. Otherwise sleep again and wake up when one of the voices is ready.

Currently in order to play a sound you have to write data, then sleep an approximate amount of time, then write, then sleep, and so on. I designed cpal this way because this is what most examples on the Internet do. But in practice it is very tedious when you have multiple sounds to update, as it is the case with rodio. By using this callbacks-like design you don't have a risk of stuttering. The coreaudio backend will also be much more natural to write.

As for why this uses futures-rs (as opposed to implementing a callback system manually), it is because futures-rs will likely become the de-facto way of performing non-blocking operations in the future.

tomaka:
I wrote a gist of what I think the coreaudio implementation should look like: https://gist.github.com/tomaka/5b0b0c77d826ec9ec9ac7f257c9b738c

I still don't have any Apple machine, so I can't test it/finish it. It would be nice if someone did it!

tomaka:
To my surprise, this seems to compile with futures 0.1. So here we go, let's merge this.

PR #130, opened 2016-09-30
update to futures 0.1.1: alsa wasapi

thiolliere:
here is a try to update cpal to use futures 0.1.1 instead of yanked futures 0.1.0, it is related to issue 129

change is made on streams: when poll return not ready, the function schedule the task on the event loop.

also there is change for users: in beep example the user spawn a task for the stream and use its own executor to handle its execution.

current state of this branch:
* alsa is OK for me: beep example works.
* wasapi compile but the execution on wine of beep.exe sounds crap
* coreaudio nothing has been made

### DIFF
--- a/examples/beep.rs
+++ b/examples/beep.rs
@@ -1,14 +1,27 @@
 extern crate cpal;
 extern crate futures;
 
-use futures::Future;
 use futures::stream::Stream;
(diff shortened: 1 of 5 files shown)


tomaka:
Thanks!

PR #131, opened 2016-10-01
Try fix the OSX code with futures

tomaka:
As usual I don't have any OSX machine. It would be nice if someone could test it?

cc @mitchmindtree maybe?


tomaka:
Rebased over #132

mitchmindtree:
Ooo looks nice. It's compiling, but the closure in the beep example is no longer being called. I'll have a closer look in a minute

PR #132, opened 2016-10-01
Try enabling travis for OSX

tomaka:


tomaka:
Merging since OSX is now tested!

PR #133, opened 2016-10-01
Do not lock inner twice. Fixes bug in osx futures 0.1.1 update.

mitchmindtree:
After applying this, #131 works for me (beep example works fine) :+1:


tomaka:
Thanks!

PR #135, opened 2016-10-02
impl play and pause for ALSA

thiolliere:
snd_pcm_pause could have been used but not all hardware implement it, so
I propose not to use it.

In this implementation:

there are two kind of scheduling: wait for resume signal and wait for
pcm to be available

if the stream is paused then it return notready and wait for resume

the event loop is different as it manages descriptors corresponding to
voices according to the nature of the scheduling.

there is still a FIXME: in voice.play the is signal is send even if
the event loop wasn't waiting for resume.
It doesn't seem to create any issue. But it happens when you write
voice.pause();voice.play();

### DIFF
--- a/src/alsa/mod.rs
+++ b/src/alsa/mod.rs
@@ -629,7 +629,7 @@ impl Voice {
     pub fn play(&mut self) {
         // If it was paused then we resume and signal
         // FIXME: the signal is send even if the event loop wasn't waiting for resume, is that an issue ?
-        if self.inner.is_paused.swap(false,Ordering::Relaxed) {
+        if self.inner.is_paused.swap(false, Ordering::Relaxed) {
             unsafe {
                 let buf = 1u64;
                 let wret = libc::write(self.inner.resume_signal,
@@ -641,7 +641,7 @@ impl Voice {
 
     #[inline]
     pub fn pause(&mut self) {
-        self.inner.is_paused.store(true,Ordering::Relaxed);
+        self.inner.is_paused.store(true, Ordering::Relaxed);
     }
 }
 
--- a/src/alsa/mod.rs
+++ b/src/alsa/mod.rs
@@ -601,9 +601,6 @@ impl Voice {
                 num_descriptors as usize
             };
 
-            // The voice is initialized as paused
-            let is_paused = Arc::new(AtomicBool::new(true));
-
             let samples_stream_inner = Arc::new(VoiceInner {
                 event_loop: event_loop.inner.clone(),
                 channel: Mutex::new(playback_handle),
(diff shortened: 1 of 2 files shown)


src/alsa/mod.rs
```diff
@@ -349,33 +395,56 @@ struct VoiceInner {
     period_len: usize,
 
     // If `Some`, something previously called `schedule` on the stream.
-    scheduled: Mutex<Option<Task>>,
+    scheduled: Mutex<Option<Scheduled>>,
+
+    // Wherease the sample stream is paused
+    is_paused: Arc<AtomicBool>,
```
tomaka: I don't think this requires an `Arc`.

src/alsa/mod.rs
```diff
@@ -526,6 +601,9 @@ impl Voice {
                 num_descriptors as usize
             };
 
+            // The voice is initialized as paused
+            let is_paused = Arc::new(AtomicBool::new(true));
```
tomaka: This line is not necessary.

src/alsa/mod.rs
```diff
@@ -535,23 +613,35 @@ impl Voice {
...
+                                       &buf as *const u64 as *const _, 8);
+                assert!(wret == 8);
+            }
+        }
     }
 
     #[inline]
     pub fn pause(&mut self) {
-        unimplemented!()
+        self.inner.is_paused.store(true,Ordering::Relaxed);
```
tomaka: Put a space before the comma please.

src/alsa/mod.rs
```diff
@@ -535,23 +613,35 @@ impl Voice {
...
         }
     }
 
     #[inline]
     pub fn play(&mut self) {
-        // already playing
-        //unimplemented!()
+        // If it was paused then we resume and signal
+        // FIXME: the signal is send even if the event loop wasn't waiting for resume, is that an issue ?
+        if self.inner.is_paused.swap(false,Ordering::Relaxed) {
```
tomaka: Space before the comma.

src/alsa/mod.rs
```diff
@@ -396,6 +465,12 @@ impl Stream for SamplesStream {
     type Error = ();
 
     fn poll(&mut self) -> Poll<Option<Self::Item>, Self::Error> {
+        // If paused then we schedule the task and return `NotReady`
+        if self.inner.is_paused.load(Ordering::Relaxed) {
```
tomaka, line 469: I'm not sure that `Relaxed` is the good thing.
If you call `voice.pause(); stream.poll();` in the same thread, is it possible that the compiler loads the value for `poll()` before writing it for the `voice`?

tomaka: People on IRC say that it's ok to leave Relaxed.

src/alsa/mod.rs
gui1117: yes, it was inherit from an old design

src/alsa/mod.rs
gui1117: OK

src/alsa/mod.rs
gui1117: OK

src/alsa/mod.rs
gui1117: OK

PR #136, opened 2016-10-18
[osx] Implement play/pause for coreaudio

HybridEidolon:
Since `play`/`pause` do not specify a Result return type, these will panic if the underlying calls to audiounit return `Err`.


tomaka:
Thanks

Issue #91, opened 2015-11-11
getters/setters should follow idiomatic Rust naming conventions


yupferris:
According to [this RFC](https://github.com/rust-lang/rfcs/blob/master/text/0344-conventions-galore.md#gettersetter-apis), it should be `set_` for setters, and no prefix for getters (currently this repo uses `get_`).

mitchmindtree:
Closing as this change has been implemented.

PR #138, opened 2016-11-02
Changed getters to match rust convention as pointed out in issue #91

torkleyy:
* Added functions `get_endpoints_list()` and `get_default_endpoint()` to Endpoint structure
* Removed the `get_` part from all getters except from the deprecated ones

### DIFF
--- a/alsa-sys/build.rs
+++ b/alsa-sys/build.rs
@@ -1,5 +1,7 @@
 extern crate pkg_config;
 
 fn main() {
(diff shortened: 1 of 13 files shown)


tomaka:
I'm in favor of the change, however it shouldn't be straight renaming.
Instead we should add `#[deprecated]` attributes to the existing methods and add the new ones in parallel, in order to not break anything.

torkleyy:
I'll refork the original branch and doing this changes

PR #143, opened 2017-01-19
Add coreaudio-rs dependency for i686-apple-darwin

yupferris:
Fixes build for 32-bit mac. Tested locally in both configs, works just fine!

Also, can I request a minor release be published after merge? Would be nice to use this ASAP :) Thanks!

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "cpal"
-version = "0.4.1"
+version = "0.4.2"
 authors = ["The CPAL contributors", "Pierre Krieger <pierre.krieger1708@gmail.com>"]
 description = "Low-level cross-platform audio playing library in pure Rust."
 repository = "https://github.com/tomaka/cpal"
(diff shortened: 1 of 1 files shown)


yupferris:
This is likely all that's needed to get rodio working for this target as well, though I haven't tested that specifically.

tomaka:
> Also, can I request a minor release be published after merge? Would be nice to use this ASAP :) Thanks!

Just modify the version number in the Cargo.toml. You can add it to this PR.

yupferris:
Should I update the one in `README.md` as well, or should that one continue to read "0.4.0" ?

yupferris:
ah nvm, I hadn't realized ["0.4.0" would be the same as "^0.4.0" in cargo](http://doc.crates.io/specifying-dependencies.html#specifying-dependencies-from-cratesio). Should be good to go here then with https://github.com/tomaka/cpal/pull/143/commits/6e8f51de09ebb45489d741262dfdf5255d853a1e :)

PR #145, opened 2017-01-25
alsa: set sw_params_set_avail_min based on get_params buffer size

jwestfall69:
snd_pcm_sw_params_set_avail_min was being hardcoded to 4096, which
seems to be problematic for lower sample rates.  This update sets
the value to the buffer size as supplied by snd_pcm_get_params(),
which is what alsa own sample code does.

This should fix https://github.com/tomaka/cpal/issues/142


tomaka:
Thanks

Issue #142, opened 2016-12-30
Running beep.rs example with ALSA only loads data to buffer once


TheVillageIdiot2:
No modifications were made to code. When I put println! statements in the stream.for_each closure, they only are called once. Issue persists with both version 0.4.0 and 0.4.1 of crate.

No panics or anything, and the program is properly blocking on the event_loop.run() call. Really not sure what could be going wrong. I can post more specific details of my system if that would help, but I'm fairly new to rust/audio stuff so not sure what would be relevant.

nchashch:
I tried to debug this problem and I have found that `libc::poll` call in `EventLoop::run` method in src/alsa/mod.rs at L258 doesn't return (it seems it just runs in an infinite loop) after a few iterations (hence a short bit of sound plays at the begining).

Also on my machine `Buffer::finish` method (defined at L674 in src/alsa/mod.rs) is called only once or twice (so `alsa::snd_pcm_writei` is called only once or twice) after that `SamplesStream::schedule` is called (maybe something goes wrong there) and then `Buffer::finish` is never called again.

jwestfall69:
Can you guys verify the above patch fixes the beep stall.  

I'm a little unsatisfied with the patch with where I'm calling alsa::snd_pcm_sw_params_set_avail_min(),.  Its inside the bits that are setting up buffer_len/period_len values, which seems like bad form.

tomaka:
@jwestfall69 The style is not a problem for me.

PR #146, opened 2017-01-25
alsa: Pass period instead of buffer to snd_pcm_sw_params_set_avail_min

nchashch:
#145 Fixed the stalling (#142), but there still was a stuttering problem.
Passing `period` instead of `buffer` to `snd_pcm_sw_params_set_avail_min` fixes the stuttering.
Also crate version is bumped to ~`0.4.3`~ `0.4.4`in this PR.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "cpal"
-version = "0.4.3"
+version = "0.4.4"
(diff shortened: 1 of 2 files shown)


jwestfall69:
Sorry I missed this PR fixing my PR.  It does stop the stuttering/poping for me too.  I thought it was being caused by a timing issue with the app I was messing them.

Think we can get this applied, and a bump to 0.4.4?

tomaka:
Thanks!

PR #151, opened 2017-02-10
Remove the `get_` prefix of methods

tomaka:


src/lib.rs
```diff
@@ -120,27 +120,50 @@ impl Iterator for EndpointsIterator {
...
+}
+
 /// An opaque type that identifies an end point.
 #[derive(Clone, PartialEq, Eq)]
 pub struct Endpoint(cpal_impl::Endpoint);
 
 impl Endpoint {
     /// Returns an iterator that produces the list of formats that are supported by the backend.
     #[inline]
+    pub fn supported_formats_list(&self) -> Result<SupportedFormatsIterator,
```
jonas-schievink: I feel like this is still a bit of a misnomer. It produces an iterator, not a list. Just "supported_formats" would be enoguh, I think.

tomaka: Ah right, that's an overlook.

Issue #148, opened 2017-02-07
CPAL not outputting sound under Mac OS X x86_64.


SirVer:
This is the followup bug to https://github.com/tomaka/rodio/issues/94.
This is a MacBookPro Retina 15 Inch, Early 2013 with Intel Core i7 running OS X El Capitan.

CPAL does not crash, but the examples do not play a sound under my Mac OS X:

~~~
$ cargo run --release --example beep
    Finished release [optimized] target(s) in 0.0 secs
     Running `target/release/examples/beep`
thread 'main' panicked at 'Failed to get default endpoint', /Users/rustbuild/src/rust-buildbot/slave/stable-dist-rustc-mac/build/src/libcore/option.rs:715
stack backtrace:
   1:        0x1084a102a - std::sys::imp::backtrace::tracing::imp::write::hd3b65cdfe843284c
   2:        0x1084a27cf - std::panicking::default_hook::{{closure}}::hf2b7428652613d83
   3:        0x1084a2477 - std::panicking::default_hook::h5da8f27db5582938
   4:        0x1084a2c96 - std::panicking::rust_panic_with_hook::hcef1e67c646c6802
   5:        0x1084a2b34 - std::panicking::begin_panic::hc2e8ca89533cd10d
   6:        0x1084a2a52 - std::panicking::begin_panic_fmt::h60990696c3c3a88d
   7:        0x1084a29b7 - rust_begin_unwind
   8:        0x1084c4c40 - core::panicking::panic_fmt::h10231c789bd0e97d
   9:        0x1084c4cad - core::option::expect_failed::h77d0b34eebcbdfc8
  10:        0x10849808e - beep::main::h7bcc7202bef6bbee
  11:        0x1084a382a - __rust_maybe_catch_panic
  12:        0x1084a2f06 - std::rt::lang_start::h87cb84a8b6cb187e

$ cargo run --release --example enumerate
   Compiling cpal v0.4.4 (file:///private/tmp/cpal)
    Finished release [optimized] target(s) in 0.83 secs
     Running `target/release/examples/enumerate`
Endpoints:
1. Endpoint "Default AudioUnit Endpoint" Audio formats:
1.1. Format { channels: [FrontLeft, FrontRight], samples_rate: SamplesRate(44100), data_type: F32 }
~~~

shockham:
@SirVer Do you get any sound with a debug build?

[...]

PR #154, opened 2017-04-07
Fix for loop in EventLoop::run optimised out in a release build on macOS

shockham:
Fix in relation to #148 with the loop in EventLoop::run optimised out in a release build on macOS, doesn't feel very elegant, so am open to suggestions for something nicer.

### DIFF
--- a/src/coreaudio/mod.rs
+++ b/src/coreaudio/mod.rs
@@ -93,12 +93,13 @@ impl<T> Buffer<T> where T: Sample {
 
 pub struct Voice {
     playing: bool,
-    audio_unit: AudioUnit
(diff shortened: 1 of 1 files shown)


jonas-schievink:
cc https://github.com/rust-lang/rust/issues/28728

shockham:
Added a commit to this which addresses tomaka/rodio#107

tomaka:
I don't know how coreaudio works, but I guess I can merge this anyway?

shockham:
Yeah I think so, just gets things working again

tomaka:
Thanks

PR #158, opened 2017-07-13
Fixes for *BSDs

johalun:
Replace linux-only eventfd() with pipe() to enable use on *BSDs for alsa.
Add FreeBSD to supported OSes. 
Print warning to stdout if using null device.

Note: not tested on Linux.

### DIFF
--- a/src/alsa/mod.rs
+++ b/src/alsa/mod.rs
@@ -26,6 +26,48 @@ pub type SupportedFormatsIterator = VecIntoIter<Format>;
 
 mod enumerate;
 
+
+struct Trigger {
+    // [read fd, write fd]
+    fds: [libc::c_int; 2],
+}
+
+impl Trigger {
+    fn new() -> Self {
+        let mut fds = [0,0];
+        match unsafe { libc::pipe(fds.as_mut_ptr()) } {
+            0 => Trigger { fds: fds },
+            _ => panic!("Could not create pipe")
+        }
+    }
+    fn read_fd(&self) -> libc::c_int {
+        self.fds[0]
+    }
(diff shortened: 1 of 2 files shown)


src/null/mod.rs
```diff
@@ -33,6 +41,7 @@ impl Iterator for EndpointsIterator {
 
 #[inline]
 pub fn get_default_endpoint() -> Option<Endpoint> {
+    println!("Unknown target system, default to null device.");
```
tomaka: Please remove this println, it's unrelated to the rest of the PR

src/null/mod.rs
```diff
@@ -14,14 +15,21 @@ use UnknownTypeBuffer;
...
     #[inline]
     pub fn run(&self) { loop { /* TODO: don't spin */ } }
 }
 
-#[derive(Default)]
 pub struct EndpointsIterator;
 
+impl Default for EndpointsIterator {
+    fn default() -> Self {
+        println!("Unknown target system, default to null device.");
```
tomaka: Same

johalun:
We can do a "alsa-pipe" especially for Unix but since Linux also has pipe() same as Unix I don't see any reason for double implementations. 

I think showing the user that null devices is being used is good. It took me a while to track down why it didn't work. Why it couldn't find any devices.. Turned out it was using null device since freebsd was unsupported.

johalun:
Fixed.

tomaka:
> I think showing the user that null devices is being used is good. It took me a while to track down why it didn't work. Why it couldn't find any devices.. Turned out it was using null device since freebsd was unsupported.

I agree but a println is not the right solution for this.

PR #161, opened 2017-10-10
Minimal ios support

matwork:


tomaka:
Thanks for the PR.

I don't know anything about Mac and iOS, so what is the difference between `DefaultOutput` and `RemoteIO`? If you want, adding a code comment that explains that would be nice.

matwork:
The DefaultOutput unit isn't available in iOS unfortunately. RemoteIO is a sensible replacement.

Here is what the iOS doc says:
iOS provides three I/O (input/output) units. The vast majority of audio-unit applications use the Remote I/O unit, which connects to input and output audio hardware and provides low-latency access to individual incoming and outgoing audio sample values. For VoIP apps, the Voice-Processing I/O unit extends the Remote I/O unit by adding acoustic echo cancelation and other features. To send audio back to your application rather than to output audio hardware, use the Generic Output unit.

See https://developer.apple.com/library/content/documentation/MusicAudio/Conceptual/AudioUnitHostingGuide_iOS/UsingSpecificAudioUnits/UsingSpecificAudioUnits.html

PR #165, opened 2017-10-12
RFC: Rework the API [WIP]

### LINKED ISSUE #141: Consider switching out of futures
And use a custom future-like API.

This is mostly just a note. I'll go into more details if I decide to work on it.

tomaka:
Close #141 

This PR modifies the API of cpal in order to not use futures anymore.

Using futures has several drawbacks. It makes the API more complex, forces the usage of Arcs/Mutexes in the backends, and could introduce semver issues in the future.

Additionally, if we wanted to make cpal compatible with tokio (so that they use the same architecture), we could only do that on Linux. The Win32 and MacOS backends use an API that is totally different from the sockets API of their platform, and therefore need their own events loop anyway. Therefore there's no real advantage to using futures. Any futures tutorial is more complex than cpal's new API anyway.

The changes also remove the `Voice` struct. A voice is now simply a concept, and is referred by a `VoiceId` that can be cloned, compared and hashed. In other words, it's now more clear that the events loop owns the voices.


tomaka:
[Link to the new API](https://github.com/tomaka/cpal/blob/f107251547b3e4586581f97b05b61b12b8632832/src/lib.rs)

tomaka:
I updated the MacOS backend, but I don't know if it's working.

HybridEidolon:
I can confirm the CoreAudio backend works fine. Both examples run fine.

tomaka:
So for now the API only proposes a `run()` method that blocks the current thread entirely, and calls a callback whenever some data is needed.

This can be a bit annoying because it forces you to spawn a background thread, and makes the library annoying to use on emscripten who doesn't allow threads (who wants to block their whole program just for sound?).

I think in the future we should add some sort of `detached_run` function that, depending on the backend, either spawns a thread (Win32/Linux), does nothing (MacOS), or sets up a callback system (Emscripten). I wanted to give it a shot, but I realized that the details look more complicated than anticipated. I think that this should be postponed.

HybridEidolon:
Probably worth mentioning that SDL_Audio's API doesn't expose a loop at all, it creates a thread or just delegates to AudioUnit executing a provided callback. `detached_run` makes sense to me, almost so far as _only_ offering that.

Additionally, the surface area for the cpal abstraction layer could probably both allow for separate, delegate backends and a core-only mode now. It'd be cool to see cpal be usable on embedded systems that have DSPs but no libstd support.

PR #167, opened 2017-10-19
examples: Avoid f32 precision loss in beep example

### LINKED ISSUE #157: Beep example distorts sine in the long run
I ran the beep example on three different machines (linux, mac and windows) all slightly adding noisy tones on top of the sine. Why?

jansol:
Closes #157


tomaka:
Thanks

Issue #175, opened 2017-10-31
Clarify whether or not a `Voice` should be paused following creation and ensure each backend abides by this.


mitchmindtree:
Currently the `beep` example calls `play` after constructing a voice, implying that upon creation it is paused by default which seems totally reasonable.

However, on Linux a `Voice` immediately begins playing during construction whether or not `play` is called.

Perhaps we should clarify that voices should be `paused` upon creation in the `build_voice` docs and fix this implementation in the ALSA backend.

tomaka:
I had already thought about this problem, but I can't decide whether voices should be initially playing or initially paused.

mitchmindtree:
I'm leaning towards paused by default, but don't really have a good reason as to why. I'm open to either, as long as we settle upon something for consistency. Maybe it's worth seeing what each of the backends do by default, and if they're all the same go with that? ALSA seems to play by default. I can't remember what coreaudio does off the top of my head.

tomaka:
The argument I see in favour of pause by default is that you might want to create multiple voices at initialization without running the events loop immediately, in which case you'd prefer the voices to be paused.

However this is a minor use case, and other than that having to cast play every time you create a voice can be annoying, and if you forget to do it then you might not understand why things aren't working.

tomaka:
Another thing to take into account is that if you create a play-by-default voice then pause it, you might have a problem because the backend attempted to get data between the creation of the voice and the pausing.

Alternatively we could add a `build_paused_voice` method.

mitchmindtree:
> Alternatively we could add a build_paused_voice method.

Yeah I like this idea - I'd be happy for us to have `build_voice` return a playing voice and then have `build_paused_voice` call `build_voice` and `pause` immediately after?

tomaka:
> Yeah I like this idea - I'd be happy for us to have build_voice return a playing voice and then have build_paused_voice call build_voice and pause immediately after?

Well, `build_paused_voice` would have to be implementation-specific. Otherwise you have the problem that I mentioned.

Alternatively, the opposite would work: `build_voice()` calls `build_paused_voice()` then `play()`.

roderickvd:
`Voice` has become `Stream` and has play/pause semantics that everybody seems happy with.

PR #176, opened 2017-10-31
Implement `pause` and `play` for ALSA backend

mitchmindtree:
This commit also ensures that the Voice is initially paused when returned to remain consistent with the curent state of the rest of the CPAL backends (at least I think they're paused by default at the moment?). Related discussion on this at #175.

The ALSA docs mention that not all hardware supports pause functionality, so we check for this using the `snd_pcm_hw_params_can_pause` function. Currently, `pause` and `play`  only do something if `can_pause` is `true`. I wonder what we should do In the case that  the hardware doesn't support pause functionality? Should we just submit silence and not call the callback? Or just do nothing at all as is currently the case?

### DIFF
--- a/src/alsa/mod.rs
+++ b/src/alsa/mod.rs
@@ -13,7 +13,7 @@ use SupportedFormat;
 use UnknownTypeBuffer;
 
 use std::{cmp, ffi, iter, mem, ptr};
-use std::sync::{Arc, Mutex};
+use std::sync::Mutex;
 use std::sync::atomic::{AtomicBool, AtomicUsize, Ordering};
 use std::vec::IntoIter as VecIntoIter;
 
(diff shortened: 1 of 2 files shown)


mitchmindtree:
I just tested this on a downstream project and although pause/play now work, the returned stream is still playing despite calling `snd_pcm_pause` before returning the voice, which I'm guessing has something to do with the code following the handling of the commands. Will address this tomorrow.

mitchmindtree:
I've removed the attempt at pausing the voice on creation for now - will address and #175 in a future PR. This should be ready to go now :+1:

src/alsa/mod.rs
```diff
@@ -309,7 +311,10 @@ struct VoiceInner {
     // Minimum number of samples to put in the buffer.
     period_len: usize,
 
-    // Wherease the sample stream is paused
+    // Whether or not the hardware supports pausing the stream.
+    can_pause: bool,
+
+    // Whether or not the sample stream is currently paused.
     is_paused: AtomicBool,
```
tomaka: Could you also change this so that it's a regular `bool` and not an `AtomicBool`?

tomaka:
Err, sorry for the ping-pong but this also needs a line in the changelog.

tomaka:
Thanks

PR #178, opened 2017-11-01
ALSA - Specify pending_trigger fd regardless of processed `Command`s

### LINKED ISSUE #177: ALSA backend call to `libc::poll` does not wakeup when building a new voice.
If `EventLoop::run`  is called (on a different thread) prior to building a new voice, the `EventLoop` thread becomes blocked [here](https://github.com/tomaka/cpal/blob/master/src/alsa/mod.rs#L391)

```rust
                let ret = libc::poll(run_context.descriptors.as_mut_ptr(),
                                     run_context.descriptors.len() as libc::nfds_t,
                                     -1 /* infinite */);

```

and does not wakeup when a new voice is created, despite the `build_voice` method writing to the file descriptor created for the purpose of waking up `poll` [here](https://github.com/tomaka/cpal/blob/master/src/alsa/mod.rs#L608)

```rust
 self.pending_trigger.wakeup();
```

If anyone has any ideas on what might be going on here that would be greatly appreciated! I'm unfamiliar with the `libc::poll` function and and its usage so I'm learning as I go here.

I wonder if it has something to do with the `descriptors` buffer being empty, and as a result `poll` has not been told what "events" it should be waiting for? Currently, a descriptor is only specified if there was some `Command` to process:

```rust
// process commands
                        run_context.descriptors = vec![
                            libc::pollfd {
                                fd: self.pending_trigger.read_fd(),
                                events: libc::POLLIN,
                                revents: 0,
                            },
                        ];
```

Perhaps a descriptor of some sort should be specified regardless of whether or not some commands were processed? Edit: This does seem to fix the issue, though I'm unsure whether or not this is the "correct" fix. @tomaka are you familiar with the alsa backend at all? If so it would be great to get your thoughts.

mitchmindtree:
This closes #177 for me, however it would be great to have someone more
familiar with ALSA have a look at this as I'm unsure whether or not we *should* be doing this before every call to `poll`.

### DIFF
--- a/src/alsa/mod.rs
+++ b/src/alsa/mod.rs
@@ -353,39 +353,36 @@ impl EventLoop {
             loop {
                 {
                     let mut commands_lock = self.commands.lock().unwrap();
-                    if !commands_lock.is_empty() {
-                        for command in commands_lock.drain(..) {
-                            match command {
-                                Command::DestroyVoice(voice_id) => {
-                                    run_context.voices.retain(|v| v.id != voice_id);
-                                },
-                                Command::NewVoice(voice_inner) => {
-                                    run_context.voices.push(voice_inner);
-                                },
-                            }
-                        }
-
-                        run_context.descriptors = vec![
-                            libc::pollfd {
-                                fd: self.pending_trigger.read_fd(),
-                                events: libc::POLLIN,
-                                revents: 0,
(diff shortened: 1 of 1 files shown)


mitchmindtree:
Closing in favour of #179.

PR #182, opened 2017-11-02
[coreaudio] Remove intermediate buffer

### LINKED ISSUE #181: [CoreAudio] Remove vec allocation per callback
The render callback passed to the audio unit currently has to allocate a new Vec buffer every time it's called, because cpal currently only exposes a single interleaved buffer of samples, but the callback may receive multiple buffers, one per channel (non-interleaved). The vec buffer is resampled into the individual callback channel buffers, which is a potentially inefficient per-sample interleaved copy.

The Rust CoreAudio binding does not currently allow you to specify the necessary parts of the `AudioStreamBasicDescription` struct in `set_stream_format` which would make the callback receive a single interleaved buffer for the channels. An interleaved incoming buffer would be passed directly to the cpal callback with no pre or post processing required, but without the necessary Rusty interface, we need to go lower into the sys bindings.

In order to specify this without modifying coreaudio-rs upstream, you would need to create the `AudioStreamBasicDescription` struct directly and use `set_property` instead. This would open the door for directly supporting other stream formats as well.

I would assume that zero allocations, zero sample processing is a design goal for the library and any muxing or resampling would be deferred to user code, so this is probably a bug.

This is a prerequisite for #180 (at least, to make it efficient).

HybridEidolon:
- Set the audio unit properties directly to make CoreAudio create an interleaved 2-channel buffer which can be passed directly to the user callback.
- Use `slice::from_raw_parts_mut` to produce a direct mutable slice of the OS buffer for the cpal callback to write to.
- Clean up some compile warnings.

No consideration of the stream format requested, this only supports f32 for now.

Fixes #181

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 - Changed the emscripten backend to consume less CPU.
 - Added improvements to the crate documentation.
+- Reduced the number of allocations in the CoreAudio backend.
 
 # Version 0.5.1 (2017-10-21)
 
--- a/src/coreaudio/enumerate.rs
+++ b/src/coreaudio/enumerate.rs
@@ -1,6 +1,5 @@
 use super::Endpoint;
 
-use Format;
 use SupportedFormat;
 
 use std::vec::IntoIter as VecIntoIter;
--- a/src/coreaudio/mod.rs
+++ b/src/coreaudio/mod.rs
@@ -14,9 +14,18 @@ use std::mem;
 use std::sync::{Arc, Mutex};
 use std::thread;
 use std::time::Duration;
+use std::slice;
 
-use self::coreaudio::audio_unit::AudioUnit;
+use self::coreaudio::audio_unit::{AudioUnit, Scope, Element};
 use self::coreaudio::audio_unit::render_callback::{self, data};
+use self::coreaudio::bindings::audio_unit::{
+    AudioStreamBasicDescription,
+    AudioBuffer,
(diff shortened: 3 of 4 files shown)


tomaka:
The code looks good. Is there a possibility that interleaved is not supported on some machines?

HybridEidolon:
I'm not actually certain, and CoreAudio's documentation is so bare that the only way to know is to have hardware to test with. There's no indication that CoreAudio may change your buffers to fit some arbitrary hardware need on the DefaultOutput, either. I can't see that being the case on DefaultOutput or SystemOutput, but it may be an issue once device enumeration is implemented.

macOS devices are generally pretty stable when it comes to these things, but I could see iOS devices not allowing interleaved buffers on RemoteIO to fit some fast path. I don't have any iOS devices to test with, unfortunately.

HybridEidolon:
Further research suggests that if the AU won't allow the stream format specified (e.g. interleaved stream), then it will return `kAudioUnitErr_FormatNotSupported`, which in this code will just Err out while building the voice.

Seeing as how someone already added iOS support, and the canonical sample format for iOS is 16-bit integers on input/output, I figure both RemoteIO and DefaultOutput do some conversion internally anyway, which means interleaved should work on iOS as well. But for HalOutput, it may be a different situation.

To make sure that VoiceIds aren't leaked in the ID slab vec while attempting to build the voice, I am going to push another commit reordering the operations a bit to fail before that.

PR #184, opened 2017-11-03
[coreaudio] Start voices on run

HybridEidolon:
This ensures that audio units aren't started until `EventLoop::run` is called, mirroring the behavior of the other backends.


tomaka:
A user could decide to pause a voice before calling `run`.

HybridEidolon:
I fixed the PR to allow pause/play to be called on the voices before `run`. It's a bit messy but it should work for now.

mitchmindtree:
Thanks for the PR @HybridEidolon! I'm going to close this now that #354 is merged and `run` is no longer a thing.

PR #186, opened 2017-11-06
Fix macos build

chyvonomys:
This PR fixes broken osx build:
1. Fixes tomaka/rodio#145 (callback sync error)
2. Fixes missing `convert_error` caused by most recent PRs. #182 #183

### DIFF
--- a/src/coreaudio/mod.rs
+++ b/src/coreaudio/mod.rs
@@ -98,9 +98,9 @@ impl EventLoop {
 
     #[inline]
     pub fn run<F>(&self, mut callback: F) -> !
-        where F: FnMut(VoiceId, UnknownTypeBuffer)
+        where F: FnMut(VoiceId, UnknownTypeBuffer) + Send
     {
-        let callback: &mut FnMut(VoiceId, UnknownTypeBuffer) = &mut callback;
+        let callback: &mut (FnMut(VoiceId, UnknownTypeBuffer) + Send) = &mut callback;
         self.active_callbacks
             .callbacks
             .lock()
--- a/src/coreaudio/mod.rs
+++ b/src/coreaudio/mod.rs
@@ -151,7 +151,7 @@ impl EventLoop {
             Scope::Input,
             Element::Output,
             Some(&asbd)
-        ).map_err(convert_error)?;
+        )?;
 
         // Determine the future ID of the voice.
         let mut voices_lock = self.voices.lock().unwrap();
--- a/src/coreaudio/mod.rs
+++ b/src/coreaudio/mod.rs
@@ -65,7 +65,7 @@ pub struct EventLoop {
 
 struct ActiveCallbacks {
     // Whenever the `run()` method is called with a callback, this callback is put in this list.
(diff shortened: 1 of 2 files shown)


src/coreaudio/mod.rs
```diff
@@ -65,7 +65,7 @@ pub struct EventLoop {
 
 struct ActiveCallbacks {
     // Whenever the `run()` method is called with a callback, this callback is put in this list.
-    callbacks: Mutex<Vec<&'static mut FnMut(VoiceId, UnknownTypeBuffer)>>,
+    callbacks: Mutex<Vec<Box<FnMut(VoiceId, UnknownTypeBuffer) + Send>>>,
```
tomaka: That's not good. The `&'static mut` has to stay.
Otherwise we have undefined behaviour.

chyvonomys: Could you please explain? Is this a reason of UB: https://github.com/chyvonomys/cpal/blob/632add7b21dd40e2d7b606f9b7cfe7a6813941cf/src/coreaudio/mod.rs#L108

tomaka: Yes. You're transmuting a reference into an owned box. Destrying the `Box` will trigger the UB, so I guess it's fine as long as we don't destroy it. But in any case I don't see a reason for your change.

chyvonomys: The reason is to fix build error (currently cpal does not build on osx). Compiler complains about those callbacks not implementing `Send` (see tomaka/rodio#145 for details). I'm a beginner in rust, and using Box + Send is what I came up with. Not sure how to fix this properly.
This is blocker for me because I use rodio from repo and it currently does not build because of this issue in cpal.

chyvonomys: Is it crucial to use that unsafe transmute there?

tomaka: You can simply use `&'static mut FnMut(VoiceId, UnknownTypeBuffer) + Send` and it should work, I think.

tomaka:
Oh wait, I see now. Unfortunately I think the proper fix is to require the `F` we pass to `run()` to be `Send`, which is a breaking change. It's deeper than just OSX.

chyvonomys:
Let's merge this to make it build at least :)

HybridEidolon:
Oof. Yeah, the Send attribute is necessary on callbacks in any scenario because the callback may be passed to background threads if the backend only allows for that. There isn't a choice otherwise in AudioUnit, because the framework handles the processing graph itself in the background.

I'll concur on putting a `Send` constraint on the callback in `run`.

tomaka:
I merged #189 but the problem of `Send` still exists.

tomaka:
Thanks!

chyvonomys:
I also tried avoiding this `mem::transmute` piece, by storing Boxed callback closures and adding lifetime parameter to them, but eventually that it is not compatible with AudioUnit's `set_render_callback` which requires callback to be `'static`. This seems incorrect to me, e.g. in beep example we build closure that is not `'static`.

tomaka:
The staticness is not a problem because we run the events loop forever. Since we borrow the object forever, it's as if the borrow was static.

PR #189, opened 2017-12-05
Fix coreaudio build

imipolexg:
The last commit to coreaudio/mod.rs added a call to convert_error, but
that function was removed earlier. Remove the call to fix the build.


tomaka:
Thanks

PR #190, opened 2017-12-11
Publish v0.6.0

mitchmindtree:
There does not seem to be any major API breakage, however the emscripten
and macos backends have been pretty heavily refactored so I thought it
best to bump to 0.6 (rather than 0.5.2) just in case there is any subtle
behavioural breakage. Happy to change this to 0.5.2 though if someone
can confirm there will be no downstream breakage.


tomaka:
Thanks

Issue #194, opened 2018-01-23
[coreaudio] Available Device and Format Enumeration


mitchmindtree:
I'm currently working on implementing device and format enumeration on macos for some commercial work I'm doing atm. Just creating this issue to track progress.

[This post](http://fdiv.net/2008/08/12/nssound-setplaybackdeviceidentifier-coreaudio-output-device-enumeration) from ~10 years ago provides a nice example of this that should be useful.

mitchmindtree:
I'm currently working on generating the remaining bindings (CoreAudio.h, AudioToolbox.h, CoreMidi.h, etc) for `coreaudio-sys` in order to implement this, however I'm currently running into [this bindgen bug](https://github.com/rust-lang-nursery/rust-bindgen/issues/104#issuecomment-360668947) which is making this tricky.

I'm currently tossing up between finding a temporary solution, an old version of rust-bindgen or waiting for a fix on that issue before going ahead with this. Alternatively, I could see if using their recommended `build.rs` script approach works in the meantime, though I'm unsure how it would change this linking issue and it might cause issues for users on macos who don't use cargo.

PR #195, opened 2018-01-23
Implement format handling in `build_voice` macos backend

### LINKED ISSUE #180: [CoreAudio] Support I16 and U16 sample formats
This isn't strictly necessary for support on macOS, but the current implementation only offers F32 sample support. For parity with WASAPI and ALSA, the CoreAudio backend should offer I16 and U16 by setting the underlying Audio Unit's stream format to the appropriate format, and setting the PCM flags as necessary.

mitchmindtree:
This is an attempt at properly handling the given format for the `build_voice` method in the coreaudio backend. Currently the `Endpoint` is still ignored. I don't have a mac on me atm but I managed to test a patch similar to this on one earlier today that ran successfully. If there are any mac users that could test this that would be greatly appreciated!

I'll likely follow this up soon with #194.

This closes #180.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- Implement format handling for macos `build_voice` method.
+
 # Version 0.6.0 (2017-12-11)
 
 - Changed the emscripten backend to consume less CPU.
--- a/src/coreaudio/mod.rs
+++ b/src/coreaudio/mod.rs
@@ -117,7 +117,7 @@ impl EventLoop {
     }
 
     #[inline]
-    pub fn build_voice(&self, _endpoint: &Endpoint, _format: &Format)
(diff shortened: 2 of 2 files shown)


mitchmindtree:
I managed to get access to a MBP, patched the commit and tested this successfully today. Not sure why travis is hanging on the macos build atm, but this MBP compiles and runs both examples nicely. I also tested the `beep.rs` example at 92,100hz to check the different format handling.

The CHANGELOG has been updated.

This should be OK to merge, but happy to wait for another mac user to test if desired.

tomaka:
Travis MacOS is probably stuck into the abysses again, but since this is a MacOS PR it seems like a good idea to actually wait for it to finish.

mitchmindtree:
Looks like travis is happy, seeing as this has been approved I'm going to merge!

Issue #187, opened 2017-11-07
The `ChannelPosition` API should be optional


mitchmindtree:
CPAL Is the first low-level audio API that I've personally come across that requires specifying abstract "channel positions" for particular channels. It gets particularly annoying when you do not know or care about the channel positions, but just want to specify some number of channels.

As an example, I'm currently working on an audio installation with 100+ channels that are laid out throughout an exhibition space. The actual positions of the speakers have no correlation with any of the offered `ChannelPosition`s whatsoever, yet as a user I'm still required to fill a `Vec` with arbitrary positions.

I wonder if we could make the channel positioning API optional somehow? In most cases that I've personally come across I've just wanted to specify a number of channels, though admittedly I've mostly worked on pro-audio stuff where I know what device I'm working with and have manual control over the positioning of each channel.

I'd also be curious as to which backends actually use the `ChannelPosition` info, and if so, how do they use it? I don't believe the macOS backend currently uses this (not sure if it provides an API to do so), and I'm unsure about linux or windows.

tomaka:
Yeah I've been wanting to remove them actually.

Windows more or less uses the channel positions, in the sense that it will return an error if the positions don't match what the hardware expects.

mitchmindtree:
Closed via #198.

PR #198, opened 2018-02-02
Remove unused `ChannelPosition` API (update all backends and examples accordingly)

mitchmindtree:
Related discussion at #187.

This removes the ChannelPosition API from the lib root and updates the
ALSA backend and examples accordingly. The other backends have not yet
been updated.

The windows backend now assumes the channel position order is equal to
the channel position mask order. E.g. channel 0 will always be front
left, channel 1 will always be front right, etc. It seems that this order was
previously enforced anyway and would return a FormatError otherwise, so
I can't imagine this should affect windows users too much. Please let me
know if I'm wrong or the windows fix could be improved!

cc @tomaka

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Unreleased
 
+- Remove unused ChannelPosition API.
 - Implement Endpoint and Format Enumeration for macos.
 - Implement format handling for macos `build_voice` method.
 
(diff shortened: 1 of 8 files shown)


mitchmindtree:
@tomaka this has been rebased onto master and should be ready to review now :+1:

tomaka:
Thanks!

Issue #29, opened 2015-03-01
SamplesRate struct


mitchmindtree:
@tomaka Just curious, do you have some plans for the `SamplesRate` struct? From what I can see it seems like it could just be a type alias instead of a struct? Just thought I'd check!

Also I think `SampleRate` (without the `s`) is the more stereotypical naming choice (or even `SampleHz` to be more precise).

tomaka:
The only purpose of `SamplesRate` is to have stronger typing.

mitchmindtree:
Perhaps then we should either:

1. do the same for both `ChannelsCount` and `BufferSize` or
2. type alias `SamplesRate` instead (like `ChannelsCount` currently is)

The `append_data` method signature seems a little inconsistent otherwise.

yupferris:
> Also I think SampleRate (without the s) is the more stereotypical naming choice (or even SampleHz to be more precise).

I know it's bikeshedding a bit, but +1 to this; it's typically called `SampleRate` and I must admit I got a bit irked inside reading it as it is now :)

mitchmindtree:
Closing this following #199.

PR #199, opened 2018-02-02
Rename `SamplesRate` to `SampleRate` and `ChannelsCount` to `ChannelCount`

mitchmindtree:
Previously mentioned at #29.

This changes the `SamplesRate` and `ChannelsCount` type names to the more idiomatic names `SampleRate` and `ChannelCount`.

The `s` on the end of `Sample` and `Channel` is unnecessary as both `Rate` and `Count` already imply plurality. A quick google will show that the term "sample rate" is ubiquitous throughout existing audio libraries, whereas there are no previous uses of "samples rate" within APIs that I can find.

This also renames the `samples_rate` fields to `sample_rate`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- Rename ChannelsCount to ChannelCount.
+- Rename SamplesRate to SampleRate.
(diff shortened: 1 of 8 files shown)


tomaka:
Thanks

PR #201, opened 2018-02-04
Update to a more general Device and Stream API. Add support for input streams (E.g. microphone). Add default format methods.

### LINKED ISSUE #117: Introduce a `Device` API.
If we do wish to support both input and output as mentioned in #116, it might be a nice idea to replace the current `EndPoint` API with a `Device` API. A `Device` may contain any number of `Input` and/or `Output` channels, perhaps offering methods to enumerate them (i.e. `device.inputs()`, `device.outputs()`) along with providing methods for checking sample rate and format compatibility.

A side-note: At an even higher level, PortAudio provides a way to enumerate `HostAPI`s at runtime (i.e. `DirectSound`/`ASIO`), however we'll probably do better to just feature-gate different backends for now, as is currently the case. I personally haven't seen any DAWs (or ever come across the need for) selecting a different audio backend at runtime, but then again I've almost solely used OS X for audio dev. Would be interesting to get other opinions on this.

### LINKED ISSUE #118: Rename `Voice` to `Stream`?
In my own experience, the term `Voice` is normally associated with higher level concepts. I.e. A polyphonic software music instrument might allow the user to set the max number of `Voice`s that it can perform before a new note would override the oldest.

On the other hand, low-level input/outputs are normally referred to as `Stream`s. See [cubeb](https://github.com/kinetiknz/cubeb/search?utf8=%E2%9C%93&q=cubeb_stream&type=Code), [portaudio](http://portaudio.com/docs/v19-doxydocs/api_overview.html), [libsound.io](https://github.com/andrewrk/libsoundio/search?utf8=%E2%9C%93&q=stream&type=Code), [coreaudio](https://developer.apple.com/library/ios/documentation/MusicAudio/Conceptual/CoreAudioOverview/CoreAudioEssentials/CoreAudioEssentials.html), ASIO stands for audio `Stream` input/output. I'm unsure of any examples in support of `Voice`.

mitchmindtree:
This update prepares for adding input stream support by removing the
`Endpoint` type (which only supports output streams) in favour of a more
general `Device` type which may support any number of input or output
streams. Previously discussed at #117.

The name `Voice` has been replaced with the more ubiquitous name
stream. See #118 for discussion on this.

Also introduces a new `StreamData` type which is now passed to the
`EventLoop::run` callback rather than the `UnknownTypeBuffer`.
`StreamData` allows for passing either `Input` data to be read, or
`Output` data to be written. ~~As of this `PR`, the `Input` variant is just
a unit type and is not yet implemented.~~

All backends have been updated for the API changes, ~~however input
stream functionality remains unimplemented across each backend.
That is, most backend implementations of `default_input_device`,
`supported_input_formats` and `default_input_format` are currently
`unimplemented!();` and panic if called. The main aim was to update
to the new API while retaining existing functionality across all backends.~~

The `beep.rs` and `enumerate.rs` examples and the crate-root
documentation have also been updated for the API changes.

Closes #117.
Closes #118.

cc @tomaka (or anyone following CPAL) this is a pretty big change, it would
be great to get your thoughts/feedback on this! I'm totally open to alternative
approaches, this is my attempt at keeping functionality familiar and similar to
the current API while also allowing for input support. I've began implementing
the input API for the coreaudio backend and will be trying to get the remaining
backends implemented over the next couple days too. I don't mind whether
they are a part of this PR or a follow up PR.

EDIT: The new API is now implemented fully across all backends! Emscripten
still needs updating to handle input streams - I'm hoping someone more
familiar with the emscripten tool-chain can do this in a follow-up PR.

### DIFF
--- a/examples/beep.rs
+++ b/examples/beep.rs
@@ -2,17 +2,10 @@ extern crate cpal;
 
 fn main() {
     let device = cpal::default_output_device().expect("Failed to get default output device");
-    //let format = device.default_output_format().expect("Failed to get default output format");
-    let format = device
-        .supported_output_formats()
-        .unwrap()
-        .next()
-        .expect("Failed to get output stream format")
-        .with_max_sample_rate();
-
+    let format = device.default_output_format().expect("Failed to get default output format");
     let event_loop = cpal::EventLoop::new();
     let stream_id = event_loop.build_output_stream(&device, &format).unwrap();
-    event_loop.play_stream(stream_id);
+    event_loop.play_stream(stream_id.clone());
 
     let sample_rate = format.sample_rate.0 as f32;
     let mut sample_clock = 0f32;
--- a/examples/beep.rs
+++ b/examples/beep.rs
@@ -2,6 +2,7 @@ extern crate cpal;
 
 fn main() {
     let device = cpal::default_output_device().expect("Failed to get default output device");
+    //let format = device.default_output_format().expect("Failed to get default output format");
     let format = device
         .supported_output_formats()
         .unwrap()
--- a/examples/beep.rs
+++ b/examples/beep.rs
@@ -1,17 +1,17 @@
 extern crate cpal;
 
 fn main() {
-    let endpoint = cpal::default_endpoint().expect("Failed to get default endpoint");
-    let format = endpoint
-        .supported_formats()
+    let device = cpal::default_output_device().expect("Failed to get default output device");
+    let format = device
+        .supported_output_formats()
         .unwrap()
         .next()
-        .expect("Failed to get endpoint format")
+        .expect("Failed to get output stream format")
         .with_max_sample_rate();
 
     let event_loop = cpal::EventLoop::new();
-    let voice_id = event_loop.build_voice(&endpoint, &format).unwrap();
-    event_loop.play(voice_id);
+    let stream_id = event_loop.build_output_stream(&device, &format).unwrap();
+    event_loop.play_stream(stream_id);
 
     let sample_rate = format.sample_rate.0 as f32;
     let mut sample_clock = 0f32;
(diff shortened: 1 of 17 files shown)


examples/beep.rs
```diff
@@ -1,55 +1,60 @@
...
-                    for out in sample.iter_mut() {
-                        *out = value;
-                    }
-                }
-            },
+    event_loop.run(move |_, data| {
+        match data {
+            cpal::StreamData::Output { buffer } => {
+                match buffer {
+                    cpal::UnknownTypeBuffer::U16(mut buffer) => {
```
tomaka: Style: this could be merged with the above `match` (even if that means duplicating the `StreamData::Output` variant).

src/alsa/enumerate.rs
```diff
@@ -7,21 +7,21 @@ use std::ffi::CStr;
 use std::ffi::CString;
 use std::mem;
 
-/// ALSA implementation for `EndpointsIterator`.
-pub struct EndpointsIterator {
+/// ALSA implementation for `Devices`.
+pub struct Devices {
```
tomaka, line 8: Should be `DevicesIter` IMO.

tomaka:
Thanks! The change is positive. We should probably make sure that the various backends have a working implementation before we ship a version on crates.io.

Please also add entries in `CHANGELOG.md`!

src/alsa/enumerate.rs
mitchmindtree: Hmmm I tend to avoid adding the `Iter` suffix unless I'm trying to avoid ambiguity between a collection and its iterator. IMO `Devices` is fine in the same way that `std::str::Chars` is OK.

mitchmindtree:
OK, input streams are now be supported across coreaudio, alsa and wasapi! This includes the `Device::default_{input/output}_format`, `Device::supported_input_formats` and `Endpoint::build_input_stream` methods.

I've fixed the `Device::name` method on windows (it now uses `FriendlyName` rather than the GUID).

I've updated the enumerate example to display both default input/output devices as well as default input/output formats and supported input/output formats for all devices.

I've also added a new `record_wav.rs` example that records three seconds of audio from the system's default input device and writes this audio to a wav at `$CARGO_MANIFEST_DIR/recorded.wav`.

All three examples work on the macos (10.10), linux and (two) windows 10 systems I've been using for testing, but more testing is of welcome of course!

I've also updated the CHANGELOG to account for all changes in this PR.

This PR should now be ready for review!

mitchmindtree:
@freesig could you test this branch (run each of the examples) on your mac and the windows machine while you have it? Also maybe let us know what mac version you're running (i've only tested on 10.10).

freesig:
@mitchmindtree It ran fine on the mac(10.13.3)
Everything worked fine on the windows except record_wav

> thread 'main' panicked at 'Failed to get default input device', src\libcore\option.rs:874:4
note: Run with `RUST_BACKTRACE=1` for a backtrace.
error: process didn't exit successfully: `target\release\examples\record_wav.exe` (exit code: 101)

mitchmindtree:
Ahh right, I guess you just don't have any input audio devices on the windows machine then (you can check this with the `enumerate` example - it should show None for default input device)? Do you have a mic or audio interface that you could use to test it with? All good if not, that behaviour is expected for that example if there's no inputs to record with.

freesig:
Yep record_wav works if I just plug a lead into the computers input.

mitchmindtree:
Nice one, thanks for that!

src/wasapi/stream.rs
```diff
@@ -0,0 +1,789 @@
...
+                        (*audio_client).Release();
+                        panic!("{:?}", e);
+                    },
+                    Ok(()) => (),
+                };
+
+                format_attempt.Format
+            };
+
+            /*
```
tomaka: Please remove

mitchmindtree:
Good spot! Fixed.

PR #202, opened 2018-02-07
Update winapi version from 0.2 to 0.3

mitchmindtree:
Adds only the necessary cargo features to reduce compile time and reduce
the chance of linking errors occurring for unused libraries (e.g.
d3d12.dll fails to link on my win10 VM).

I thought I'd try and land this before before working on the wasapi
backend implementation for #201.

Tested both beep.rs and enumerate.rs and they work fine with the update.


tomaka:
Thanks!

PR #211, opened 2018-02-17
Add an example that demonstrates feeding input directly to output with some latency

mitchmindtree:
I created this example to test:

1. the latency required on my system when feeding the defualt input directly back into the default output and
2. that stream IDs are actually unique between different streams and that they match the yielded StreamData.

Just thought I'd post in case you'd like to include it as another example.


mitchmindtree:
Going to merge this as it's not immediately obvious how to feed the input into the output without an example.

PR #216, opened 2018-04-03
Fixed alsa: Device not available not returning Err

jojolepro:
A step towards not crashing when alsa is not running or when I don't have any audio devices plugged in :P

Side note: Rodio is running on cpal 0.6 to which I can't find the source code, so it would need to be updated to the newer cpal to fix the bug we are having with amethyst.

### DIFF
--- a/src/alsa/mod.rs
+++ b/src/alsa/mod.rs
@@ -92,7 +92,7 @@ impl Device {
         ) {
             -2 |
             -16 /* determined empirically */ => return Err(FormatsEnumerationError::DeviceNotAvailable),
-            e => check_errors(e).expect("device not available")
+            e => if check_errors(e).is_err(){ return Err(FormatsEnumerationError::DeviceNotAvailable) }
         }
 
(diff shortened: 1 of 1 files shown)


tomaka:
I'm conflicted about this fix. The `-16` case just above your code is what is supposed to happen when a device is not available.

We don't want to return `DeviceNotSupported` when a different problem happens, but on the other hand it's entirely possible that there are values other than `-16` that correspond to a non-available device.

jojolepro:
The error I'm having normally says "device not available". So returning devicenotsupported is wrong, as you said.

tomaka:
What I mean is that the error value you are receiving from ALSA probably indicates an unavailable device, but other error codes probably indicate other kinds of errors.
Therefore we shouldn't treat them in the same way without a distinction.

derekdreery:
Could you have an `Unknown` error that's a catch-all. Then you can always handle specific errors later.

mitchmindtree:
Closing in favour of #248.

PR #219, opened 2018-04-18
Add missing braking changes introduced into 0.7

frazar:
In the process of porting a library from `cpal` 0.6 to 0.7, I noticed some breaking changes were not mentioned in the changelog.


tomaka:
Thanks!

Issue #204, opened 2018-02-12
Provide a `Host` API? (A path towards supporting alt-backends like ASIO, Windows Exclusive Mode, PulseAudio, Jack)


mitchmindtree:
PortAudio (a C library with similar goals to CPAL) provide what they call a `Host` API for choosing between different audio APIs that might exist on a single system.

Any operating system may have any number of `Host` APIs to the system's audio devices. E.g. on Linux this might include `Alsa`, `Jack` and `PulseAudio`. In the case of Windows, perhaps `Asio`, `Shared` and `Exclusive` hosts could be included. In the case of macOS and iOS, `CoreAudio` would likey be the only host in most cases.

I imagine general use might look something like this:

```rust
let host = cpal::default_host();
let device = host.default_output_device().unwrap();
// etc
```

Selecting a specific host might look something like this:

```rust
let host = cpal::hosts()
    .find(|host| host.name().contains("jack"))
    .expect("cannot find jack audio backend");
let device = host.default_input_device().unwrap();
// etc
```

This way a low-latency application could attempt to select low-latency supporting hosts with something like this:

```rust
let host = if cfg!(target_os = "linux") {
    cpal::hosts().find(|h| h.name().contains("jack")).unwrap_or_else(cpal::default_host)
} else if cfg!(target_os = "windows") {
    cpal::hosts().find(|h| h.name().contains("asio")).unwrap_or_else(cpal::default_host)
} else {
    cpal::default_host()
};
```

mitchmindtree:
It's looking like my current project may need support for ASIO soon. @tomaka it would be great to get some feedback from you on this API as I'd like to go ahead and begin implementing it soon (along with the ASIO backend) if you agree it sounds like a decent approach?

tomaka:
I'm really not a fan of this. The "host" is supposed to be cpal because cpal is supposed to be as low-level as possible.

I'm ok with adding platform-specific methods that allow choosing a backend though.

mitchmindtree:
OK I see where you're coming from, I'll refer to winit's approach when attempting to implement this.

mitchmindtree:
OK, I've been reflecting a lot on this over the past year and a half since opening this issue and I'm convinced that providing some sort of Hosts API is the correct approach for CPAL.

> The "host" is supposed to be cpal because cpal is supposed to be as low-level as possible.

I totally agree that CPAL should be as low-level as possible, however I don't think it is possible for CPAL to be "the" host (singular) unless two conditions are met:

[...]

PR #221, opened 2018-04-22
Asio Backend Support

freesig:
# Support for ASIO backend on windows
## Current state
This is a very early stage attempt. There's lots to do __before__ merging. I just want to start getting some feedback.

__Feedback:__
I'm fairly new to Rust so there's definitely going to be mistakes / non-idiomatic code in here.
Feel free to point out anything you think is weird / wrong or ask for explanations.

__Things that are working:__
- The enumerate example 
- The beep example
- Building the ASIO library and linking to it
- Creating the bindings to the library with bindgen

__Things that still need to be done (non-exhaustive):__
- [x] The record_wav example
- [ ] Supporting all the possible ASIO formats
- [x] Allow the CPAL user to set if the want ASIO or WASAPI at with an initialize call
- [x] Check for potential threading issue _(I'm new to this)_
- [ ] Stop any panics that could cross FFI boundaries
- [x] Add comments and example usage docs
- [x] Make the interleaving faster and avoid copying

## Usage
1. You need to download the __[ASIO SDK](https://www.steinberg.net/en/company/developers.html)__ and unzip it somewhere.
2. Put the full path into the environment variable CPAL_ASIO_DIR
eg. with powershell ($env:CPAL_ASIO_DIR="C:\Users\you\my_libs\ASIOSDK2.3")
3. Now you should be able to run "cargo run --example enumerate"
__Note:__
The beep example is currently hard set to select [ASIO 4 All](http://www.asio4all.org/) because there's no way to get the default asio driver and this made testing easier.

## Resources
- __[This](http://www.grame.fr/ressources/publications/Pa_ASIO.pdf)__ is a good write up about how portaudio built there asio backend.
- Theres a pdf in the ASIO sdk called _ASIO SDK 2.3.pdf_ that explains the how ASIO works
- Theres an example of how to write an ASIO host in c++ in ASIOSDK2.3/host/sample/hostsample.cpp

__Note:__ I had to change branches and make this pr because of conflict problems.


.gitignore
```diff
@@ -3,3 +3,10 @@
 .cargo/
 .DS_Store
 recorded.wav
+asio-sys/target
+asio-sys/Cargo.lock
+asio-sys/.cargo/
+asio-sys/.DS_Store
+*~
+*.swap
+*.swo
```
mitchmindtree: I think git allows you to have a personal, custom global `.gitignore` which allows you to not have to add in things like editor-specific extensions like these to the `.gitignore` of the repositories you contribute to. This can be handy as some maintainers prefer not to allow for items in the `.gitignore` that aren't specific to the project itself (I personally don't really care, just thought I'd add!).

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "cpal"
-version = "0.8.1"
+version = "0.7.0"
```
mitchmindtree: Looks like you may have accidentally decremented the version here - maybe copy paste?

Cargo.toml
```diff
@@ -10,10 +10,14 @@ keywords = ["audio", "sound"]
 
 [dependencies]
 lazy_static = "0.2"
+itertools = "0.7.8"
```
mitchmindtree: It's best to avoid dependencies that aren't totally necessary in super low-level crates like this, mainly just because some folks are compiling on super slow devices (like Pi's, etc) where each crate can add a lot of time onto the first build. At least in this case it should be pretty easy to work around this one.

asio-sys/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "asio-sys"
+version = "0.1.0"
+authors = ["Tom Gowan <s3549186@student.rmit.edu.au>"]
```
mitchmindtree: Maybe a more personal email might be better for the long term?

examples/enumerate.rs
```diff
@@ -1,15 +1,19 @@
...
-    println!("Default Input Device:\n  {:?}", cpal::default_input_device().map(|e| e.name()));
-    println!("Default Output Device:\n  {:?}", cpal::default_output_device().map(|e| e.name()));
+    println!(
+        "Default Input Device:\n  {:?}",
+        cpal::default_input_device().map(|e| e.name())
+    );
+    println!(
+        "Default Output Device:\n  {:?}",
+        cpal::default_output_device().map(|e| e.name())
+    );
```
mitchmindtree, line 11: Normally it's best to save formatting changes for a separate PR as it's easy to cause conflicts and can make it a bit harder to review the actual changes, that said it's easy to accidentally add these changes when running rustfmt (especially if it runs on save). Arguably CPAL should already be formatted, but it's been around quite a bit longer than rustfmt has.

src/os/windows/mod.rs
```diff
@@ -0,0 +1,12 @@
+pub enum Backend {
+    Wasapi,
+    Asio,
+}
+
+// TODO This needs to be set once at run time
+// by the cpal user
+static backend: Backend = Backend::Asio;
```
mitchmindtree: If this was behind a `Mutex`, it could probably be safely set during runtime via something like `cpal::os::windows::initialise_asio()`, e.g.:

```rust
extern crate cpal;

fn main() {
    // Only generate code for this block if we're targeting windows.
    #[cfg(target_os = "windows")]
    {
        if cpal::os::windows::is_asio_supported() {
            cpal::os::windows::initialise_asio().unwrap();
        }
    }
    // All following cpal function calls and items will now use the `asio` module internally now rather than `wasapi`.
    let event_loop = cpal::EventLoop::new();
}
```

That said, it would probably also be possible for a user to call this while an EventLoop is already running which I imagine would be highly unsafe. I wonder if there's a nice way to check if an `EventLoop` is already running and panic if so. Or perhaps this could be done via a special `EventLoop` constructor? This would still allow a user to call things like `default_format/device` etc though before hand, in turn retrieving WDM devices and attempting to use them with an ASIO EventLoop.

@tomaka it would be great to get your input on how you would like to handle the os-specific functionality. In this case, ASIO is an entirely independent event loop that doesn't even use the WDM API. This is partly why I recommended a `Host` API in #204 so that we could use types to ensure conflicts like this don't occur. We will likely run into the same issues when attempting to add support for Jack and PulseAudio.

src/platform/windows/asio/device.rs
```diff
@@ -0,0 +1,212 @@
...
+                sample_rate: SampleRate(sample_rate.rate), 
+                data_type: format.data_type}
+            },
+            Err(e) => {
+                println!("Error retrieving sample rate: {}", e);
+                format
+            },
+        };
+        
+        let format = match sys::get_data_type(&self.driver_name) {
```
mitchmindtree: It might be nicer to just get the channels, sample rate and data type independently and *then* construct the `Format`, rather than constructing the format multiple times with redundant fields.

src/platform/windows/asio/stream.rs
```diff
@@ -0,0 +1,334 @@
...
+    asio_stream: Arc<Mutex<Option<sys::AsioStream>>>,
+    stream_count: Cell<usize>,
+    callbacks: Arc<Mutex<Vec<&'static mut (FnMut(StreamId, StreamData) + Send)>>>,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+pub struct StreamId(usize);
+
+pub struct InputBuffer<'a, T: 'a> {
+    buffer: &'a mut [T],
```
mitchmindtree: This shouldn't need `mut` as the input stream buffer is read-only.

src/platform/windows/asio/stream.rs
```diff
@@ -0,0 +1,334 @@
...
+}
+pub struct OutputBuffer<'a, T: 'a> {
+    buffer: &'a mut [T],
+}
+
+impl EventLoop {
+    pub fn new() -> EventLoop {
+        EventLoop {
+            asio_stream: Arc::new(Mutex::new(None)),
+            stream_count: Cell::new(0),
```
mitchmindtree: The `EventLoop` might already not be sync, but just thought I'd mention that I think it is a little unidiomatic to have private fields that are both `Sync` (i.e. Arc<Mutex<T>> for safely sharing between threads) and fields that are not (`Cell` in this case is not threadsafe). A good alternative in this case might be to use `Arc<AtomicUsize>` which can be safely updated across threads via it's atomic `store`, `load`, etc operations.

All this said, it's probably not a good idea to share the `EventLoop` between multiple threads anyway, but this is probably not the place for that discussion.

src/platform/windows/asio/stream.rs
```diff
@@ -0,0 +1,334 @@
...
+    }
+
+    pub fn build_input_stream(
+        &self,
+        device: &Device,
+        format: &Format,
+        ) -> Result<StreamId, CreationError> {
+        let stream_type = sys::get_data_type(&device.driver_name).expect("Couldn't load data type");
+        match sys::prepare_stream(&device.driver_name) {
+            Ok(stream) => {
```
mitchmindtree: To avoid the epically deep nesting here, it can sometimes be nicer to assign the result of the match expression to a variable, e.g this could be:
```rust
let stream = match sys::prepare_stream(&device.driver_name) {
    Ok(stream) => stream,
    Err(err) => return Err(CreationError::DeviceNotAvailable), 
};
```

src/platform/windows/asio/stream.rs
```diff
@@ -0,0 +1,334 @@
...
+                {
+                    *self.asio_stream.lock().unwrap() = Some(stream);
+                }
+                let count = self.stream_count.get();
+                self.stream_count.set(count + 1);
+                let asio_stream = self.asio_stream.clone();
+                let callbacks = self.callbacks.clone();
+                let bytes_per_channel = format.data_type.sample_size();
+                let num_channels = format.channels.clone();
```
mitchmindtree: Is it possible to validate each of the format fields to ensure they are supported by the ASIO device and set within the ASIO device?

src/platform/windows/asio/stream.rs
mitchmindtree: Something similar could be done for some of the following match exprs too.

src/platform/windows/asio/stream.rs
```diff
@@ -0,0 +1,334 @@
...
+                                            let length = channels[0].len();
+                                            for i in 0..length{
+                                                for channel in &channels{
+                                                    buffer.push(channel[i]);
+                                                }
+                                            }
+                                            buffer
+                                        }
+
+                                        let mut channels: Vec<Vec<$SampleType>> = vec![Vec::new(); num_channels as usize];
```
mitchmindtree: Normally it's best to avoid dynamically allocating on the audio thread itself where possible as allocation normally takes a non-deterministic amount of time and can cause glitching on some systems. Perhaps this can be stored within the event loop so that it may be re-used between calls?

src/platform/windows/asio/stream.rs
```diff
@@ -0,0 +1,334 @@
...
+                                     $AsioType:ty,
+                                     $AsioTypeIdent:ident) => {
+                                        // Buffer that is filled by cpal.
+                                        let mut cpal_buffer: Vec<$SampleType> = vec![0 as $SampleType; cpal_num_samples];
+                                        //  Call in block because of mut borrow
+                                        {
+                                        }
+                                        // Function for deinterleaving because
+                                        // cpal writes to buffer interleaved
+                                        fn interleave(channels: Vec<Vec<$SampleType>>) -> Vec<$SampleType>{
```
mitchmindtree: It might be sightly more efficient (and avoid requiring dynamically allocating) to allow for re-using the interleaved-samples buffer, maybe something like this:

```rust
fn interleave(channels: &[Vec<$SampleType>], interleaved: &mut Vec<$SampleType>) {
    ...
}
```

src/platform/windows/asio/stream.rs
```diff
@@ -0,0 +1,334 @@
...
+                                    sys::AsioSampleType::ASIOSTInt32LSB => {
+                                        try_callback!(I16, i16, i16, i32, i32);
+                                    }
+                                    sys::AsioSampleType::ASIOSTInt16LSB => {
+                                        try_callback!(I16, i16, i16, i16, i16);
+                                    }
+                                    sys::AsioSampleType::ASIOSTFloat32LSB => {
+                                        try_callback!(F32, f32, f32, f32, f32);
+                                    }
+                                    sys::AsioSampleType::ASIOSTFloat32LSB => {
```
mitchmindtree: Should this be `ASIOSTFloat64LSB` instead? It looks like this branch condition might be identical to the one above, but my eyes could be missing something

src/platform/windows/asio/stream.rs
```diff
@@ -0,0 +1,334 @@
...
+                                                .offset(asio_stream.buffer_size as isize * i as isize);
+                                            let asio_buffer: &'static mut [$AsioType] =
+                                                std::slice::from_raw_parts_mut(
+                                                    buff_ptr,
+                                                    asio_stream.buffer_size as usize);
+                                            for asio_s in asio_buffer.iter(){
+                                                channel.push( (*asio_s as i64 *
+                                                               ::std::$SampleTypeIdent::MAX as i64 /
+                                                               ::std::$AsioTypeIdent::MAX as i64) as $SampleType);
+                                            }
```
mitchmindtree: I don't think this sample format conversion is correct for the float types, as PCM floats are normally bound from `-1.0` to `1.0` rather than their max value. I could be wrong though and it might be that ASIO does this differently, though I doubt they would.

src/platform/windows/asio/stream.rs
```diff
@@ -0,0 +1,334 @@
...
+                                                                target: Some(super::super::OutputBuffer::Asio(buff))
+                                                            })
+                                            }
+                                            );
+                                    }
+                                    // Function for deinterleaving because
+                                    // cpal writes to buffer interleaved
+                                    fn deinterleave(data_slice: &mut [$SampleType],
+                                                    num_channels: usize) -> Vec<Vec<$SampleType>>{
+                                        let mut channels: Vec<Vec<$SampleType>> = Vec::new();
```
mitchmindtree: As mentioned on the `interleave` function above, it might be more efficient to allow for re-using the one `channels` Vec via a `&mut` argument rather than re-allocating each time this is called.

src/platform/windows/asio/stream.rs
```diff
@@ -0,0 +1,334 @@
...
+                    match callbacks.first_mut() {
+                        Some(callback) => {
+                            macro_rules! try_callback {
+                                ($SampleFormat:ident,
+                                 $SampleType:ty,
+                                 $SampleTypeIdent:ident,
+                                 $AsioType:ty,
+                                 $AsioTypeIdent:ident) => {
+                                    // Buffer that is filled by cpal.
+                                    let mut cpal_buffer: Vec<$SampleType> = vec![0 as $SampleType; cpal_num_samples];
```
mitchmindtree: As above, it could be more efficient to store a buffer within the EventsLoop and re-use it.

src/platform/windows/asio/stream.rs
```diff
@@ -0,0 +1,334 @@
...
+                                                buffer: UnknownTypeOutputBuffer::$SampleFormat(
+                                                            ::OutputBuffer{
+                                                                target: Some(super::super::OutputBuffer::Asio(buff))
+                                                            })
+                                            }
+                                            );
+                                    }
+                                    // Function for deinterleaving because
+                                    // cpal writes to buffer interleaved
+                                    fn deinterleave(data_slice: &mut [$SampleType],
```
mitchmindtree: This data slice is only read from and never written too within this function body and in turn could just be `&` rather than `&mut`.

src/platform/windows/asio/stream.rs
```diff
@@ -0,0 +1,334 @@
...
+                                            .offset(asio_stream.buffer_size as isize * i as isize);
+                                        let asio_buffer: &'static mut [$AsioType] =
+                                            std::slice::from_raw_parts_mut(
+                                                buff_ptr,
+                                                asio_stream.buffer_size as usize);
+                                        for (asio_s, cpal_s) in asio_buffer.iter_mut()
+                                            .zip(&channel){
+                                                *asio_s = (*cpal_s as i64 *
+                                                           ::std::$AsioTypeIdent::MAX as i64 /
+                                                           ::std::$SampleTypeIdent::MAX as i64) as $AsioType;
```
mitchmindtree: I think this sample format conversion might have the same issue I mentioned above in the input stream callback.

src/platform/windows/asio/stream.rs
```diff
@@ -0,0 +1,334 @@
...
+        }
+    }
+}
+
+pub fn play_stream(&self, stream: StreamId) {
+    sys::play();
+}
+
+pub fn pause_stream(&self, stream: StreamId) {
+    sys::stop();
```
mitchmindtree: It looks like these `start` and `stop` functions would start/stop *all* streams rather than just the stream with the given Id. Does ASIO let you play/pause individual streams? If not, perhaps we could have a flag for each stream which indicates whether or not it is playing and then check this within the callback to determine whether the user callback should be called? @tomaka thoughts?

src/platform/windows/asio/stream.rs
```diff
@@ -0,0 +1,334 @@
...
+}
+pub fn run<F>(&self, mut callback: F) -> !
+where
+F: FnMut(StreamId, StreamData) + Send,
+{
+    let callback: &mut (FnMut(StreamId, StreamData) + Send) = &mut callback;
+    self.callbacks
+        .lock()
+        .unwrap()
+        .push(unsafe { mem::transmute(callback) });
```
mitchmindtree: I know that one of the other backend has this same behaviour, but it might be nice to write a comment on top of this unsafe to explain why the `transmute` is necessary (e.g. to allow for convincing the compiler that the lifetime is static).

src/os/windows/mod.rs
mitchmindtree: Further, what if a user wants to create both wasapi and an ASIO streams?

src/platform/windows/mod.rs
```diff
@@ -0,0 +1,302 @@
...
+        match self {
+            &EventLoop::Wasapi(ref d) => match stream {
+                StreamId::Wasapi(s) => d.play_stream(s),
+                _ => unreachable!(),
+            },
+            &EventLoop::Asio(ref d) => match stream {
+                StreamId::Asio(s) => d.play_stream(s),
+                _ => unreachable!(),
+            },
+        }
```
mitchmindtree, line 219: A more concise way of writing this might be:

```rust
match (self, stream) {
    (&EventLoop::Wasapi(ref d), StreamId::Wasapi(s)) => d.play_stream(s),
    (&EventLoop::Asio(ref d), StreamId::Asio(s)) => d.play_stream(s),
    _ => unreachable!(),
}
```
Same goes for a couple of the below methods.

src/platform/windows/mod.rs
```diff
@@ -0,0 +1,302 @@
...
+                ::UnknownTypeInputBuffer::U16(ob) => ob,
+            }
+        },
+        ::StreamData::Output(b) => b,
+    }
+    match un_buf {
+        ::Un
+    }
+}
+*/
```
mitchmindtree: Can this be removed?

Cargo.toml
freesig: whoops

Cargo.toml
freesig: As in dont use itertools in the asio-sys?

asio-sys/Cargo.toml
freesig: ah this must be from my git.

examples/enumerate.rs
freesig: yeh that's strange. I ran rustfmt on the asio-sys crate but it shouldn't have run on cpal. This might have been from my editor. I haven't actually changed anything in this file. So could probably just overwrite it with the master version

src/os/windows/mod.rs
freesig: I was imagining some type of late init. Where backend can be initialized only once but via a call. It does seem that you can use lazy_static! to initialize via a call.
ie this runs the initialize for that variable and any other asio stuff but can only be run once.
`cpal::os::windows::initialise_asio().unwrap();`

The Kotlin language has a great example of this called [lateinit](https://kotlinlang.org/docs/reference/properties.html#late-initialized-properties-and-variables).
Basically it allows a immutable variable to be initialized later but only once.
Not sure if something similar exists for Rust

src/platform/windows/asio/stream.rs
freesig: Ah yes good point. I think this conversion should probably be a separate macro from the try_callback! I imagine that there needs to be different conversions for int uint and float

src/platform/windows/asio/stream.rs
freesig: I could be wrong but I think this is a limit in ASIO. That you can only have one stream at a time. This might mean we need to combine the input / output into a single stream. Possibly it could be wrapped internally inside the asio-sys crate it still appears from the outside to be "separate streams". I will look into this a bit more.

freesig:
I noticed that this isn't passing tests on Android because of this:

> 130 | use null as platform;
>     |     ---------------- previous import of the module `platform` here
> ...
> 141 | mod platform;
>     | ^^^^^^^^^^^^^ `platform` redefined here

Might need to change the name of the platform module?

Cargo.toml
mitchmindtree: Yep - I don't think it's actually being used anymore anyway is it?

freesig: I'm just using it for step() but there is a step_by in std iter that does the same thing. Only problem is it's nightly

src/platform/windows/asio/device.rs
freesig: Yep I agree. I wasn't sure what behavior it should have here?
If it can't get one of the details ie sample rate should it panic or return a default?

src/platform/windows/asio/stream.rs
```diff
@@ -0,0 +1,333 @@
...
+                                    fn deinterleave(data_slice: &mut [$SampleType],
+                                                    num_channels: usize) -> Vec<Vec<$SampleType>>{
+                                        let mut channels: Vec<Vec<$SampleType>> = Vec::new();
+                                        for i in 0..num_channels{
+                                            let mut it = data_slice.iter().skip(i).cloned();
+                                            let channel = it.step(num_channels).collect();
+                                            channels.push(channel);
+                                        }
+                                        channels
+                                    }
```
mitchmindtree: You should be able to get around the need for `step` with something along these lines:
```rust
let channel_len = data_slice.len() / num_channels;
let mut channels: Vec<_> = (0..num_channels).map(|_| Vec::with_capacity(channel_len)).collect();
for (i, &sample) in data_slice.iter().enumerate() {
    let ch = i % num_channels;
    channels[ch].push(sample);
}
```

src/platform/windows/asio/device.rs
mitchmindtree: I think the best might be to return `Err(DefaultFormatError::StreamTypeNotSupported)`.

src/platform/windows/mod.rs
freesig: Nah it doesn't work because self is taken by ref a while stream_id is moved. Match complains about different bindings in same match.

src/platform/windows/asio/stream.rs
```diff
@@ -50,8 +51,8 @@ impl EventLoop {
                 {
                     *self.asio_stream.lock().unwrap() = Some(stream);
                 }
-                let count = self.stream_count.get();
-                self.stream_count.set(count + 1);
+                let count = self.stream_count.load(Ordering::SeqCst);
+                self.stream_count.store(count + 1, Ordering::SeqCst);
```
mitchmindtree: Just thought I'd mention there's a [`AtomicUsize::fetch_add` method](https://doc.rust-lang.org/std/sync/atomic/struct.AtomicUsize.html#method.fetch_add) that simplifies this slightly.

freesig:
So it doesn't appear that ASIO4All is passing on the correct bit rate when the actual device is changed.
It does detect a the correct rate when a different asio driver (for example dante) is running.
This might be a bug with ASIO4All but seeing as that is the most popular choice for people using asio we might need to investigate this further.

src/platform/windows/asio/stream.rs
freesig: Hey @mitchmindtree I've had a go at removing the buffers and putting them outside the callback but it's not working. I don't think my approach is right.
If you checkout the latest commit on stream.rs for build_input_stream you can see what I'm trying todo using enums to create different buffer types. But I'm not sure this is the right way to go about it. I'm stuck on how to get the buffer out of the enum.

asio-sys/src/lib.rs
```diff
@@ -0,0 +1,550 @@
...
+}
+
+pub struct AsioStream {
+    pub buffer_infos: [AsioBufferInfo; 2],
+    driver: ai::AsioDrivers,
+    pub buffer_size: i32,
+}
+
+// This is a direct copy of the ASIOSampleType
+// inside ASIO SDK.
```
mitchmindtree: Was bindgen unable to automatically generate this?

.gitignore
mitchmindtree: Ping :)

asio-sys/build.bat
```diff
@@ -0,0 +1,2 @@
+@CALL "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\bin\amd64\vcvars64.bat" %*
+@start powershell
```
mitchmindtree: I think we can probably remove this file in favour of [point 7. in the step-by-step guide](https://github.com/freesig/cpal/issues/6#issue-369520527).

asio-sys/src/bindings/errors.rs
```diff
@@ -0,0 +1,140 @@
...
+pub enum AsioError {
+            NoDrivers, 
+            HardwareMalfunction,
+            InvalidInput, 
+            BadMode, 
+            HardwareStuck, 
+            NoRate, 
+            ASE_NoMemory, 
+            UnknownError, 
+}
```
mitchmindtree, line 22: Formatting

asio-sys/src/bindings/mod.rs
```diff
@@ -0,0 +1,884 @@
...
+// Bindings import
+use self::asio_import as ai;
+
+// TODO I dont think this is needed anymore
+/*
+pub struct CbArgs<S, D> {
+    pub stream_id: S,
+    pub data: D,
+}
+*/
```
mitchmindtree: Can you delete this?

asio-sys/src/bindings/mod.rs
```diff
@@ -0,0 +1,884 @@
...
+        let mut my_driver_name = CString::new(driver_name).expect("Can't go from str to CString");
+        let raw = my_driver_name.into_raw();
+        let mut driver_info = ai::ASIODriverInfo {
+            _bindgen_opaque_blob: [0u32; 43],
+        };
+        unsafe {
+            // Destroy old drivers and load new drivers.
+            let load_result = drivers.load(raw);
+            // Take back ownership
+            my_driver_name = CString::from_raw(raw);
```
mitchmindtree: Rather than doing:

1. `CString::into_raw`
2. Pass raw pointer to `drivers.load`
3. Take ownership back with `CString::from_raw`

You should be able to just keep the original `CString` and then do `drivers.load(my_driver_name.as_ptr())`

asio-sys/src/bindings/mod.rs
```diff
@@ -0,0 +1,884 @@
...
+                let mut buffer_info_convert =
+                    mem::transmute::<Vec<AsioBufferInfo>, Vec<ai::ASIOBufferInfo>>(buffer_infos);
+                let mut callbacks_convert =
+                    mem::transmute::<AsioCallbacks, ai::ASIOCallbacks>(callbacks);
+                drivers
+                    .asio_create_buffers(
+                        buffer_info_convert.as_mut_ptr(),
+                        num_channels as i32,
+                        pref_b_size,
+                        &mut callbacks_convert,
```
mitchmindtree: Here I think you might be able to get away without using the `transmute` for the `buffer_infos` above - e.g. you could probably do something like this:

```rust
.asio_create_buffers(
    buffer_infos.as_mut_ptr() as *mut _,
    num_channels as i32,
    pref_b_size,
    &mut callbacks_convert,
)
```

The reason I mention this is that `transmute` doesn't do much for us in terms of checking sizes here, as a call to `mem::size_of::<Vec<T>>()` will always be the same for any `Vec` no matter what `T` is as the size of a `Vec` in memory is just the pointer, length and capacity. The result of casting the pointer like I've done above should be the same.

This way you might not have to transmute back to `Vec<AsioBufferInfo>` again below.

asio-sys/src/bindings/mod.rs
```diff
@@ -0,0 +1,884 @@
...
+        println!("start result: {:?}", result);
+    }
+}
+
+/// Stops input and output streams playing
+pub fn stop() {
+    unsafe {
+        // TODO handle result instead of printing
+        let result = get_drivers().asio_stop();
+        println!("stop result: {:?}", result);
```
mitchmindtree: `println!` here and in the play fn above

asio-sys/src/bindings/mod.rs
```diff
@@ -0,0 +1,884 @@
...
+/// happen in the correct state.
+/// TODO it is possible to enforce most of this
+/// at compile time using type parameters.
+/// All calls have results that are converted
+/// to Rust style results.
+impl AsioWrapper {
+    /// Load the driver.
+    /// Unloads the previous driver.
+    /// Sets state to Loaded on success.
+    unsafe fn load(&mut self, raw: *mut i8) -> bool {
```
mitchmindtree: Could be worth renaming `raw: *mut i8` to `name: *mut c_char` for clarity. (`c_char` is in `std::os::raw`)

src/os/mod.rs
```diff
@@ -0,0 +1 @@
+pub mod windows;
```
mitchmindtree, line 7: It could be worth adding a doc comment to the top of this module that explains its purpose, e.g.

"The `cpal::os` module provides operating-system-specific functionality. If you are using this module within a cross-platform project, you may wish to use `cfg(target_os = "<os_name>")` to ensure that you only use the OS-specific items when compiling for that OS."

src/os/windows/mod.rs
```diff
@@ -0,0 +1,31 @@
+use std::sync::Mutex;
+
+#[derive(Clone)]
+pub enum BackEnd {
```
mitchmindtree, line 8: Could be worth adding a doc comment here mentioning that `Wasapi` is the default so that you can tell from reading the docs.

src/platform/windows/asio/asio_utils/mod.rs
```diff
@@ -0,0 +1,43 @@
...
+/// cpal writes to buffer interleaved
+/// cpal: LRLRLRLR
+/// asio: LLLLRRRR
+/// More then stereo:
+/// cpal: 123123123123
+/// asio: 111122223333
+pub fn deinterleave<T>(cpal_buffer: &[T], asio_channels: &mut [Vec<T>])
+where
+    T: Copy,
+{
```
mitchmindtree, line 38: Could be worth asserting that:
```rust
assert!(
    cpal_buffer.len() % asio_channels.len() == 0,
    "the length of the interleaved buffer must be a multiple of the expected number of channels"
);
```
at the start of the function?

src/platform/windows/asio/device.rs
```diff
@@ -0,0 +1,195 @@
...
+    }
+}
+
+impl Eq for Device {}
+
+impl Hash for Device {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.name.hash(state);
+    }
+}
```
mitchmindtree, line 40: It's probably fine to leave out the `PartialEq`, `Eq` and `Hash` impls in favour of the user of the type using `device.name` as it's maybe a bit clearer what's going on?

src/platform/windows/asio/stream.rs
```diff
@@ -0,0 +1,908 @@
...
+pub struct EventLoop {
+    /// The input and output ASIO streams
+    asio_streams: Arc<Mutex<sys::AsioStreams>>,
+    /// List of all CPAL streams
+    cpal_streams: Arc<Mutex<Vec<Option<Stream>>>>,
+    /// Total stream count
+    stream_count: AtomicUsize,
+    /// The CPAL callback that the user gives to fill the buffers.
+    /// TODO This should probably not be in a Vec as there can only be one
+    callbacks: Arc<Mutex<Vec<&'static mut (FnMut(StreamId, StreamData) + Send)>>>,
```
mitchmindtree: Yeah I think the TODO on this makes sense

src/platform/windows/asio/stream.rs
```diff
@@ -0,0 +1,908 @@
...
+        // If they try and use and unavailable rate then panic
+        let sample_rate = sample_rate.0;
+        let ref mut streams = *self.asio_streams.lock().unwrap();
+        if sample_rate != drivers.get_sample_rate().rate {
+            if drivers.can_sample_rate(sample_rate) {
+                drivers
+                    .set_sample_rate(sample_rate)
+                    .expect("Unsupported sample rate");
+            } else {
+                panic!("This sample rate {:?} is not supported", sample_rate);
```
mitchmindtree: I think there is a `FormatNotSupported` variant of the `CreationError` enum that could be returned here instead of `panic!`ing.

src/platform/windows/asio/stream.rs
```diff
@@ -0,0 +1,908 @@
...
+                // Set the input callback.
+                // This is most performance critical part of the ASIO bindings.
+                sys::set_callback(move |index| unsafe {
+                    // if not playing return early
+                    // TODO is this lock necessary
+                    {
+                        if let Some(s) = cpal_streams.lock().unwrap().get(count - 1) {
+                            if let Some(s) = s {
+                                if !s.playing {
+                                    return ();
```
mitchmindtree: Why do we only check one of the `cpal_streams` here when deciding whether or not we read or write from any of the streams?

src/platform/windows/asio/stream.rs
```diff
@@ -0,0 +1,908 @@
...
+                                }
+                            }
+                        }
+                    }
+                    // Get the stream
+                    // TODO is this lock necessary
+                    if let Some(ref asio_stream) = asio_streams.lock().unwrap().input {
+                        // Get the callback
+                        // TODO is this lock necessary
+                        let mut callbacks = callbacks.lock().unwrap();
```
mitchmindtree: For these `TODO: is this lock necessary` comments, I think the answer is just "yes if any data needs to be mutated". `EventLoop::run` should only be called once, and I'd imagine that ASIO behind the scenes calls these callbacks one at a time for each stream unless they mention otherwise in the docs so I don't think there should be any major risk of blocking here.

src/platform/windows/asio/stream.rs
```diff
@@ -0,0 +1,908 @@
...
+                                    };
+                                    ($AsioTypeIdent:ident,
+                                    $SampleType:ty,
+                                    $SampleTypeIdent:ident,
+                                    $Sample:expr
+                                ) => {
+                                        (*$Sample as i64 * ::std::$SampleTypeIdent::MAX as i64
+                                            / ::std::$AsioTypeIdent::MAX as i64) as $SampleType
+                                    };
+                                };
```
mitchmindtree: One concern here is that if in the future someone adds a `u32` or `u8` case down below that this `convert_sample!` macro won't behave as expected.

I think we'd actually be better off *only* supporting the sample formats that directly map to supported CPAL sample formats and not doing any conversions at all. This way we're not hiding the cost of conversions for any of the sample formats. I don't think there's any good reason why CPAL only has the 4 formats it currently does - I'm guessing it was all tomaka could be bothered to support at the time or the only formats necessary on whatever platforms were initially implemented at CPAL's birth. In the future we can add more variants to CPAL's `SampleFormat` enum and support more formats and then add those branches back in here below.

src/platform/windows/asio/stream.rs
```diff
@@ -0,0 +1,908 @@
...
+                    // Get the stream
+                    // TODO is this lock necessary
+                    if let Some(ref asio_stream) = asio_streams.lock().unwrap().input {
+                        // Get the callback
+                        // TODO is this lock necessary
+                        let mut callbacks = callbacks.lock().unwrap();
+
+                        // Theres only a single callback because theres only one event loop
+                        // TODO is 64bit necessary. Might be using more memory then needed
+                        match callbacks.first_mut() {
```
mitchmindtree: The indentation here is getting pretty crazy, I'm finding it a bit tricky to keep track of what scope I'm in further down the file. Can we maybe do something like this to get rid of a couple levels of indentation:
```rust
let callback = match callbacks.first_mut() {
    None => return (),
    Some(callback) => callback,
};
```
Or maybe change `callbacks` from a `Vec` to a single callback like the TODO in the `EventLoop` type mentions?

src/platform/windows/asio/stream.rs
```diff
@@ -0,0 +1,908 @@
...
+                                                        .store(true, Ordering::SeqCst);
+                                                    sys::SILENCE_FIRST
+                                                        .store(false, Ordering::SeqCst);
+                                                    true
+                                                } else {
+                                                    false
+                                                }
+                                            },
+                                            _ => unreachable!(),
+                                        };
```
mitchmindtree: Can you clarify what's going on here with the `sys::SILENCE_FIRST` and `SILENCE_SECOND` statics here and what the index means for me? Does this refer to the input and output buffers for a single device? Can the index change depending on the order in which they're created or something?

src/platform/windows/asio/stream.rs
```diff
@@ -0,0 +1,908 @@
...
+                                            let asio_buffer: &'static mut [$AsioType] =
+                                                std::slice::from_raw_parts_mut(
+                                                    buff_ptr,
+                                                    asio_stream.buffer_size as usize,
+                                                );
+                                            for (asio_s, cpal_s) in
+                                                asio_buffer.iter_mut().zip(channel)
+                                            {
+                                                if silence {
+                                                    *asio_s = 0.0 as $AsioType;
```
mitchmindtree: This isn't correct for unsigned integers, as I'm pretty sure silence is `std::<uint_type>::MAX / 2 + 1`.

src/platform/windows/asio/stream.rs
```diff
@@ -0,0 +1,908 @@
...
+        if let Some(s) = streams
+            .get_mut(stream_id.0)
+            .expect("Bad pause stream index")
+        {
+            s.playing = false;
+        }
+        let any_playing = streams
+            .iter()
+            .filter(|s| if let Some(s) = s { s.playing } else { false })
+            .next();
```
mitchmindtree: There's an `Iterator::any` method that might be more concise than using `filter`, `next` and then checking the `Option`.

src/platform/windows/asio/stream.rs
```diff
@@ -0,0 +1,908 @@
...
+    }
+
+    /// Destroy the cpal stream based on the ID.
+    /// If no cpal streams exist then destory the
+    /// ASIO streams and clean up
+    pub fn destroy_stream(&self, stream_id: StreamId) {
+        let mut streams = self.cpal_streams.lock().unwrap();
+        streams.get_mut(stream_id.0).take();
+        let count = self.stream_count.load(Ordering::SeqCst);
+        self.stream_count.store(count - 1, Ordering::SeqCst);
```
mitchmindtree: Similarly to the `fetch_add` comment I made a while back, there's a [AtomicUsize::fetch_sub](https://doc.rust-lang.org/std/sync/atomic/struct.AtomicUsize.html#method.fetch_sub) method which could replace these two lines.

src/platform/windows/asio/stream.rs
```diff
@@ -0,0 +1,908 @@
...
+        if count == 1 {
+            *self.asio_streams.lock().unwrap() = sys::AsioStreams {
+                output: None,
+                input: None,
+            };
+            sys::clean_up();
+        }
```
mitchmindtree: What happens if a user wants to destroy their existing stream (so that there are no streams) before creating a new one (I can imagine this might be pretty common if working on a DAW-like program or something)? It might honestly not be worth worrying about this ref count and cleaning up until #245 is addressed - I can't really think of a reliable way to do it otherwise, not off the top of my head anyways.

src/platform/windows/asio/stream.rs
```diff
@@ -0,0 +1,908 @@
...
+        F: FnMut(StreamId, StreamData) + Send,
+    {
+        let callback: &mut (FnMut(StreamId, StreamData) + Send) = &mut callback;
+        self.callbacks
+            .lock()
+            .unwrap()
+            .push(unsafe { mem::transmute(callback) });
+        loop {
+            // Might need a sleep here to prevent the loop being
+            // removed in --release
```
mitchmindtree, line 943: Yes, this hard loop will either get optimised out or will loop at 100% CPU in its current state. Maybe see how the coreaudio backend handles this.

src/platform/windows/asio/stream.rs
freesig: if we are going to assert no unsigned types because asio doesn't use them can I ignore this?

src/platform/windows/asio/stream.rs
freesig: Well it would destroy the ASIO connection and then recreate it. 
But the overhead on destroy and recreate is pretty small.
I agree through, this isn't an idea solution.

freesig: Do you think I should remove this?

src/platform/windows/asio/stream.rs
freesig: I'm not sure If we can remove this. I have removed the all unsigned formats.
CPAL uses i16 but most asio drivers are i32 so it would be pretty limiting. I don't think dante does i16 so it would a block for what we are trying to do right now.

src/platform/windows/asio/stream.rs
freesig: It's the double buffer. 
So you need the first callback that ASIO calls to silence the buffer but then none of the other callbacks to silence it until it switchs back again. But because there's no way to know for sure which callback will be called first because one could be destroyed then I need these globals to track whether or not it has been done

src/platform/windows/asio/device.rs
freesig: Its required by the cpal/lib.rs:
```rust
pub struct Device(platform::Device);
```

src/platform/windows/asio/stream.rs
freesig: This happens for each callback. The `.get(count - 1)`  ensures you are just not reading writing for this callback - stream pair

freesig: `set _callback()` adds this closure to a Vec

src/platform/windows/asio/stream.rs
mitchmindtree: Yep :+1:

src/platform/windows/asio/stream.rs
mitchmindtree: Yeah I don't think it's necessary. I think it's fine to leave cleanup in the `EventLoop` drop impl like you've already got, then it should "just work" once #245 is solved.

src/platform/windows/asio/stream.rs
freesig: I'm a bit confused as to what you mean here?
Isn't the format already correct because they got it from a device

examples/enumerate.rs
freesig: I'm not sure ho to undo this?

src/platform/windows/asio/stream.rs
mitchmindtree: While you *can* get the format via `default_format` or checking supported formats, it's totally possible for a user to submit their own format struct here and request a number of channels, sample format or sample rate that isn't actually supported. No doubt a `panic!` or segfault will probably occur below in the callback if they try to do this, but I just thought it might be a good idea to check this here first if possible and return a `FormatNotSupported` error. Not a big deal if it's too much trouble.

src/platform/windows/asio/stream.rs
mitchmindtree: Ah I see, maybe this could be in a short comment above that section for future readers?

examples/enumerate.rs
mitchmindtree: No worries, not a big deal :+1:

src/platform/windows/asio/stream.rs
freesig: Ok yeh I can do this in the get streams. I'm already checking the sample rate there. 
__Channels:__
How does this work? Do I just check if the channels they want are less then the asio stream num channels?
__Sample Format:__
This is a bit harder. I guess I could make a function that takes a CPAL format and returns a ASIO format or CreationError. Then use this function in the device supported formats aswell. I just want to avoid having two different places in the code that map CPAL sampel formats to ASIO sample formats

src/platform/windows/asio/stream.rs
mitchmindtree: Ahhh I see I see, nevermind me!

src/platform/windows/asio/stream.rs
mitchmindtree: Ahh really? I would have thought it would have defaulted to f32 over i32 on both ASIO and Dante. If it is the case that `i32` is pretty common then we might as well leave it as is for now as you mention and add an `I32` `SampleFormat` variant in a future PR.

src/platform/windows/asio/stream.rs
freesig: although there is a bug with this because count is already one less then this streams id because it comes from fetch_add

src/platform/windows/asio/stream.rs
freesig: actually it's only SampleFormat::u16 that isn't supported so I'll just check that

mitchmindtree:
:warning: **Note for those testing this PR** :warning: 

Make sure you first read and follow the the step-by-step setup for using ASIO on Windows in the updated README. Hopefully we can automate more of this process in the future.

mitchmindtree:
Just leaving a note here that my current intention is to assist @ishitatsuyuki in addressing #268 and #204 first, then I'll look into rebasing this on top of that work.

freesig:
What are your thoughts on using the [sample](https://crates.io/crates/sample) crate to handle conversions between types?

PR #232, opened 2018-07-03
Upgrade lazy_static

Xaeroxe:
I'm trying to reduce duplicate dependencies in Amethyst, releasing this would help me in this endeavor.


tomaka:
I don't really see the point of reducing duplicate dependencies for such small crates, buy why not I guess.

tomaka:
Thanks

PR #235, opened 2018-07-14
Switching the headset on and off

ArtemGr:
CPAL was panicking when I was switching my headset off.

Also, it was impossible to detect default device changes by comparing the current default device with the previously obtained one.

The patch fixes these problems on Windows, I can now keep streaming an audio across the changes of the default device.

### DIFF
--- a/src/wasapi/device.rs
+++ b/src/wasapi/device.rs
@@ -542,10 +542,6 @@ impl PartialEq for Device {
         //
         // In this code section we're trying to use the GetId method for the device comparison, cf.
         // https://docs.microsoft.com/en-us/windows/desktop/api/mmdeviceapi/nf-mmdeviceapi-immdevice-getid
-        //
-        // If `GetId` mysteriously fails then we fall back to pointer comparison
-        // (we could panic, treating the `GetId` failure as something unexpected,
-        // but I simply don't know how reliable the `GetId` on Windows is).
         unsafe {
             struct IdRAII (LPWSTR);
             /// RAII for device IDs.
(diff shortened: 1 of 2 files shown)


ArtemGr:
I don't know whether this kind of `Device` comparison makes sense, haven't looked at the rest of the library code much. Though if the `Device` comparison needs to compare the instances rather than the represented devices then it should be documented in the code. The POLA, domain-driven thing to do here would be to compare the represented devices, not the instances, I think.

desttinghim:
@mitchmindtree You recently merged another pull request, could you review this one too? It's been almost a year since this was open, it should either be closed or merged, imo.

mitchmindtree:
@desttinghim for future reference, I don't always have access to OSes other than Linux to test properly locally. If one or two users can test and confirm that things are working smoothly and a patch behaves as expected, it makes things a little easier for me :)

The main change of this PR (updating the `PartialEq` implementation) seems reasonable enough to me and my limited understanding of Windows. Some of the code formatting seems very strange and to ignore the existing style, but I might try and run rustfmt to make everything consistent soon. Skipping errors around invalid devices in the main callback concerns me a little, but it's probably better than `panic!`ing as is currently the case and the true fix is probably to improve error handling for the crate across the board. I'll have a look at the existing error handling PRs next I get a chance.

I'm happy to merge for now as I see you've based #275 on this branch, so I'll assume you've tested it and it works for you at least.

desttinghim:
@mitchmindtree I am on Linux too. I'm a Core Developer for veloren, which is relying on cpal and rodio for audio. We have a couple of people on there that run Windows and have tested the changes. I can probably get them to test future changes as well.

mitchmindtree:
Great to hear! I might ping you for future windows/linux related stuff if that's OK? No worries if you're too busy though :)

Just so you're aware, I'm planning on having a bit of a sprint on CPAL within the next day or two. We also use it for commercial work via [nannou](https://github.com/nannou-org/nannou) and I want to land to #221 but it will likely mean addressing #204 first. I'm going to look at the existing PRs and issues first to see what I can land before starting work on addressing #204.

desttinghim:
Feel free to ping me! I can do linux testing easily enough, but again, I'd have to ask other people to test out windows for me. Getting the host api working would be great! I use jack on my computer, so that would help me out. I don't know how much I'll be able to help though.

ishitatsuyuki:
@mitchmindtree I'm planning to have a major API overhaul with regard to Host API, and error handling. I'm also thinking of make changes to or getting rid of EventLoop, because on many platforms audio processing is supposed to happen on a high priority thread.

What place should we discuss on this? I hang out at #audio in Amethyst Discord, but based on your preference we could discuss on IRC #rust-gamedev, this issue tracker or wherever you want. If you prefer an actual API, I can also work on my fork and experiment with what I prefer before making the actual proposal.

mitchmindtree:
Ahh that's great to hear @ishitatsuyuki!

### Host API

I actually just posted another comment on this [here](https://github.com/tomaka/cpal/issues/204#issuecomment-499123642) - would you mind if we discussed details related to the Host API there? Would love to hear your thoughts on this, your personal requirements and plans, etc!

### Error Handling

Awesome. The more I think on this lately the more I think this should probably be the number 1 priority at the moment. It reaches into absolutely everything API wise. Perhaps it's worth attempting to address this before trying to tackle the Host API? Would you mind if we discussed this further at #268?

### Removing the EventLoop

I quite like the idea of this, though I can't quite recall the feasibility of this off the top of my head. I know that this would definitely work much more nicely with the CoreAudio backend. I think it also makes more sense conceptually to just register callbacks on a per-stream basis, especially seeing as different streams might run on different devices anyway. It would certainly be more consistent with all other more established audio APIs that I've personally used. Do you want to open an issue on this and maybe discuss your plans/reasoning?

### Comms

If it's OK, I'd prefer to continue discussion on GitHub as I find it a little easier to keep track and think things through, and the decision making process stays a little more accessible for other users. Does that sound OK?

ishitatsuyuki:
Thanks for the response, that was very helpful. I've left comments on each topic, and let's discuss on GitHub as you said.

Issue #205, opened 2018-02-12
Add an option for requesting buffers of a specific size when building a stream


mitchmindtree:
Currently the buffer size varies greatly between different backends. E.g. coreaudio seems to consistently provide buffers of 512 frames, whereas the alsa and wasapi backends can vary greatly between 250 and 1000+ frames.

Some backends provide an API for requesting buffers of a specific size. This can be useful for tweaking the trade-off between cpu usage and low-latency (often useful in pro-audio applications).

This could possibly be a 3rd argument to the `build_{input/output}_stream` methods, e.g.

```rust
pub enum BufferSize {
    Default,
    Fixed(usize),
}

let input = event_loop.build_input_stream(&device, &format, BufferSize::Default).unwrap();
let output = event_loop.build_output_stream(&device, &format, BufferSize::Fixed(128)).unwrap();
```

Alternatively perhaps a `buffer_size` field can be added to the `Format` type?

From what I can remember not all backends or devices can guarantee that they can provide buffers of arbitrary, fixed sizes, so we'd either have to emphasise that this is a "best-effort" API or add an intermediary buffer so that we could manually provide buffers of a specified fixed size in the case that a backend or device cannot.

lpil:
I'm building an application that aims to have relatively low latency using ALSA. It'd be great for my use case if a small buffer size could be requested.

Thanks,
Louis

RamiHg:
Would love for this feature to exist as well. It is currently the only thing blocking me from using CPAL.

mrDIMAS:
Any news on when it will be added?

mitchmindtree:
I plan to address this as a part of [this work](https://nannou.cc/posts/moss_grant_announce), I'm just waiting to hear back from Mozilla on the final part of the agreement. I'll report back here once I begin working on this issue!

lpil:
Brilliant! Thank you

dansgithubuser:
I'm eager to get this solved, happy to contribute if possible -- I think my main unknown is how to test across different platforms/hosts. Couldn't find a contribution policy -- what's your setup?

Boscop:
Any update on this? :)

mitchmindtree:
Yes, please see #401. @JoshuaBatty and I are busy over the next couple of days but will continue work on this toward the end of this week.

mrDIMAS:
Any news?

mitchmindtree:
Closed via #401. Thanks @JoshuaBatty!

PR #239, opened 2018-08-19
Add a buffer size requesting option

0e4ef622:
Partially implements #205.

I've only implemented it for alsa, so if someone else could implement it for the other backends or at least point me in the right direction that would be great.

Also I'm not sure if and how it should return the actual buffer size. Right now it takes a mutable reference for the buffer size and then modifies it to reflect the actual buffer size.

EDIT: I think I've figured out how to implement it for WASAPI, will implement when I find time. Still clueless on coreaudio though.

EDIT2: Ok WASAPI won't let you set the buffer size, but there's [IAudioClient3](https://docs.microsoft.com/en-us/windows/desktop/api/audioclient/nn-audioclient-iaudioclient3) which might work except it's not in winapi-rs. And it's only supported in Windows 10 and up. ¯\\\_(ツ)\_/¯

### DIFF
--- a/examples/beep.rs
+++ b/examples/beep.rs
@@ -4,7 +4,7 @@ fn main() {
     let device = cpal::default_output_device().expect("Failed to get default output device");
     let format = device.default_output_format().expect("Failed to get default output format");
     let event_loop = cpal::EventLoop::new();
-    let stream_id = event_loop.build_output_stream(&device, &format).unwrap();
+    let stream_id = event_loop.build_output_stream(&device, &format, &mut cpal::BufferSize::Default).unwrap();
     event_loop.play_stream(stream_id.clone());
 
     let sample_rate = format.sample_rate.0 as f32;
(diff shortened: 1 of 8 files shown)


0e4ef622:
ack it doesn't seem to work in windows 10 it just gets ignored.

0e4ef622:
From [here](https://docs.microsoft.com/en-us/windows/desktop/api/audioclient/nf-audioclient-iaudioclient-initialize):
>For a shared-mode stream that uses event-driven buffering, the caller must set both hnsPeriodicity and hnsBufferDuration to 0. The Initialize method determines how large a buffer to allocate based on the scheduling period of the audio engine.

... Ok nevermind then.

mmckegg:
@0e4ef622 I just submitted a [PR](https://github.com/0e4ef622/cpal/pull/1) on your fork to add some coreaudio support. So far only for output streams.

Some good info here: https://developer.apple.com/library/archive/technotes/tn2321/_index.html#//apple_ref/doc/uid/DTS40013499-CH1-THE_I_O_BUFFER_SIZE

mitchmindtree:
Closing now that #401 has landed. Thanks for all your work on this!

PR #241, opened 2018-09-12
WASPI - Fix stream start after destroy

appaquet:
On Windows, creating a stream, destroying it and then trying to re-create one was not working. The issue was that the index used by the PlayStream command to find the stream & handle was the StreamId, which keeps incrementing. The second stream had stream id 1, but its index was 0. 

This fixes the issue by making sure we're not using the stream id as an offset, but always finding the right stream by its id. It may not be as performant as a direct access, but we should rarely have more than a few streams.

### DIFF
--- a/src/wasapi/stream.rs
+++ b/src/wasapi/stream.rs
@@ -449,21 +449,27 @@ impl EventLoop {
                             }
                         },
                         Command::PlayStream(stream_id) => {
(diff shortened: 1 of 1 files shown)


tomaka:
Oh thanks!

willstott101:
Would it be possible to get this released to crates.io?

Issue #231, opened 2018-06-29
Assertion failed: !buffer.is_null


SuperCuber:
Trying to run the example in the documentation for recording audio.

# Program
```rust
fn main() {
    // Setup the default input device and stream with the default input format.
    let device = cpal::default_input_device().expect("Failed to get default input device");
    println!("Default input device: {}", device.name());
    let format = device.default_input_format().expect("Failed to get default input format");
    println!("Default input format: {:?}", format);
    let event_loop = cpal::EventLoop::new();
    let stream_id = event_loop.build_input_stream(&device, &format).expect("Failed to build input stream");
    event_loop.play_stream(stream_id);
    event_loop.run(move |_, data| {
        if let cpal::StreamData::Input { buffer } = data {
            use cpal::UnknownTypeInputBuffer::*;
            match buffer {
               U16(buf) => println!("{:?}", &*buf),
               I16(buf) => println!("{:?}", &*buf),
               F32(buf) => println!("{:?}", &*buf),
            }
        }
    })
}
```

# Output
```
Default input device: Microphone (Logitech G430 Gaming Headset)
Default input format: Format { channels: 1, sample_rate: SampleRate(44100), data_type: F32 }
thread 'main' panicked at 'assertion failed: !buffer.is_null()', C:\Users\Glen\.cargo\registry\src\github.com-1ecc6299db9ec823\cpal-0.8.1\src\wasapi\stream.rs:523:29
stack backtrace:
   0: std::sys::windows::backtrace::unwind_backtrace
             at C:\projects\rust\src\libstd\sys\windows\backtrace\mod.rs:65
   1: std::sys_common::backtrace::_print
             at C:\projects\rust\src\libstd\sys_common\backtrace.rs:71
   2: std::sys_common::backtrace::print
             at C:\projects\rust\src\libstd\sys_common\backtrace.rs:59
   3: std::panicking::default_hook::{{closure}}
             at C:\projects\rust\src\libstd\panicking.rs:211
   4: std::panicking::default_hook
             at C:\projects\rust\src\libstd\panicking.rs:227
   5: std::panicking::rust_panic_with_hook
             at C:\projects\rust\src\libstd\panicking.rs:463
   6: std::panicking::begin_panic<str*>
             at C:\projects\rust\src\libstd\panicking.rs:397
   7: cpal::cpal_impl::stream::EventLoop::run_inner
[...]

PR #244, opened 2018-10-10
Assertion failed in Windows

azyobuzin:
I have encountered the panic described in #231 (but different test case). The cause is that `IAudioCaptureClient::GetBuffer` method may return `AUDCLNT_S_BUFFER_EMPTY`, which is success HRESULT, and make the buffer null.

ref: [IAudioCaptureClient::GetBuffer | Microsoft Docs](https://docs.microsoft.com/ja-jp/windows/desktop/api/audioclient/nf-audioclient-iaudiocaptureclient-getbuffer)


MOZGIII:
Why is it not merged yet?

mitchmindtree:
Thanks @azyobuzin and sorry about the delay! This looks fine to me, I'm happy to merge.

Issue #206, opened 2018-02-12
[ALSA] Enumerate example panics in cpal::cpal_impl::Device::supported_formats


mateon1:
Trying to run the enumerate example causes a panic:

```gdb
mateon@hydra:/shared/dev/rust/cpal$ RUST_BACKTRACE=1 cargo run --example enumerate
    Finished dev [unoptimized + debuginfo] target(s) in 0.0 secs
     Running `target/debug/examples/enumerate`
Default Input Device:
  Some("default")
Default Output Device:
  Some("default")
Devices: 
ALSA lib pcm_dmix.c:1052:(snd_pcm_dmix_open) unable to open slave
ALSA lib pcm_dsnoop.c:618:(snd_pcm_dsnoop_open) unable to open slave
ALSA lib pcm_dmix.c:1052:(snd_pcm_dmix_open) unable to open slave
ALSA lib pcm_dsnoop.c:618:(snd_pcm_dsnoop_open) unable to open slave
1. "iec958:CARD=PCH,DEV=0"
Error: DeviceNotAvailable
ALSA lib pcm_dmix.c:1052:(snd_pcm_dmix_open) unable to open slave
ALSA lib pcm_dmix.c:990:(snd_pcm_dmix_open) The dmix plugin supports only playback stream
ALSA lib pcm_dmix.c:990:(snd_pcm_dmix_open) The dmix plugin supports only playback stream
2. "dmix:CARD=PCH,DEV=1"
ALSA lib pcm_dmix.c:990:(snd_pcm_dmix_open) The dmix plugin supports only playback stream
thread 'main' panicked at 'device not available: "Invalid argument"', libcore/result.rs:945:5
stack backtrace:
   0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace
             at libstd/sys/unix/backtrace/tracing/gcc_s.rs:49
   1: std::sys_common::backtrace::print
             at libstd/sys_common/backtrace.rs:71
             at libstd/sys_common/backtrace.rs:59
   2: std::panicking::default_hook::{{closure}}
             at libstd/panicking.rs:380
   3: std::panicking::default_hook
             at libstd/panicking.rs:396
   4: std::panicking::rust_panic_with_hook
             at libstd/panicking.rs:576
   5: std::panicking::begin_panic
             at libstd/panicking.rs:537
   6: std::panicking::begin_panic_fmt
             at libstd/panicking.rs:521
   7: rust_begin_unwind
             at libstd/panicking.rs:497
[...]

PR #248, opened 2018-11-06
Added error handling for unknown ALSA device errors

### LINKED ISSUE #247: cpal crash on alsa backend (`beep` example)
I'm getting a crash on linux from cpal:

 - Command: `$ cargo run --example beep`

 - Stdout: `thread 'main' panicked at 'hardware params could not be set: "Invalid argument"', libcore/result.rs:1009:5`

 - stack backtrace:

```
  ...snip...
  10: cpal::cpal_impl::set_hw_params_from_format
             at src/alsa/mod.rs:805
  11: cpal::cpal_impl::EventLoop::build_output_stream
             at src/alsa/mod.rs:700
  12: cpal::EventLoop::build_output_stream
             at ./src/lib.rs:434
  13: beep::main
             at examples/beep.rs:7
   ...snip...
```

RoflCopter24:
Instead of taking the easy way out and killing the whole program by panicking, device enumeration and stream creation will now report the error variant 'Unknown' as suggested in #216 

Also added an "InvalidArgument" variant for the infamous -22 error (this should resolve #247 and make #206 more handable)

### DIFF
--- a/src/alsa/mod.rs
+++ b/src/alsa/mod.rs
@@ -92,7 +92,10 @@ impl Device {
         ) {
             -2 |
             -16 /* determined empirically */ => return Err(FormatsEnumerationError::DeviceNotAvailable),
-            e => check_errors(e).expect("device not available")
+            -22 => return Err(FormatsEnumerationError::InvalidArgument),
+            e => if check_errors(e).is_err() {
+                return Err(FormatsEnumerationError::Unknown)
+            }
         }
 
         let hw_params = HwParams::alloc();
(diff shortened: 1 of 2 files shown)


derekdreery:
LGTM I was playing with something like this myself in https://github.com/tomaka/cpal/pull/223#issuecomment-435906472

derekdreery:
I'm just going to take this opportunity to moan about the awful error reporting in alsa, and C libs in general. They give you nothing to work with but a meaningless posix error code.

RoflCopter24:
@derekdreery I really like your idea to get the error message directly from alsa. That should enable more differentiated error handling so that if you get some weird error, at least you get somewhat of an idea what went wrong.

That beeing written, let me add some rant as well: 
I very much dislike the prominence with which Rust advertises panics - I think they should not be used in library crates at all. If a library fails its task, there is absolutely no reason to crash a whole thread or maybe even the whole program. That's what Results/Options are for!
Often I get the impression, that ppl are too lazy to implement proper error handling and think 'I throw this panic in here, so I can get on with my life and be done'.

derekdreery:
@RoflCopter24 I agree with you that panicking is often the wrong thing to do for stable libraries, but all libraries started out as prototypes, and when you're prototyping you don't want to spend ages on errors. It just needs someone to improve error handling here, which anyone can do I think.

desttinghim:
What is holding this Pull Request back?

derekdreery:
Personally, I think this should be merged, as it is an improvement on the current state of things.

desttinghim:
@tomaka Can you comment?

PR #253, opened 2019-01-15
impl Debug for Device

chris-morgan:
The internal alsa, null and emscripten Device implementations already implemented Debug; but the coreaudio and wasapi ones, and therefore also the wrapper, did not.

I decided to eschew the `Device(…)` wrapping in the outer layer (hence a custom implementation rather than `#[derive(Debug)]`), because `Device(Device)`, `Device(Device { … })` and so forth all look better without the extra `Device(…)` wrapping.

On the wasapi and coreaudio implementations I put both the pointer and name. Name because it’s useful, pointer because on Windows at least I believe duplicated names are possible. (e.g. two monitors that include monitors, of the same type; I haven’t strictly confirmed this, because I killed those off harshly on my machine and don’t want to reinstate them.)

I do not have access to a macOS device to confirm that the coreaudio implementation is sane, but I think it is.

### DIFF
--- a/src/coreaudio/mod.rs
+++ b/src/coreaudio/mod.rs
@@ -15,6 +15,7 @@ use UnknownTypeInputBuffer;
 use UnknownTypeOutputBuffer;
 
 use std::ffi::CStr;
+use std::fmt;
 use std::mem;
 use std::os::raw::c_char;
 use std::ptr::null;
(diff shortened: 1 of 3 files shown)


michaelenger:
Ran this on macOS 10.14 and it looks good 👍 
```rust
for device in cpal::devices() {
    dbg!(device);
}
```

PR #258, opened 2019-03-20
Added test that demonstrates cpal crashing with ALSA when doing long input-reading sessions

MOZGIII:

PR #261, opened 2019-03-31
Added crash example

MOZGIII:
See #258 and #215


MOZGIII:
Obsolete after alsa fix

PR #267, opened 2019-04-29
Fix #254, logic error in frame/sample size

### LINKED ISSUE #254: Segfault if delaying call to event_loop.run()
Running this simple code :

```
use std::io;

// Setup ports, and start threads
fn main() {
    // Setup the default input device and stream with the default input format.
    let device = cpal::default_input_device()
        .expect("Failed to get default input device");
    println!("Default input device: {}", device.name());

    // Get the default sound input format
    let format = device
        .default_input_format()
        .expect("Failed to get default input format");
    println!("Default input format: {:?}", format);

    // Start the cpal input stream
    let event_loop = cpal::EventLoop::new();
    let stream_id = event_loop
        .build_input_stream(&device, &format)
        .expect("Failed to build input stream");
    event_loop.play_stream(stream_id);

    // Await user input
    println!("Press enter/return to start reading frequencies ...");
    let mut user_input = String::new();
    io::stdin().read_line(&mut user_input).ok();

    // Spawn the audio input reading thread
    std::thread::spawn(move || {
        event_loop.run(move |_, _| {});
    });

    // Wait for user input to quit
    println!("Press enter/return to quit...");
    let mut user_input = String::new();
    io::stdin().read_line(&mut user_input).ok();
}
```

with the output:

```
  Compiling improve v0.1.0 (/home/lgarczyn/Projects/improve)
    Finished dev [unoptimized + debuginfo] target(s) in 0.47s
     Running `target/debug/improve`
Default input device: "default"
Default input format: Format { channels: 2, sample_rate: SampleRate(44100), data_type: F32 }
Press enter/return to start reading frequencies ...

Press enter/return to quit...
double free or corruption (!prev)
[1]    17669 segmentation fault (core dumped)  cargo run`
```
The code doesn't crash on MacOS, and if I remove the "Await user input" lines, the code works fine. 
I'm using Linux Mint, using both Alsa and Pulseaudio

ishitatsuyuki:
The issue originated from an logic error. Let's avoid casting like a blind by the way.

Vec are now zero-filled from the start. There's no point in leaving heap uninitialized ([ref](https://github.com/rust-lang/rust/issues/54628)).

And hey, please do not wrap the entire function in unsafe. It's a disaster to audit.

gg.

### DIFF
--- a/src/alsa/mod.rs
+++ b/src/alsa/mod.rs
@@ -517,7 +517,7 @@ impl EventLoop {
                     }
 
                     // Determine the number of samples that are available to read/write.
-                    let available = {
+                    let available_samples = {
                         let available = alsa::snd_pcm_avail(stream_inner.channel); // TODO: what about snd_pcm_avail_update?
 
                         if available == -32 {
@@ -533,7 +533,7 @@ impl EventLoop {
                         }
                     };
 
-                    if available < stream_inner.period_len {
+                    if available_samples < stream_inner.period_len {
                         i_descriptor += stream_inner.num_descriptors as isize;
                         i_stream += 1;
                         continue;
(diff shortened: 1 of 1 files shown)


mitchmindtree:
Thanks! I just tested this locally and all seems fine :+1:

appveyor seems to be failing due to network errors, but seeing as this only touches ALSA code I think it's fine to ignore it and merge as travis is passing.

PR #269, opened 2019-04-30
Better buffer handling

ishitatsuyuki:
- ALSA backend: reuse the buffers
- Make `InputBuffer` and `OutputBuffer` types just a wrapper of slice
  * Buffer is now submitted at the end of callback

### DIFF
--- a/src/alsa/mod.rs
+++ b/src/alsa/mod.rs
@@ -15,7 +15,7 @@ use SupportedFormat;
 use UnknownTypeInputBuffer;
 use UnknownTypeOutputBuffer;
 
-use std::{cmp, ffi, iter, mem, ptr};
+use std::{cmp, ffi, mem, ptr};
 use std::sync::Mutex;
 use std::sync::atomic::{AtomicUsize, Ordering};
 use std::vec::IntoIter as VecIntoIter;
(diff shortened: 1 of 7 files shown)


ishitatsuyuki:
Travis did not react to draft->ready change, triggering it.

ishitatsuyuki:
I decided to fix all the borrow checking and misc issues. **This PR is now ready to merge.**

mitchmindtree:
Thanks for this @ishitatsuyuki, it looks mostly good to me! Were you able to run the examples on each of the platforms? If not, which ones are you yet to test? I'm about to head to sleep now but I'll try to remember to test this on my linux machine tomorrow. Unfortunately there's no way for me to mark github issue notifications as unread and many things slip by me as a result, so feel free to ping me if I haven't gotten around to this in a couple days.

ishitatsuyuki:
I have tested on Linux with ALSA, with or without PulseAudio. Other platforms are not tested.

(As you can see, the code change for them should be minimal. So the risk that they break is small.)

mitchmindtree:
OK I've just tested the beep example locally on both windows and linux and it seems to be running smoothly still. I'm happy to merge!

Issue #268, opened 2019-04-30
Better error handling interface


ishitatsuyuki:
The codebase is currently filled extensively with `expect()` and `unwrap()`. The thing is that device disconnect does happen, and we don't want to crash the entire application/game when your Bluetooth connection get lost.

A brief idea is to add some error event/callback to event loop. Probably need some more research on how other libraries do this.

mitchmindtree:
Right now I see two main areas of improvement for error handling that need to be addressed.

### 1. Synchronous API functions

Currently it is possible to return errors properly in functions like `default_output_device()` and `device.default_output_format()`, however in some cases there aren't enough descriptive variants on the associated error `enum` to provide enough information. As a result, this leads to

- `panic!`ing instead as there is no matching error variant
- lumping in errors that are only vaguely related to the actual variant

This leads me to believe that we need a more robust error type for handling these cases. Here are some ideas for improvements that come to mind:

- **Adding more variants to the error enum.** Enum variants are especially useful on errors when actually handling them. E.g. if an attempted format was incompatible, we can match on the associated variant and instead try another format. I think it would be worth reviewing the kinds of errors that can occur in each of the different functions across each of the different backends to see what kind of common variants we can add. I think it might be worth adding variants even if they are only specific to one or two backends. We could add doc comments describing which platforms these errors can occur on. CPAL users can then also be aware of the full set of potential errors that might occur across each of the different backends. This would surely be more useful than only offering the ability to handle the subset of errors that can occur on every different backend.
- **Allowing to provide backend-specific error causes.** While the actual errors that occur could be described in the error enum associated with each function, the cause for the error might be different depending on the platform. Perhaps it's worth adding a `cause: String` field or something along these lines that allow for adding the word-for-word cause provided by the platform's specific error message?

### 2. Errors that occur while running the callback

These are errors that occur during the "looping" process where the `EventLoop` waits on file descriptors or some other fence until the host is ready to receive or deliver more audio and the user callback is called. Often-times this is on a separate thread to which the `EventLoop` is `run` (macOS comes to mind).

@ishitatsuyuki yes it seems like some sort of callback approach might be the only way to handle errors directly where they occur. I think I like your idea of providing a separate callback specifically for errors. I guess the other approach would be to change the user audio callback argument to something like `Result<Buffer, SomeError>`? Either way, both of these options would allow the user to do any cleanup necessary of state in the audio callback and potentially send the error over a channel to the GUI/CLI thread to provide feedback to users :+1:

[...]

PR #275, opened 2019-06-03
Don't panic when setting hardware params in alsa

desttinghim:
This PR is based off of #235 in the hopes that it will be merged sometime soon. This PR returns errors from set_hw_params_from_format instead of panicking: This prevents a crash on linux when a user is running jack and selects it as the audio source.


mitchmindtree:
Thanks, this looks fine to me for now, but hopefully we can improve from returning results with `String`s to proper errors as a part of upcoming work on #268.

desttinghim:
Agreed, I definitely wouldn't keep the strings around forever.

PR #276, opened 2019-06-06
Release 0.9.0

AustinJ235:


AustinJ235:
hmm didn't think till after I added this, but I guess updating hound and lazy static would be breaking changes.

mitchmindtree:
Thanks for this @AustinJ235!

No I think you're fine, `hound` is just a dev-dependency (only used in an example) and I believe `lazy_static` is only used internally and isn't exposed publicly.

That said, I have just gone through the recent PRs and realised that #248 is in fact a breaking change as it adds variants to some public error types. Would you mind changing this PR to update to version 0.9.0? Also, could you add some entries to the CHANGELOG for #248, #267 and #269? I just realised they're not mentioned yet.

PR #280, opened 2019-06-07
Remove unnecessary panic! from OutputBuffer Deref impl

### LINKED ISSUE #228: Panic when calling len() on OutputBuffer
It looks like `len` triggers the deref implementation, which panics for an `OutputBuffer`.

The error was hard to figure out, since it said "It is forbidden to read from the audio buffer", but I didn't think that I was reading it by calling `len()`. Could this be allowed or caught at compile time?

mitchmindtree:
Since #269 this `panic!` is certainly unnecessary as `InputBuffer` and
`OutputBuffer` are a thin wrapper around a slice. That said, I'm
struggling to understand exactly why this `panic!` was necessary in the
first place.

This closes #228.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -602,7 +602,7 @@ impl<'a, T> Deref for OutputBuffer<'a, T>
 
     #[inline]
     fn deref(&self) -> &[T] {
-        panic!("It is forbidden to read from the audio buffer");
+        self.buffer
     }
 }
 


ishitatsuyuki:
Just for clarity: the buffers used to have uninitialized memory that could be considered UB. After the refactoring they contain *unspecified* but valid content (e.g. no sNaN).

mitchmindtree:
Ahh of course I remember now, thanks for the hot tip!

PR #283, opened 2019-06-13
Use channels instead of mutex

ishitatsuyuki:
ALSA and WASAPI backends now doesn't use a Mutex that is locked every time data is processed. These two backends were straightforward to modify, other backends were not modified.

Changes are self descriptive.

### DIFF
--- a/src/wasapi/stream.rs
+++ b/src/wasapi/stream.rs
@@ -16,6 +16,7 @@ use std::mem;
 use std::ptr;
 use std::slice;
 use std::sync::Mutex;
+use std::sync::mpsc::{channel, Sender, Receiver};
 use std::sync::atomic::AtomicUsize;
 use std::sync::atomic::Ordering;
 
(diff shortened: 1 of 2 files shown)


src/wasapi/stream.rs
```diff
@@ -405,14 +404,7 @@ impl EventLoop {
 
     #[inline]
     pub fn destroy_stream(&self, stream_id: StreamId) {
-        unsafe {
-            self.commands
-                .lock()
-                .unwrap()
-                .push(Command::DestroyStream(stream_id));
-            let result = synchapi::SetEvent(self.pending_scheduled_event);
```
mitchmindtree, line 413: Just double checking, do we not need this anymore? I'm not familiar enough to remember the purpose of this line - is it to wake up the processing if stopped perhaps?

src/wasapi/stream.rs
```diff
@@ -618,17 +611,19 @@ impl EventLoop {
 
     #[inline]
     pub fn play_stream(&self, stream: StreamId) {
-        unsafe {
-            self.commands.lock().unwrap().push(Command::PlayStream(stream));
-            let result = synchapi::SetEvent(self.pending_scheduled_event);
```
mitchmindtree, line 623: Same with this

src/wasapi/stream.rs
ishitatsuyuki: I did a silent refactor to remove the duplicates - the logic is moved to a new `push_command` function.

mitchmindtree: Ahhh yes I see now, my bad, I missed it on initial review!

PR #285, opened 2019-06-19
Update feedback.rs

mohanson:


mohanson:
The origin code paniced on my computer

```rs
thread 'main' panicked at 'Failed to get default format: StreamTypeNotSupported', src\libcore\result.rs:997:5
stack backtrace:
```

When I changed it, it goes well. But I'm not very sure is it right.

mitchmindtree:
I'm guessing the source of this issue is that many commercial devices like laptops only have monophonic microphone input, but the vast majority also have stereo output. Thus, when we request to try and open an input stream using the default output format it would panic as it's not possible to open a stereo stream with a monophonic input. I agree that this fix will likely improve the situation. A better fix might be to use the `min` channels between both the input and output formats. If anyone would like to do a PR for that in the future, feel free!

Issue #278, opened 2019-06-06
Getting rid of the EventLoop


ishitatsuyuki:
More and more platforms are running audio with synchronous callback in a high-priority thread. That means that thread management should happen inside cpal, not in user code.

The blocking design of `EventLoop` should be replaced by some interface of registering callback. What's unfortunate is that we would need not only `Send` but also the `'static` bound on the callback if it's implemented naively. Do we have anything more Rust-ic than callbacks (which is C-ish and often messy)?

It's obvious that we need different callbacks for different `Host`s (assuming it get implemented), but should we support driving multiple streams with the same callback?

mitchmindtree:
Ahh this reminds me that, if we are to get rid of the `EventLoop`, we're going to have to spawn a thread internally for platforms that don't automatically use a separate, high-priority thread.

E.g. the current ALSA implementation is just a big synchronous loop that blocks on file descriptors and at a quick glance the WASAPI implementation of `run` also looks quite similar. I believe this was the original rational for the `EventLoop` - to provide a consistent blocking API so that the behaviour is the same across all platforms regardless of whether or not the platform calls the callback given to `run` on a separate thread or not.

I'm not sure I feel comfortable with CPAL spawning an entire thread within the wasapi and alsa implementations just to allow for non-blocking API consistency. Right now, the user can gain a *kind* of consistent, non-blocking API by calling `fn run` on a separate thread. While this is annoying for coreaudio users, the alternative approach (of spawning a thread within the wasapi and alsa backends in order to remove the blocking EventLoop) is equally inconvenient to wasapi and alsa users. There are also many different ways to spawn a thread, and I imagine it might be difficult for CPAL to do this in a way that is optimal for all users.

Also, before doing something like this it would likely be worth considering how web support will look. I haven't looked at that API yet, but I imagine it will be important to consider as I'm aware just "spawning a thread" in web APIs is not a simple task.  Hopefully someone with some more knowledge on this can comment.

I do agree that the current situation with the blocking `EventLoop` is not great, but after giving it some thought I'm not convinced that removing the `EventLoop` won't just move the current issues to somewhere else...

## EventLoop Issues

Perhaps it is first worth outlining the specific issues that currently exist with the `EventLoop`:

- In attempting to provide a unified blocking API, it requires users of platforms that only offer non-blocking (high-priority) APIs to spawn one extra thread than they should otherwise have to.
[...]

PR #286, opened 2019-06-20
Error Handling Overhaul

mitchmindtree:
The aim of this PR is to overhaul the way that errors are handled throughout the crate as discussed within #268.

I will continue to update the following list and commit to this PR as I make progress. Feel free to comment at any stage!

## TODO:

- [x] Integrate the `failure` crate to make error implementation and propagation more ergonomic.
- [x] Rename existing error types to more closely match the methods from which they may be returned. Carry this on going forward as a general convention.
- [x] Review every function of the synchronous API and their implementations within each of the different backends. Remove `panic!`s and improve error handling where necessary, which may include adding new error types or new variants to existing error types.

## Follow-up PR:

- [ ] Change the user callback to emit an `Event` enum as described [here](https://github.com/tomaka/cpal/issues/268#issuecomment-499887735).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -9,6 +9,7 @@ license = "Apache-2.0"
 keywords = ["audio", "sound"]
 
 [dependencies]
+failure = "0.1.5"
 lazy_static = "1.3"
 
 [dev-dependencies]
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -7,9 +7,6 @@ environment:
       target: x86_64-pc-windows-msvc
     - channel: nightly
       target: x86_64-pc-windows-msvc
-    # GNU
-    - channel: stable
-      target: x86_64-pc-windows-gnu
 
 matrix:
   allow_failures:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -5,21 +5,11 @@ environment:
     # MSVC
     - channel: stable
       target: x86_64-pc-windows-msvc
-    - channel: stable
-      target: i686-pc-windows-msvc
     - channel: nightly
       target: x86_64-pc-windows-msvc
-    - channel: nightly
-      target: i686-pc-windows-msvc
     # GNU
     - channel: stable
       target: x86_64-pc-windows-gnu
-    - channel: stable
-      target: i686-pc-windows-gnu
-    - channel: nightly
-      target: x86_64-pc-windows-gnu
-    - channel: nightly
-      target: i686-pc-windows-gnu
 
 matrix:
   allow_failures:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,16 +1,38 @@
+os: Visual Studio 2015
+
+environment:
+  matrix:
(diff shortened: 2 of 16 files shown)


mitchmindtree:
OK this should be just about ready to go, but I still want to test the examples on macOS and check that windows is compiling and running correctly (appveyor seems to be erroring out early on something unrelated).

cc @ishitatsuyuki would you mind reviewing this if you get a chance in the meantime? It would be great to get your thoughts.

ishitatsuyuki:
Here are my thoughts after roughly looking through:
- The AppVeyor failure is likely due to failure pulling in a new dependency that requires C compiler. We should grab MinGW GCC somehow, or test with MSVC instead.
- Do you think it's better to make use of `Termination` (`?` in `main()`)? Personally I prefer it bacause it's shorter to write that unwrap, and also displays the error more cleanly.

Regarding error types on play/pause methods: should we make it synchronous as described in the last line of https://github.com/tomaka/cpal/issues/278#issuecomment-501704876? Running the operations synchronously will allow us to return error (if it exists) on all platforms, making it more consistent.

mitchmindtree:
> The AppVeyor failure is likely due to failure pulling in a new dependency that requires C compiler. We should grab MinGW GCC somehow, or test with MSVC instead.

Ahh I see, I'm quite unfamiliar with appveyor, do you happen to know off the top of your head what needs to be added to our appveyor.yml to get this working again?

> Do you think it's better to make use of Termination (? in main())? Personally I prefer it bacause it's shorter to write that unwrap, and also displays the error more cleanly.

This sounds good to me! Do you still get the line number of the source of the error? I guess this error overhaul should make it clear where the error is coming from anyway. I'll give this a go!

> Regarding error types on play/pause methods: should we make it synchronous as described in the last line of #278 (comment)? Running the operations synchronously will allow us to return error (if it exists) on all platforms, making it more consistent.

Yeah I agree we should. I kind of had this in mind while making the new `PlayStreamError` and `PauseStreamError` types, I figured that while they're basically just wrappers around `BackendSpecificError` for now, we could expand upon them as we work through #278. Do you mind if this work waits until #278?

In the meantime I'm going to spin up a VM to test windows locally, and see if I can borrow a friend's MBP to test the coreaudio backend.

ishitatsuyuki:
> do you happen to know off the top of your head what needs to be added to our appveyor.yml to get this working again?

Sorry no, I only looked at the error message and I have no experience with AppVeyor.

> Do you mind if this work waits until #278?

Yes, that was just a by-the-way proposal. Let's revisit after we redesign things around EventLoop.

mitchmindtree:
OK I've tested locally on both wasapi and coreaudio and can confirm everything compiles and that the examples all run as well as they did beforehand. I'm going to do some investigating into this appveyor issue now.

ishitatsuyuki:
Tip: you can check code by installing the `x86_64-pc-windows-gnu` target on Linux and then use either Cargo or RLS.

PR #288, opened 2019-06-21
User Callback API `StreamEvent` Overhaul

### LINKED ISSUE #268: Better error handling interface

mitchmindtree:
Add new `StreamEvent` type - enables more flexible user callback API.

This adds the following types:

- `StreamEvent`
- `CloseStreamCause`
- `StreamError`

These allow for notifying the user of the following events:

- A stream has been played.
- A stream has been paused.
- A stream has been closed due to the user destroying stream.
- A stream has been closed due to an error.

## TODO

- [x] Add new `StreamEvent` API to top level.
- [x] Update examples for new `StreamEvent` API.
- [x] Update ALSA backend.
- [x] Update WASAPI backend.
- [x] Update CoreAudio backend.
- [x] Update emscripten backend.
- [x] Update null backend.

Closes #268.

### DIFF
--- a/examples/beep.rs
+++ b/examples/beep.rs
@@ -17,7 +17,16 @@ fn main() -> Result<(), failure::Error> {
         (sample_clock * 440.0 * 2.0 * 3.141592 / sample_rate).sin()
     };
 
-    event_loop.run(move |_, data| {
+    event_loop.run(move |id, event| {
+        let data = match event {
+            cpal::StreamEvent::Data(data) => data,
+            cpal::StreamEvent::Close(cpal::StreamCloseCause::Error(err)) => {
+                eprintln!("stream {:?} closed due to an error: {}", id, err);
+                return;
+            }
+            _ => return,
+        };
+
         match data {
             cpal::StreamData::Output { buffer: cpal::UnknownTypeOutputBuffer::U16(mut buffer) } => {
                 for sample in buffer.chunks_mut(format.channels as usize) {
--- a/examples/beep.rs
+++ b/examples/beep.rs
@@ -17,14 +17,13 @@ fn main() -> Result<(), failure::Error> {
         (sample_clock * 440.0 * 2.0 * 3.141592 / sample_rate).sin()
     };
 
-    event_loop.run(move |id, event| {
-        let data = match event {
-            cpal::StreamEvent::Data(data) => data,
-            cpal::StreamEvent::Close(cpal::StreamCloseCause::Error(err)) => {
+    event_loop.run(move |id, result| {
+        let data = match result {
+            Ok(data) => data,
+            Err(err) => {
                 eprintln!("stream {:?} closed due to an error: {}", id, err);
                 return;
             }
-            _ => return,
         };
 
         match data {
--- a/examples/beep.rs
+++ b/examples/beep.rs
@@ -21,7 +21,7 @@ fn main() -> Result<(), failure::Error> {
         let data = match result {
             Ok(data) => data,
             Err(err) => {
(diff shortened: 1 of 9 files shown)


mitchmindtree:
cc @ishitatsuyuki it would be great to get your thoughts on this too as it aims to address the other half of #268. My plan is to do the coreaudio backend tomorrow and then it should be ready to go.

mitchmindtree:
OK I think this should be ready to go now. I'll leave this open for a bit longer for feedback. In the meantime, I'm going to start a branch addressing #204 based on this branch.

ishitatsuyuki:
I'm sorry I'm a bit late! Thanks for implementing this, but I'm not sure I agree with the need of play/pause events. First, the ability to inject events into the audio processing thread varies a lot between platforms, and this seems to introduce more problems. For example, the macOS implementation relies on the existence of Mutex to deliver such events, and this will not change even after the EventLoop refactor. And second, I'm not sure if exposing such events would be useful, as it's not something the low level API implements. Did you have some particular library in mind when making this?

Given that waking up the audio thread with arbitrary events is not portable, I think we are left with two options:

- Pass any events if exist when the data callback is called. Which means, it takes `(Id, Buffer, Option<ErrorsAndEvents>)`. (Or using `Vec` and other combinations of flags. This could also include xrun informations.)
- Use another thread and callback solely for errors (basically Android's API).

Also a small note: device are often considered *invalidated* when it gets disconnected. It's a state where you cannot perform operations, but still need to destroy it as usual, therefore is not the same as *close*.

mitchmindtree:
> but I'm not sure I agree with the need of play/pause events. First, the ability to inject events into the audio processing thread varies a lot between platforms, and this seems to introduce more problems

Yes I think you're right, originally I thought it might be nice for users to have access to play/pause events in case they wanted to reset some state (like oscillator phases e.g.) but this is not really essential and not worth the burden added to CPAL. I'll remove the Play and Pause events.

> Given that waking up the audio thread with arbitrary events is not portable, I think we are left with two options:
> 
>    * Pass any events if exist when the data callback is called. Which means, it takes (Id, Buffer, Option<ErrorsAndEvents>). (Or using Vec and other combinations of flags. This could also include xrun informations.)
>    * Use another thread and callback solely for errors (basically Android's API).

I think I tend to agree with [your comment](https://github.com/tomaka/cpal/issues/268#issuecomment-499744064) from #268:

> Personally I think error callback forms a pretty hard-to-use API.
> 
> I think the only error we should expose on all platforms is device disconnect, and for that purpose we should not pass the burden of synchronization to users.

I'm leaning towards the first option, however I think the signature should probably be something like `(Id, StreamDataResult, StreamFlags)`. `StreamDataResult` would be an alias for `Result<StreamData, StreamError>` as in many cases device invalidation indicates that the data is invalid, and even if the data is valid the data and error can be delivered via two separate calls.

I think it's still important that `StreamError` can indicate both `DeviceInvalidated` and `BackendSpecific` errors, as backends like wasapi and alsa can both error for other reasons like file descriptor errors, etc. 

`StreamFlags` could be a bitflags type where events such as buffer underruns and other events can be indicated, but this could be addressed in a follow up PR.

What are your thoughts?

mitchmindtree:
OK, I've made the changes mentioned in my last comment - I'm going to land this as is for now, but if anyone has any feedback or input feel free to share here or open a follow-up PR!

Issue #259, opened 2019-03-21
Pulse Audio support


knappador:
Tagging #250 regarding the sibling Linux sound server.  I'm about to release a program under LGPL3 to which I own the full copyright (for now),  but I don't want to discourage anyone from harvesting any bits and pieces in the name of getting all sound server API's under one roof, so I [released an MIT licensed gist](https://gist.github.com/knappador/bda55d5eca849c452c7bcc32c521e5f4).

The main value of this gist for consumption is having a bit more abstract example than the documentation libpulse.  Likely I'm going to overhaul it in the owning repo within the month.

My goal was not to get down to the rust libpulse bindings sys crate or talk directly to the server, but must mention that I quickly developed strong antipathy for the mainloop threaded API, which is [inherited from the C client.](https://github.com/jnqnfe/pulse-binding-rust/blob/012e15c673916ce37218729c49f6611ec580b970/pulse-sys/src/mainloop/threaded.rs)  Possibly the client-server interaction scheme has a deep pathological will to make pseudo-synchronous operations rely on libpulse's behavior and polling loop.  Attempting to circumvent libpulse may involve worse hazards than managing them.

Playing directly to ALSA is a bad idea on Linux as it takes exclusive access to devices, and it's a bit weird to create monitors to listen to what other programs play.

knappador:
Mainloop not being thread safe and including an RC is quite a pain.  Implementation in the case that there are several possible reader threads (I have this use case) will require some form of thread parking or _possibly_ scoped threads to hold a mainloop for any length of time.  Multiple contexts seems bad style.  Basically to make non-blocking calls from an application, you have to poll or implement callbacks into your own event loop mechanism and still pseudo-serialize the calls to pulse or implement a state tracking API around the API :shell: :shell: :shell:

MOZGIII:
Technically, when `cpal` is using ALSA on the properly configured system with PulseAudio installed and running, it's most likely passing the data through to the pulse via the plugin.
See https://wiki.debian.org/PulseAudio.

That said, I definetly see the value of talking to Pulse directly without going through asla-lib, becuase it adds just another layer of bugs and latency.

knappador:
> it's most likely passing the data through to the pulse via the plugin

Is "it" Cpal or ALSA?

I read the wiki but did not come away understanding an obviously related set of implications.  It seems implications could exist both for extending CPAL or configuring PA & ALSA so that CPAL uses PA without libpulse.

[...]

Issue #250, opened 2018-11-24
Support jack


derekdreery:
Would you consider a PR for jack support (it would be behind a feature for linux)?

knappador:
@derekdreery  did you get a PR going?  I obviously don't have merge access on this repo but am willing to help work on a fork that targets sound server API's.

derekdreery:
I haven't done any work on this yet.

mitchmindtree:
One important consideration yet to be settled upon is the way in which we allow for supporting multiple possible audio backends per platform (e.g. alsa, pulse or jack on linux, wdm or asio on windows, etc). Right now, CPAL assumes just one per backend per platform (WDM on windows, ALSA on linux, CoreAudio on macos).

I posted some ideas regarding this at #204 which I still think might be a good direction, I'm hoping to get some time to flesh this out a little more in the near future. There's also an ASIO PR #221 which makes some adjustments to the module system to make this multiple-backend-per-platform support possible.

Anyway, just wanted to mention this as something to keep in mind!

knappador:
https://github.com/tomaka/cpal/issues/204#issuecomment-365276479

>  The "host" is supposed to be cpal because cpal is supposed to be as low-level as possible.

The only reason to go below a sound server is IMO to be a sound server or an embedded device library.  It's probably worth checking that perception before I get the idea to work on importing my work on PulseAudio.

Is the best solution ultimately going to make an abstraction layer over devices and then an abstraction layer over sound servers where the two are independent?  I expect the answer cannot be a  clear yes or no because some platforms might not have one or the other.  If it can't be either, it almost has to be both.

x37v:
> Would you consider a PR for jack support (it would be behind a feature for linux)?

@derekdreery Jack works on osx and windows as well.

ErikNatanael:
I thought I'd have a look at integrating the [Rust jack bindings](https://github.com/RustAudio/rust-jack) over the weekend. Give me a shout if you have tips/tricks/WIPs/want to collaborate on this!

derekdreery:
@ErikNatanael awesome - I've got quite a lot of work on this weekend, but any questions you have post them here. :)

ErikNatanael:
[...]

PR #289, opened 2019-06-24
Introduce a `Host` API

mitchmindtree:
This is an implementation of the API described at #204. Please see that
issue for more details on the motivation.

-----

A **Host** provides access to the available audio devices on the system.
Some platforms have more than one host available, e.g.
wasapi/asio/dsound on windows, alsa/pulse/jack on linux and so on. As a
result, some audio devices are only available on certain hosts, while
others are only available on other hosts. Every platform supported by
CPAL has at least one **DefaultHost** that is guaranteed to be available
(alsa, wasapi and coreaudio). Currently, the default hosts are the only
hosts supported by CPAL, however this will change as of landing #221 (cc
@freesig). These changes should also accommodate support for other hosts
such as jack #250 (cc @derekdreery) and pulseaudio (cc @knappador) #259.

This introduces a suite of traits allowing for both compile time and
runtime dispatch of different hosts and their uniquely associated device
and event loop types.

A new private **host** module has been added containing the individual
host implementations, each in their own submodule gated to the platforms
on which they are available.

A new **platform** module has been added containing platform-specific
items, including a dynamically dispatched host type that allows for
easily switching between hosts at runtime.

The **ALL_HOSTS** slice contains a **HostId** for each host supported on
the current platform. The **available_hosts** function produces a
**HostId** for each host that is currently *available* on the platform.
The **host_from_id** function allows for initialising a host from its
associated ID, failing with a **HostUnavailable** error. The
**default_host** function returns the default host and should never
fail.

Please see the examples for a demonstration of the change in usage. For
the most part, things look the same at the surface level, however the
role of device enumeration and creating the event loop have been moved
from global functions to host methods. The enumerate.rs example has been
updated to enumerate all devices for each host, not just the default.

**TODO**

- [x] Add the new **Host** API
- [x] Update examples for the new API.
- [x] Null host
- [x] ALSA host
- [x] WASAPI host
- [x] CoreAudio host
- [x] Emscripten host

**Follow-up PR**
- [ ] ASIO host #221

cc @ishitatsuyuki more to review for you if you're interested.


src/host/alsa/mod.rs
```diff
@@ -7,16 +7,21 @@ use ChannelCount;
 use BackendSpecificError;
 use BuildStreamError;
 use DefaultFormatError;
+use Device as DeviceTrait;
 use DeviceNameError;
+use DevicesError;
+use EventLoop as EventLoopTrait;
 use Format;
+use Host as HostTrait;
```
ishitatsuyuki, line 15: Personally I’d prefer names like `AlsaHost` and avoid renaming imports, but I’ll leave it up to you

src/host/emscripten/mod.rs
```diff
@@ -10,17 +10,125 @@ use stdweb::web::set_timeout;
...
+    }
+}
+
+impl HostTrait for Host {
+    type Devices = Devices;
+    type Device = Device;
+    type EventLoop = EventLoop;
+
+    fn is_available() -> bool {
+        // Assume ALSA is always available on linux/freebsd.
```
ishitatsuyuki: Copy paste leftover

src/lib.rs
```diff
@@ -132,51 +148,203 @@ extern crate lazy_static;
...
+///
+/// For example, WASAPI is the standard audio host API that ships with the Windows operating
+/// system. However, due to historical limitations with respect to performance and flexibility,
+/// Steinberg created the ASIO API providing better audio device support for pro audio and
+/// low-latency applications. As a result, it is common for some devices and device capabilities to
+/// only be available via ASIO, while others are only available via WASAPI.
+///
+/// Another great example is the Linux platform. While the ALSA host API is the lowest-level API
+/// available to almost all distributions of Linux, its flexibility is limited as it requires that
+/// each process have exclusive access to the devices with which they establish streams. PortAudio
```
ishitatsuyuki: PulseAudio

derekdreery:
I'm excited by this. After it lands, I'd like to take a look at the alsa implementation to see if I can improve it. I think we cut some corners with regards to selection of parameters for the connection, with the consequence that we have panics and some valid configurations of audio hardware fail to work.

ishitatsuyuki:
One thing: I think DefaultHost should be replaced in favor of the polymorphic version, therefore the defaults should be determined at runtime.

This is because on Linux the preferred audio API varies a lot, depending on the audio server installed (PulseAudio or Jack) or even drivers (some users may be using OSS).

src/host/alsa/mod.rs
mitchmindtree: Yes I normally would agree, but in this case having each of the private backend types was necessary to keep the dynamic dispatch code generation much easier. I have re-exported the platform-specific types in the `platform_impl` module with proper names as you describe (e.g. `Host` as `AlsaHost`) so that users can still use these directly with friendlier names if they require :)

mitchmindtree:
> One thing: I think DefaultHost should be replaced in favor of the polymorphic version, therefore the defaults should be determined at runtime.
> 
> This is because on Linux the preferred audio API varies a lot, depending on the audio server installed (PulseAudio or Jack) or even drivers (some users may be using OSS).

Yes this is a good point, I've added a commit that makes this change.

PR #290, opened 2019-06-25
Remove macOS sleep loop in favour of using `Condvar`

mitchmindtree:
This solution was originally posted by @HybridEidolon in #185. Sorry it
took so long! I thought it might be easier to open a new PR as half of
your implementation here has already been implemented in a following PR
(namely, the change from an unnecessary `Vec` of callbacks to a single
user callback).

Closes #185.

If someone on macOS could double check this is working that would be
appreciated! Otherwise I'll merge soon anyway as the change should be
pretty simple/straightforward.


mitchmindtree:
Just tested locally on a friends MBP and all seems to be well!

Issue #156, opened 2017-06-06
Low latency support


Boscop:
Are there any plans for ASIO support? Or WASAPI exclusive mode?

mitchmindtree:
@Boscop any chance could you elaborate on the benefits of addingn ASIO or exclusive mode support for cpal contributors like myself who aren't familiar?

mitchmindtree:
@Boscop you might be interested in #204 and possibly #205. This has also been previously raised in #106.

roderickvd:
ASIO has been added. Exclusive mode: #106 and others.

PR #292, opened 2019-06-28
Add ASIO Host for Windows (rebase and refactor of #221)

mitchmindtree:
The refactor includes:

- The removal of most global state.
- Uses RAII for cleaner management of state transitions.
- Removes the hefty macros throughout the asio implementation.
- Adapts to the newly introduced `Host` API.

Support for ASIO has been gated behind an `asio` feature which is disabled by default. This is because the setup required for ASIO is fairly non-trivial and as it requires bindgen it introduces quite a few more dependencies.

A section has been added to the README describing in detail how to download, install and build CPAL with support for an ASIO host. Please see the more recent commit messages below for more details.

Closes #221.

cc @freesig
cc @Boscop this is related to #156 

Edit: Oh also, I've only been able to confirm the enumerate and beep examples work, though the beep example would distort heavily and cause my VM to crash. Originally I thought this might be my implementation, but I noticed this same behaviour happened in other software when running ASIO streams too. If anyone reading happens to have ASIO installed on a native system who could test the `beep`, `record_wav` and `feedback` examples that would be greatly appreciated!


ishitatsuyuki:
I think I have a native ASIO output (DJ controller), but I’m a bit short of time this week so please wait a bit!

freesig:
I can test this today

freesig:
Beep runs on windows. Sounds correct but doesn't respond to system volume but maybe that's an issue with this machine. I'll do a more thorough test soon.

ishitatsuyuki:
I don’t think ASIO is supposed to respond to system volume (at least the regular Windows one), am I getting something wrong?

ASIO devices often have dedicated software or hardware knobs.

mitchmindtree:
Thanks both!

Just a reminder to modify the examples when running them to use ASIO with:
```rust
// let host = cpal::default_host();
```
This is because when building with `--features "asio"` the default remains `Wasapi` - activating the feature just makes the `Asio` host available.

mitchmindtree:
Also I forgot to mention - when the `asio` feature is enabled, the `enumerate.rs` example `panic!`s after enumerating all ASIO devices, right before enumerating the WASAPI devices. It looks like this is due to [this](https://github.com/tomaka/cpal/blob/master/src/host/wasapi/com.rs#L11-L14), but I think this can be addressed in a following PR - I'm not yet sure what the API implications will be.

README.md
```diff
@@ -13,14 +13,90 @@ This library currently supports the following:
...
+WASAPI. To do so, follow these steps:
+
+1. **Download the ASIO SDK** `.zip` from [this
+   link](https://www.steinberg.net/en/company/developers.html). The version as
+   of writing this is 2.3.1.
+2. Extract the files and place the directory somewhere you are happy for it to stay
+   (e.g. `~/.asio`).
+3. Assign the full path of the directory (that contains the `readme`, `changes`,
+   `ASIO SDK 2.3` pdf, etc) to the `CPAL_ASIO_DIR` environment variable. [How to
+   set persisting Environment Variables on
```
ishitatsuyuki: I don't think it has to be persistent environment variable, I'd prefer the `$env:` PowerShell tip (that set it for the current session).

README.md
```diff
@@ -13,14 +13,90 @@ This library currently supports the following:
...
+6. If you don't have any ASIO devices or drivers available, you can [**download
+   and install ASIO4ALL**](http://www.asio4all.org/). Be sure to enable the
+   "offline" feature during installation despite what the installer says about
+   it being useless.
+7. **Loading VCVARS**. `rust-bindgen` uses the C++ tool-chain when generating
+   bindings to the ASIO SDK. As a result, it is necessary to load some
+   environment variables in the command prompt that we use to build our project.
+   On 64-bit machines run:
+   ```
+   "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\vcvarsall.bat" amd64
```
ishitatsuyuki, line 66: This version of Visual Studio is pretty old, recent versions have different naming.

README.md
```diff
@@ -13,14 +13,90 @@ This library currently supports the following:
...
+   of writing this is 2.3.1.
+2. Extract the files and place the directory somewhere you are happy for it to stay
+   (e.g. `~/.asio`).
+3. Assign the full path of the directory (that contains the `readme`, `changes`,
+   `ASIO SDK 2.3` pdf, etc) to the `CPAL_ASIO_DIR` environment variable. [How to
+   set persisting Environment Variables on
+   Windows](https://gist.github.com/mitchmindtree/92c8e37fa80c8dddee5b94fc88d1288b#file-windows_environment_variables-md).
+   The existence of the `CPAL_ASIO_DIR` environment variable determines whether
+   or not CPAL will attempt to build the ASIO SDK and generate bindings to it.
+4. **Download and install LLVM** from
```
ishitatsuyuki: It isn't very clear why LLVM is needed. Turns out it's for bindgen.

ishitatsuyuki:
Tested this with DDJ-RB ASIO. **Doesn't work.** A brief debugging shows that the callback isn't called at all.

mitchmindtree:
Thanks for trying @ishitatsuyuki! I've addressed your notes on the README.

Re no playback: I just to clarify, you're sure that PortAudio is using its ASIO host to interface with the device and not a WASAPI one? And that the `asio` feature was enabled and you made the host selection code modification to the examples that I mentioned above? Could you share the output of the `enumerate.rs` example using the `Asio` host? Sorry if these seem obvious, but I have to check as it's tricky to imagine why your callback is not being called when it is for myself and @freesig.

@freesig was able to get this running but noticed some conflicts when attempting to open both an input and output stream at once, but I've since fixed this issue.

We're also hearing some odd artifacts in the audio output stream but I'm finding it difficult to track down the cause. Initially I suspected it was the lack of support for ASIO in my Virtualbox VM but freesig shared a recording of the artifacts they're hearing and it sounds similar to what I'm hearing in my VM.

mitchmindtree:
Also these android CI bugs seems to be appearing from the latest version of the backtrace crate, I've opened up an issue [here](https://github.com/rust-lang/backtrace-rs/issues/220).

ishitatsuyuki:
```
Supported hosts:                                                                                                                                                                      [Asio, Wasapi]                                                                                                                                                                    Available hosts:                                                                                                                                                                      [Asio, Wasapi]                                                                                                                                                                    Asio                                                                                                                                                                                  Default Input Device:                                                                                                                                                                 Some("DDJ-RB ASIO")                                                                                                                                                               Default Output Device:                                                                                                                                                                Some("DDJ-RB ASIO")                                                                                                                                                               Devices:                                                                                                                                                                            1. "DDJ-RB ASIO"                                                                                                                                                                      Default input stream format:                                                                                                                                                          Format { channels: 2, sample_rate: SampleRate(44100), data_type: I16 }                                                                                                            All supported input stream formats:                                                                                                                                                   1.1. SupportedFormat { channels: 1, min_sample_rate: SampleRate(44100), max_sample_rate: SampleRate(44100), data_type: I16 }                                                        1.2. SupportedFormat { channels: 2, min_sample_rate: SampleRate(44100), max_sample_rate: SampleRate(44100), data_type: I16 }                                                      Default output stream format:                                                                                                                                                         Format { channels: 4, sample_rate: SampleRate(44100), data_type: I16 }                                                                                                            All supported output stream formats:                                                                                                                                                  1.1. SupportedFormat { channels: 1, min_sample_rate: SampleRate(44100), max_sample_rate: SampleRate(44100), data_type: I16 }                                                        1.2. SupportedFormat { channels: 2, min_sample_rate: SampleRate(44100), max_sample_rate: SampleRate(44100), data_type: I16 }                                                        1.3. SupportedFormat { channels: 3, min_sample_rate: SampleRate(44100), max_sample_rate: SampleRate(44100), data_type: I16 }                                                        1.4. SupportedFormat { channels: 4, min_sample_rate: SampleRate(44100), max_sample_rate: SampleRate(44100), data_type: I16 }                                                  Wasapi
```

ishitatsuyuki:
The driver calls `asioMessage` on initialization: this is currently not implemented and we return a **broken** value.

According to the log `selector: 7, vaule: 0` (typo intentionally same with the code), this maps to the following query:

```
kAsioSupportsTimeInfo,		// if host returns true here, it will expect the
							// callback bufferSwitchTimeInfo to be called instead
							// of bufferSwitch
```

And the driver then goes on to use `bufferSwitchTimeInfo` which is completely unimplemented either. Things are now broken.

asio-sys/build.rs
```diff
@@ -0,0 +1,193 @@
...
+use walkdir::WalkDir;
+
+const CPAL_ASIO_DIR: &'static str = "CPAL_ASIO_DIR";
+
+const ASIO_HEADER: &'static str = "asio.h";
+const ASIO_SYS_HEADER: &'static str = "asiosys.h";
+const ASIO_DRIVERS_HEADER: &'static str = "asiodrivers.h";
+
+fn main() {
+    // If ASIO directory isn't set silently return early
```
ishitatsuyuki, line 16: I think this behavior can create some confusion, because I just accidentally forgot to set and this created mysterious errors.

If we exited with an error code (or panicked), would it cause any problems to non-ASIO users?

asio-sys/src/bindings/mod.rs
```diff
@@ -0,0 +1,758 @@
...
+extern "C" fn sample_rate_did_change(_s_rate: c_double) -> () {
+    println!("sample rate changed");
+}
+
+/// Messages for ASIO
+/// This is not currently used
+extern "C" fn asio_message(
+    _selector: c_long, _value: c_long, _message: *mut (), _opt: *mut c_double,
+) -> c_long {
+    println!("selector: {}, vaule: {}", _selector, _value);
```
ishitatsuyuki: I think this file has got a lot of typos. Let's run a spell checker.

mitchmindtree: This `println` must have slipped in with the last commit! I'll remove this when addressing your other comment re handling asio_message.

asio-sys/build.rs
mitchmindtree: I think this is still left over from the original implementation and I haven't considered it properly since this large refactor.

I think the original intention was to allow for checking for ASIO availability downstream at run-time, so rather than failing to build, we assume the cpal user (or downstream crate user) wants to support it only if it is available on the system. That said, shouldn't the `asio` feature already cover this? I might have a think on this some more and get back to you.

mitchmindtree:
OK thank you for digging deeper and providing more info, this is useful!

It seems your driver is the first we've come across to support the TimeInfo buffer switch and asio message callback, as neither of these callbacks have been called by our drivers. I will review the ASIO sample implementation of `bufferSwitchTimeInfo` and `asioMessage` and see how well I can match the behaviour before pinging you for another test :pray:

mitchmindtree:
@ishitatsuyuki OK, I've addressed the typos and lack of handling of the timed buffer switching and messaging callbacks.

The implementation of, `asio_message`, `buffer_switch` and `buffer_switch_time_info` have been refactored to match the implementation demonstrated in the `hostsample.cpp` file that comes with the SDK. As in the sample, many of the messages kinds are still yet to be handled properly, however we should now at least be handling the time info version of the buffer switch callback properly.

asio-sys/src/bindings/mod.rs
```diff
@@ -0,0 +1,908 @@
...
+}
+
+/// Message callback for ASIO to notify of certain events.
+extern "C" fn asio_message(
+    selector: c_long,
+    value: c_long,
+    _message: *mut (),
+    _opt: *mut c_double,
+) -> c_long {
+    let mut ret = 0;
```
ishitatsuyuki: This looks rather weird, any reason not to use match *expressions* or return statement?

asio-sys/src/bindings/mod.rs
```diff
@@ -0,0 +1,908 @@
...
+                    ret = 1;
+                }
+                _ => (),
+            }
+        }
+
+        ai::kAsioResetRequest => {
+            // Defer the task and perform the reset of the driver during the next "safe" situation
+            // You cannot reset the driver right now, as this code is called from the driver. Reset
+            // the driver is done by completely destruct is. I.e. ASIOStop(), ASIODisposeBuffers(),
```
ishitatsuyuki: is -> it

asio-sys/src/bindings/mod.rs
```diff
@@ -0,0 +1,908 @@
...
+        }
+
+        _ => (), // Unknown/unhandled message type.
+    }
+
+    ret
+}
+
+/// Similar to buffer switch but with time info
+/// Not currently used
```
ishitatsuyuki: This is now used, this line can be deleted

src/platform/mod.rs
```diff
@@ -516,6 +526,16 @@ mod platform_impl {
 // TODO: Add `Asio asio` once #221 lands.
```
ishitatsuyuki: Let's remove this.

ishitatsuyuki:
How is it going regarding the artifact thing you mentioned before? Is it the one that happens after Ctrl+C, or is it during playback? I don't think I have experienced the latter. If it's still an issue, I think we should fix it somehow before merging.

mitchmindtree:
Thanks again @ishitatsuyuki, I've addressed those remaining nits!

> How is it going regarding the artifact thing you mentioned before? Is it the one that happens after Ctrl+C, or is it during playback?

The artifacts are occurring during playback, and yes they are still occurring. I guess a more correct description would be that output sounds distorted, but consistently so. Here's a recording of how the beep example sounds:

https://drive.google.com/file/d/1ObAlvTTvAo8dT-_j5IIkKfcR8v9rQ9Ua/view?usp=sharing

This was recorded using ASIO4ALL in my Windows VM. Originally, I would have disregarded this as an issue with the VM as Virtualbox does not have great support for ASIO at all. E.g. running other ASIO applications consistently crashes my entire VM (the CPAL examples when using ASIO do this also, but very rarely in comparison). However, @freesig sent me a recording that demonstrated similar behaviour, *also* using ASIO4ALL but on a native Windows machine.

Curiously though, input streams seem to be working perfectly fine. E.g. using a version of the `record_wav.rs` example modified to use ASIO, I was able to take 7 consecutive recordings, play them back in VLC and they sounded perfectly clean each time.

In our tests, it seems to only be via ASIO4ALL that this output stream distortion occurs. E.g. when using Dante Via, the beep example sounds clean as expected.

## Things I've Tried

I've read through the `asio-sys` source and asio host implementation multiple times now without any success of finding out what's going on. I've also ready through the portaudio ASIO host implementation and the `hostsample.cpp` that ships with the ASIO SDK and nothing stood out to me as a possible cause. Here are the things I've tried:

**Checking sample type conversions**

As ASIO4ALL requests the `ASIOSTInt32LSB` format, I thought the issue might be to do with the conversion to `i16` that takes place in order to be compatible with CPAL's API. To check this, I made a branch that adds `SampleFormat::I32` support (along with the necessary `Unknown*` buffer variants and `i32` `Sample` trait implementation) allowing us to avoid the need for sample type conversion altogether. However after all of this, the sine distortion sounded exactly the same. You can find this branch [here](https://github.com/mitchmindtree/cpal/tree/asio_add_i32_support). It should also be noted that the input stream seems perfectly capable of performing the reverse sample type conversion, so considering this along with the result of my `i32` sample format test I'm lead to believe sample type conversion cannot be the issue.

**Checking that sample type index is correct**

I verified that the sample type index does indeed match the index specified by the ASIO SDK (I must have checked this 5 times).

**Checking endianness**

I suspected there might be an incorrect endianness conversion somewhere, however changing the conversion to the other direction simply created silence and removing the conversion had no effect at all. This is to be expected as the LSB indicates ASIO is expecting little-endian data, which is the native endianness of Windows anyway. Thus, I've ruled out endianness as the potential problem.

**Checking channel layout**

The next possibility that crossed my mind was that perhaps our expected channel layout was wrong. E.g. perhaps there is some undocumented case where ASIO might deliver interleaved buffers? Or maybe our pointer offsets are wrong? To check this, I tried writing the beep example to only the first channel, and sure enough, only audio came out of the left speaker. Next, I tried writing to only the right channel, and again audio only came out of the right speaker. It seems that the data we are writing is indeed making it to the correct ASIO channels at least.

**Checking for clipping**

I ruled out the possibility that the beep example was clipping under the `I16` format by reducing the amplitude of the signal to an 8th of what it was before the sine wave callback returned. When I did this, the distorted sine wave sounded exactly as it did before, just quieter, so the cause of the distortion does not seem to be due to clipping.

---

## Summary

At this point, I'm running out of ideas and time to test them. I'm leaning towards merging this PR in its current state, opening an issue about the ASIO4ALL sine distortion and coming back to it once I get some time to do so in the future. This PR at least makes it possible for users to begin using ASIO and may inspire more contributions to the `Host` in the future.

ishitatsuyuki:
That explanation was very helpful! I think we can merge this then.

Ralith:
Why is bindgen being invoked at buildtime, incurring very heavy foreign dependencies, rather than invoked manually with the results committed?

ishitatsuyuki:
I think it's just a trade-off. Installing libclang on Windows is kinda annoying, but a brief look at bindgen [documentation](https://rust-lang.github.io/rust-bindgen/library-usage.html) suggests that it should be a build dependency.

Ralith:
Does asio actually have sufficiently complex platform dependent ABI as to
require this considerable increase in build complexity?

ishitatsuyuki:
I'm not sure I understand what you're trying to say.

First ASIO SDK ships with the source of client and driver libraries (as a source available license), and not binary. IIRC it uses COM, a well-defined protocol on Windows, so ABI issues are not a thing.

Second, running bindgen at build time is recommended by upstream, and I think the benefits it provide is worth the LLVM dependency. We don't need to commit generated file into repository, and having it as a part of build script makes it more straightforward to tune its options.

Installing LLVM is not that much work considering that ASIO SDK also has to be downloaded and put in to a environment variable.

Ralith:
The only benefit to running bindgen at buildtime, as explained by the documentation you linked, is robustness to complex platform-dependent variations in ABI. If the ABI is stable, there is no benefit, and there is a large cost. LLVM is much more troublesome than most foreign dependencies, because it is very large and does not have a stable ABI, making it painful under ideal conditions and a nightmare when the version required by bindgen is different than the version that is packaged for your environment.

mitchmindtree:
@Ralith I do feel your pain w.r.t. dependencies, LLVM requirements etc. I think I'd be open to a PR that commits src generated by the current bindgen setup. It would be nice to hold onto this asio binding generation code though as it contains a whitelist and might be useful if we require regenerating it in the future. Perhaps we could still include it in the repo as a `bin` crate that is not depended on by CPAL itself, but can be used by maintainers if necessary if some new API is added to ASIO or something?

Ralith:
Makes sense to me!

PR #295, opened 2019-06-29
Refactor `Host` and related traits into a new `traits` module

### LINKED ISSUE #294: Unique naming to distinguish between traits and platform types?
I'm currently having a go at updating come downstream code to the latest `Host` API changes and it has made the shared names between the traits and the platform hosts a little frustrating.

Specifically, the `cpal::{Host, Device, EventLoop, StreamId}` traits and the `cpal::platform::{Host, Device, EventLoop, StreamId}` types are often needed in the same module, which means either only importing the traits and using the types via the `cpal::platform::` namespace or importing both but importing either the platform types or the traits as some other name.

Also, seeing as the traits are in the crate root, it produces some confusing errors when updating code to the new API, as previously e.g. `cpal::EventLoop` would refer to a concrete type, whereas now it refers to a trait.

## Potential Solution

One solution that might make transitioning easier could be to move the traits into their own module (perhaps a `traits` module) and re-export the dynamically dispatched platform types from the crate root. This way, there should be a lot less breakage in existing user code, as the root will still contain all the default concrete types for `EventLoop`, `Device` and `StreamId`, and although their methods won't be immediately available the compiler should suggest importing the necessary traits.

This solution would require renaming the `Host`, `Device`, `EventLoop` and `StreamId` traits to something friendlier if we want to avoid naming import collisions with the concrete types. The first idea that comes to mind is `HostTrait`, `DeviceTrait`, etc, but I'm open to suggestions!

Another benefit of moving all traits into their own module would be that users could import them all at once `use cpal::traits::*;` without polluting the namespace with other items.

cc @ishitatsuyuki would be great to get your thoughts on this!

mitchmindtree:
This is a draft implementation of #294. I'll leave this open for
feedback and potentially better trait naming suggestions or better
solutions in general!

Closes #294.

cc @ishitatsuyuki


mitchmindtree:
Seeing as there has been no further feedback on this PR or at #294 I'm going to move on and merge this. Feel free to open a follow-up PR if you feel a better solution exists!

Issue #272, opened 2019-05-20
Popping/hitching in beep example


Ralith:
On Linux with alsa backed by pulse, I'm experiencing extremely erratic artifacts in the beep example. Sometimes they'll occur practically continuously, and sometimes they won't occur at all, but they're common on average and unmistakeable when you hear them.

swooster:
I've encountered the same thing on the master branch (0354548426f7b4c48392cae9876a3e77ea5c1c40) on windows 7... Sometimes I go several minutes without hitches, sometimes they're only a few seconds apart. I've tried checking if it's my audio by listening to a 1000hz sine wave on youtube, and that hasn't yet hitched once. Perhaps this is grasping at straws, but any hitches I encountered occurred when typing in Discord.

Ralith:
On current master, I can only reproduce this by e.g. aggressively cycling between windows while running the example; running heavy compiles in the background doesn't affect it. Other programs like mpv, however, don't glitch no matter what I try, so this remains an issue.

sniperrifle2004:
I wonder if this might be caused by the buffer not being filled up (#432)?

gintsgints:
I can confirm, this is still issue with my Mac M3, and Windows.

roderickvd:
This sounds like cpal being starved (in case of high CPU load) or not warmed up enough (in case of low CPU load) when set up with the default buffer configuration .Can you confirm still on cpal v0.17 or latest master? There have been significant improvements to buffering.

gintsgints:
I think for me problem was, - I created handle in thread and read file in thread, which was dropt after beep. It worked much better when thread played raw data - https://github.com/gintsgints/raiti/blob/master/src/beeper.rs

PR #301, opened 2019-07-09
Getting rid of the EventLoop

### LINKED ISSUE #278: Getting rid of the EventLoop

ishitatsuyuki:
Close #278. I rushed for a buildable version so you can see it, there's still a lot of thing to do.

- [ ] Check for `can_pause` and optionally return an error
- [ ] Polymorphic struct versions of Traits will be explicitly `!Sync`, `!Send`
- [x] Take an additional error callback as described in https://github.com/tomaka/cpal/issues/268#issuecomment-508905225
- [ ] Update documentation
- [ ] Checkbox for ASIO backend so I don't miss it because it's without CI

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 .cargo/
 .DS_Store
 recorded.wav
+rls*.log
--- a/examples/beep.rs
+++ b/examples/beep.rs
@@ -1,37 +1,34 @@
 extern crate cpal;
 extern crate failure;
 
-use cpal::traits::{DeviceTrait, EventLoopTrait, HostTrait};
+use cpal::traits::{DeviceTrait, StreamTrait, HostTrait};
 
 fn main() -> Result<(), failure::Error> {
     let host = cpal::default_host();
     let device = host.default_output_device().expect("failed to find a default output device");
     let format = device.default_output_format()?;
-    let event_loop = host.event_loop();
-    let stream_id = event_loop.build_output_stream(&device, &format)?;
-    event_loop.play_stream(stream_id.clone())?;
-
     let sample_rate = format.sample_rate.0 as f32;
+    let channels = format.channels;
     let mut sample_clock = 0f32;
 
     // Produce a sinusoid of maximum amplitude.
-    let mut next_value = || {
+    let mut next_value = move || {
         sample_clock = (sample_clock + 1.0) % sample_rate;
         (sample_clock * 440.0 * 2.0 * 3.141592 / sample_rate).sin()
     };
 
-    event_loop.run(move |id, result| {
+    let stream = device.build_output_stream(&format, move |result| {
         let data = match result {
             Ok(data) => data,
             Err(err) => {
-                eprintln!("an error occurred on stream {:?}: {}", id, err);
+                eprintln!("an error occurred on stream: {}", err);
                 return;
             }
         };
 
         match data {
             cpal::StreamData::Output { buffer: cpal::UnknownTypeOutputBuffer::U16(mut buffer) } => {
-                for sample in buffer.chunks_mut(format.channels as usize) {
+                for sample in buffer.chunks_mut(channels as usize) {
                     let value = ((next_value() * 0.5 + 0.5) * std::u16::MAX as f32) as u16;
                     for out in sample.iter_mut() {
                         *out = value;
@@ -39,7 +36,7 @@ fn main() -> Result<(), failure::Error> {
                 }
             },
             cpal::StreamData::Output { buffer: cpal::UnknownTypeOutputBuffer::I16(mut buffer) } => {
-                for sample in buffer.chunks_mut(format.channels as usize) {
+                for sample in buffer.chunks_mut(channels as usize) {
                     let value = (next_value() * std::i16::MAX as f32) as i16;
                     for out in sample.iter_mut() {
                         *out = value;
@@ -47,7 +44,7 @@ fn main() -> Result<(), failure::Error> {
                 }
             },
             cpal::StreamData::Output { buffer: cpal::UnknownTypeOutputBuffer::F32(mut buffer) } => {
-                for sample in buffer.chunks_mut(format.channels as usize) {
+                for sample in buffer.chunks_mut(channels as usize) {
                     let value = next_value();
                     for out in sample.iter_mut() {
                         *out = value;
@@ -56,5 +53,10 @@ fn main() -> Result<(), failure::Error> {
             },
             _ => (),
         }
-    });
+    })?;
+    stream.play()?;
+    
+    std::thread::sleep(std::time::Duration::from_millis(1000));
+
+    Ok(())
 }
--- a/examples/beep.rs
+++ b/examples/beep.rs
@@ -17,15 +17,7 @@ fn main() -> Result<(), failure::Error> {
         (sample_clock * 440.0 * 2.0 * 3.141592 / sample_rate).sin()
     };
 
-    let stream = device.build_output_stream(&format, move |result| {
-        let data = match result {
-            Ok(data) => data,
-            Err(err) => {
-                eprintln!("an error occurred on stream: {}", err);
-                return;
-            }
-        };
-
+    let stream = device.build_output_stream(&format, move |data| {
         match data {
             cpal::StreamData::Output { buffer: cpal::UnknownTypeOutputBuffer::U16(mut buffer) } => {
                 for sample in buffer.chunks_mut(channels as usize) {
@@ -53,6 +45,8 @@ fn main() -> Result<(), failure::Error> {
             },
             _ => (),
         }
+    }, move |err| {
+        eprintln!("an error occurred on stream: {}", err);
     })?;
     stream.play()?;
     
--- a/examples/feedback.rs
+++ b/examples/feedback.rs
@@ -44,15 +44,7 @@ fn main() -> Result<(), failure::Error> {
 
     // Build streams.
     println!("Attempting to build both streams with `{:?}`.", format);
-    let input_stream = input_device.build_input_stream(&format, move |result| {
-        let data = match result {
-            Ok(data) => data,
-            Err(err) => {
-                eprintln!("an error occurred on input stream: {}", err);
-                return;
-            },
-        };
-
+    let input_stream = input_device.build_input_stream(&format, move |data| {
         match data {
             cpal::StreamData::Input {
                 buffer: cpal::UnknownTypeInputBuffer::F32(buffer),
@@ -69,15 +61,10 @@ fn main() -> Result<(), failure::Error> {
             },
             _ => panic!("Expected input with f32 data"),
         }
+    }, move |err| {
+        eprintln!("an error occurred on input stream: {}", err);
     })?;
-    let output_stream = output_device.build_output_stream(&format, move |result| {
-        let data = match result {
-            Ok(data) => data,
-            Err(err) => {
-                eprintln!("an error occurred on output stream: {}", err);
-                return;
-            },
-        };
+    let output_stream = output_device.build_output_stream(&format, move |data| {
         match data {
             cpal::StreamData::Output {
                 buffer: cpal::UnknownTypeOutputBuffer::F32(mut buffer),
@@ -98,6 +85,8 @@ fn main() -> Result<(), failure::Error> {
             },
             _ => panic!("Expected output with f32 data"),
         }
+    }, move |err| {
+        eprintln!("an error occurred on output stream: {}", err);
     })?;
     println!("Successfully built streams.");
 
--- a/examples/feedback.rs
+++ b/examples/feedback.rs
@@ -9,29 +9,26 @@
 extern crate cpal;
 extern crate failure;
 
-use cpal::traits::{DeviceTrait, EventLoopTrait, HostTrait};
+use cpal::traits::{DeviceTrait, HostTrait, StreamTrait};
 
 const LATENCY_MS: f32 = 150.0;
 
 fn main() -> Result<(), failure::Error> {
     let host = cpal::default_host();
-    let event_loop = host.event_loop();
 
     // Default devices.
-    let input_device = host.default_input_device().expect("failed to get default input device");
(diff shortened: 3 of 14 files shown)


ishitatsuyuki:
Examples are updated.

ishitatsuyuki:
The ALSA backend is ready for review.

ishitatsuyuki:
cc @mitchmindtree

mitchmindtree:
Nice @ishitatsuyuki! I've only had a chance to look at the examples so far and I think they're looking great!

I haven't had a chance to dig into the ALSA implementation just yet, but in the meantime a couple things that come to mind are:

- I wonder if we should we make sure that `Stream`s are `Sync` and `Send` for each platform? This is something that wasn't super clear to me w.r.t. the `EventLoop`. From memory, most hosts had an `unsafe impl Send`/`Sync` but for the most part it has been undocumented and none of the impls explained why it was safe. I think it's nice from a usability point of view, but 1. we should check that it's possible to do this in a safe manner for each platform and 2. maybe we should add a small test that tries to pass a `Stream` to a empty function that expects `T: Send + Sync`? I hope this isn't derailing this PR too much, it's just something that might be worth clearing up early as I imagine it might affect implementation details for newer `Stream` impls.
- I want to make sure we have a clear path ahead for handling the web audio API. Do you think it will be possible to spawn both an input and output stream with the `stdweb` restrictions mentioned [here](https://github.com/tomaka/cpal/issues/245#issuecomment-437141514)? My understanding of the restriction is still a little vague - I guess the main concern is that ensuring building a stream does not block the entire thread like `EventLoop::run` currently does due to calling `stdweb::event_loop`.

I will aim to do a more thorough review of the ALSA backend soon!

ishitatsuyuki:
> I wonder if we should we make sure that Streams are `Sync` and `Send` for each platform?

No, I intentionally exclude some of the platforms. `Sync` and `Send` are only implemented if you are allowed to access things within the callback, which rules out Android.

In general, the synchronization story is unclear on many platforms, and I think it's the best to code in a style that uses message passing to serialize access to the underlying API.

> Do you think it will be possible to spawn both an input and output stream with the stdweb restrictions

My stance doesn't change from before, but I think that there is no problem doing that. Effectively we submit a task to the web event loop, instead of spawning a thread, and that perfectly fits into how the web API was supposed to work.

And a disclaimer: I made some pretty wild refactor when modifying the ALSA backend (the code moved isn't preserved as-is), sorry if it seems hard to review.

ishitatsuyuki:
Rebased, but... I'm still struggling to motivate myself to work on the other backends.

gentoid:
I may try to help with Windows platform. I have a machine with Win 10. Though please expect that there may be many questions :-D

ishitatsuyuki:
This is my first try at writing mentoring instructions, sorry if it's not detailed enough! Feel free to ask any questions.

So the API changes are like this:
- StreamId is removed; the internal loop no longer handles multiple streams
- EventLoop is removed; `build_input/output_stream` are moved onto `Device`, `play/pause` are moved onto a new `Stream` struct (and trait)
- The callback methods are now passed to `build_input/output_stream` and they are now two different methods for data/error instead of one.

We first start with changing the main loop:

1. Change `RunContext` so it only contains one `Arc<StreamInner>`, not the Vec.
https://github.com/RustAudio/cpal/blob/9e3a19c352bf7de8a7775a940b2a233244706b70/src/host/wasapi/stream.rs#L58

1. Move the `run_inner` function out as a free function. The signature should be `fn run_inner(run_context: RunContext, data_callback: &mut dyn FnMut(StreamData), error_callback: &mut dyn FnMut(StreamResult)` (returns `()` instead of `!`).
https://github.com/RustAudio/cpal/blob/9e3a19c352bf7de8a7775a940b2a233244706b70/src/host/wasapi/stream.rs#L453

1. Delete these lines.
https://github.com/RustAudio/cpal/blob/9e3a19c352bf7de8a7775a940b2a233244706b70/src/host/wasapi/stream.rs#L455-L465
https://github.com/RustAudio/cpal/blob/9e3a19c352bf7de8a7775a940b2a233244706b70/src/host/wasapi/stream.rs#L468-L478

1. Whenever you encounter a long, old error handling template, replace it with `error_callback(err); break 'stream_loop`. You may need to use `into`, remove `Err(..)` (because callback no longer takes a `Result`) so it compiles.

1. There is only one stream, so delete these *two* lines.
https://github.com/RustAudio/cpal/blob/9e3a19c352bf7de8a7775a940b2a233244706b70/src/host/wasapi/stream.rs#L502-L503

1. Remove this line.
https://github.com/RustAudio/cpal/blob/master/src/host/wasapi/stream.rs#L627

1. Also alter `process_commands`. Remove the `NewStream` and `DestroyStream` variant as described below, and change the error handling as described above.

And now let's go on to the other API changes:

1. Create the new `Stream` struct. It should contain:
  - `thread: Option<JoinHandle<()>>`, this is the thread that we create with this new API.
  - `commands`, `pending_scheduled_event`: copy these from `EventLoop`.

1. Remove the `NewStream` and `DestroyStream` variant from `Command`.

1. Move the `build_input/output_stream` methods onto `Device`. You need to alter function signature and move referenced function (some of used functions are in stream.rs). When constructing a Stream, create a `std::thread::Thread` that indirectly calls `run_inner`, and use the `JoinHandle` returned. The construction of other fields should be copied from `EventLoop::new`.  
For how to create the thread, see the ALSA implementation for an example:
https://github.com/RustAudio/cpal/blob/0ee93de213adfb793ba3e5e331cce8e0139b1654/src/host/alsa/mod.rs#L706-L708

1. Implement `Drop` for `Device`. It should contain two parts: one is from `EventLoop`
https://github.com/RustAudio/cpal/blob/9e3a19c352bf7de8a7775a940b2a233244706b70/src/host/wasapi/stream.rs#L653-L660
And another to join the thread, like the ALSA implementation.
https://github.com/RustAudio/cpal/blob/0ee93de213adfb793ba3e5e331cce8e0139b1654/src/host/alsa/mod.rs#L719-L720
Use this snippet instead of `trigger.wakeup()`.
https://github.com/RustAudio/cpal/blob/9e3a19c352bf7de8a7775a940b2a233244706b70/src/host/wasapi/stream.rs#L646-L649

At this point I think I've covered most of the changes. There might be other compile errors so fix them if they're straightforward. I'll be ready to answer questions otherwise.

gentoid:
woah, great writeup! This will help a lot! Thanks

gentoid:
Just to let you know. I'm working on that, and the instruction is really helpful 👍

gentoid:
So I fixed the API regarding the instruction, then fixed some compile errors. But there's one left I cannot fix:
```
error[E0277]: `*mut host::wasapi::winapi::um::audioclient::IAudioClient` cannot be sent between threads safely
  --> src\host\wasapi\stream.rs:96:22
   |
96 |         let thread = thread::spawn(move || {
   |                      ^^^^^^^^^^^^^ `*mut host::wasapi::winapi::um::audioclient::IAudioClient` cannot be sent between threads safely
   |
   = help: within `host::wasapi::stream::StreamInner`, the trait `std::marker::Send` is not implemented for `*mut host::wasapi::winapi::um::audioclient::IAudioClient`
   = note: required because it appears within the type `host::wasapi::stream::StreamInner`
   = note: required because of the requirements on the impl of `std::marker::Send` for `std::sync::Arc<host::wasapi::stream::StreamInner>`
   = note: required because it appears within the type `host::wasapi::stream::RunContext`
   = note: required because it appears within the type `[closure@src\host\wasapi\stream.rs:96:36: 98:10 run_context:host::wasapi::stream::RunContext, data_callback:D, error_callback:E]`
   = note: required by `std::thread::spawn`

error[E0277]: `*mut std::ffi::c_void` cannot be sent between threads safely
  --> src\host\wasapi\stream.rs:96:22
   |
96 |         let thread = thread::spawn(move || {
   |                      ^^^^^^^^^^^^^ `*mut std::ffi::c_void` cannot be sent between threads safely
   |
   = help: within `host::wasapi::stream::StreamInner`, the trait `std::marker::Send` is not implemented for `*mut std::ffi::c_void`
   = note: required because it appears within the type `host::wasapi::stream::StreamInner`
   = note: required because of the requirements on the impl of `std::marker::Send` for `std::sync::Arc<host::wasapi::stream::StreamInner>`
   = note: required because it appears within the type `host::wasapi::stream::RunContext`
   = note: required because it appears within the type `[closure@src\host\wasapi\stream.rs:96:36: 98:10 run_context:host::wasapi::stream::RunContext, data_callback:D, error_callback:E]`
   = note: required by `std::thread::spawn`
```

I need your help with that.
Here's the branch with my changes added https://github.com/RustAudio/cpal/compare/master...gentoid:no-eventloop

ishitatsuyuki:
@gentoid Much appreciate the contribution! I merged your branch and fixed compilation issues.

Here's an explanation of what I changed:

- `Arc<StreamInner>` -> `StreamInner`. My mistake, we didn't need Arc here because we send command to the event loop instead of directly manipulating the stream.
- `unsafe impl Send for RunContext`: this is the bare minimum to fix the Send error you mentioned.
- Visibility changes. Since most structs here are only used internally, it's perfectly fine to use `pub` as long as we don't reexport them in mod.rs.
- Some mutability and borrow-checker fix.

ishitatsuyuki:
I haven't tested the WASAPI backend, I will get to that next time I use Windows if you don't beat me to it.

gentoid:
I'm actually happy to contribute! Maybe I'll have a chance to have a look at it today. Thank you!

gentoid:
I've fixed all the warning raised by Clippy in the `wasapi` module - pushed to my branch. Except `mem::uninitialized`, I saw there's an issue there for that.

Clippy also throws 2 errors (while `cargo check` doesn't):

```
    //     error: this comparison involving the minimum or maximum element for this type contains a case that is always true or always false
    //    --> src\host\wasapi\stream.rs:227:19
    //     |
    // 227 |     debug_assert!(result >= winbase::WAIT_OBJECT_0);
    //     |                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    //     |
    //     = note: `#[deny(clippy::absurd_extreme_comparisons)]` on by default
    //     = help: because winbase::WAIT_OBJECT_0 is the minimum value for this type, this comparison is always true
    //     = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#absurd_extreme_comparisons


error: casting from `*const [u8; 16]` to a more-strictly-aligned pointer (`*const usize`) (1 < 8 bytes)
   --> src\host\wasapi\device.rs:361:37
    |
361 |             let ptr_usize: usize = *(&property_value.data as *const _ as *const usize);
    |                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: `#[deny(clippy::cast_ptr_alignment)]` on by default
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#cast_ptr_alignment
```

gentoid:
I guess I could help with ASIO too. Will try to update it using the same instruction

ishitatsuyuki:
Merged that.

I removed the first assertion warned by clippy. We shouldn't rely on Windows constants being a particular value, but in this case the debug assertion is useless because a subtraction happens afterward which is checked in debug. So this justify the change.

Regarding the second alignment error, this is pretty much a hack and I think we'd better leave it as-is. I looked up a bit and it seems that API is old so hopefully we can replace that with calls to newer APIs that requires less hacky conversion.

If you can help with the ASIO backend too, great and thanks a lot! I'll check later if there's anything that need special handling.

ishitatsuyuki:
I fixed the WASAPI backend to the point that beep works. Others are unverified.

crusadergo:
Hi there! I'm newbie with Rust and i have Mac machine. I'd like to help you with this cpal project, because I believe, it will help me to learn Rust more deeply :)

ishitatsuyuki:
I finished the CoreAudio backend myself since it's quite different from other platforms and somewhat more simple to do.

@crusadergo Sorry I've already finished the code, but if you can please help with testing! Checkout the PR as described [here](https://help.github.com/en/articles/checking-out-pull-requests-locally) and then run `cargo run --example beep` (and maybe other examples too) to verify that the code works.

icefoxen:
Tested the current branch on Windows 10, it seems to work but with caveats.  The beep example works fine, trying to port it to a `ggez` program produces something sounding Wrong, maybe more sawtooth-y.  May well be my program doing something wrong in terms of timing, code is [here](https://hg.sr.ht/~icefox/blep/) for reference.  I'll play with giving the playing sound its own thread or something.

Tested on Linux, on Ubuntu it seems to work fine (even my `ggez` program), on Debian the no-eventloop `beep` example works fine-ish but seems to occasionally stutter a little on startup (haven't checked if the old one did that).  On Debian my `blep` example doesn't output sound at all (though PulseAudio's equalizer shows it outputting).  NO IDEA where to start debugging that.

ishitatsuyuki:
@icefoxen I inspected your code and it seems to be calling `sin` in a nested manner: is that what you meant to write?

When getting no audio, the thing I usually do is to insert a debug println inside the callback. Although if you're getting mixer indicators that might be useless.

Ralith:
> on Debian the no-eventloop beep example works fine-ish but seems to occasionally stutter a little on startup (haven't checked if the old one did that).

The beep example has been glitchy for ages; see #272.

icefoxen:
>  it seems to be calling sin in a nested manner: is that what you meant to write?

It was, until I'd forgotten I'd written that way.  Unfortunately, fixing it only makes the sound differently wrong.  XD  I'll investigate more later when I have time.

ishitatsuyuki:
Closing in favor of #354.

Issue #296, opened 2019-07-03
Update compiler for the Android builder


ishitatsuyuki:
`backtrace-rs` is requiring a newer version of compiler, per https://github.com/rust-lang/backtrace-rs/issues/220.

The Android builder is relying on a private image with unknown maintenance status:

https://github.com/tomaka/cpal/blob/master/.circleci/config.yml#L8

cc @tomaka Are you interested in updating the image?

LastExceed:
This can be closed, see the above linked PR

PR #305, opened 2019-07-18
Disable Android CI target

ishitatsuyuki:
The Android build has been broken for a while; we don't have cfg-gated code for Android anyway for now.


mitchmindtree:
Yeah, we can always re-add if we get a response at #296.

PR #307, opened 2019-07-24
Remove the (arbitary) restriction on buffer size

derekdreery:
For me, cpal fails with EINVAL when cpal tries to set hw_params. I've found that relaxing the buffer size constraint solves the problem for me.

### DIFF
--- a/src/host/alsa/mod.rs
+++ b/src/host/alsa/mod.rs
@@ -1117,19 +1117,6 @@ unsafe fn set_hw_params_from_format(
         return Err(format!("channel count could not be set: {}", e));
     }
 
-    // TODO: Review this. 200ms seems arbitrary...
-    /*
-    let mut max_buffer_size = format.sample_rate.0 as alsa::snd_pcm_uframes_t /
-        format.channels as alsa::snd_pcm_uframes_t /
-        5; // 200ms of buffer
-    if let Err(e) = check_errors(alsa::snd_pcm_hw_params_set_buffer_size_max(pcm_handle,
-                                                             hw_params.0,
-                                                             &mut max_buffer_size))
-    {
-        return Err(format!("max buffer size could not be set: {}", e));
-    }
-    */
-
     if let Err(e) = check_errors(alsa::snd_pcm_hw_params(pcm_handle, hw_params.0)) {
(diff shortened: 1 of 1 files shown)


derekdreery:
This would be good for me to land as currently I have to patch cpal to build any rust project that relies on it (there are a lot).

derekdreery:
> Commenting code out is not desirable as a patch. Version control systems have their purpose, so please remove the code you think is obsolete instead.

Happy to do this, are you ok with the patch in principal?

ishitatsuyuki:
The patch looks fine to me. @mitchmindtree Can you review and squash-merge?

derekdreery:
I can squash commits if you like

ishitatsuyuki:
No it’s fine, just have to wait on @mitchmindtree because I don’t have write access

mitchmindtree:
Sorry about the delay - thanks for the PR @derekdreery and @ishitatsuyuki for the review!

ishitatsuyuki:
I think this is leading to libalsa-pulse using the maximum buffer size (which takes a lot of time to fill), so this probably need a follow-up.

alexheretic:
This changes seems to cause all sound to be delayed for me.

derekdreery:
Hmm, ok maybe the solution is to use `set_buffer_size_near`, since I think this always succeeds (I need to check this)

alexheretic:
With the 200ms max/near I get 
```
[src/host/alsa/mod.rs:892] buffer_len = 8820
```

Without anything I get the delay with
```
[src/host/alsa/mod.rs:892] buffer_len = 1048576
```

ishitatsuyuki:
Yeah, I think PulseAudio is the worst offender here. Other implementations also contain a correction for that: https://github.com/mpv-player/mpv/blob/cea4ff3e5fb5766d34f068742d00b127fa66013f/audio/out/ao_alsa.c#L800

alexheretic:
I'm having to revert this to use master because of the large delay it causes. It works fine for me using `set_buffer_size_near` instead of max, so can we put that back in?

ishitatsuyuki:
@alexheretic file a PR and I will merge.

PR #313, opened 2019-08-04
Make sure mul happens before div

derekdreery:
otherwise the expression evaluates to 0

### DIFF
--- a/src/host/emscripten/mod.rs
+++ b/src/host/emscripten/mod.rs
@@ -173,7 +173,7 @@ impl EventLoop {
                         None => continue,
                     };
 
-                    let mut temporary_buffer = vec![0.0; 44100 * 2 / 3];
+                    let mut temporary_buffer = vec![0.0; (44100 * 2) / 3];
 
                     {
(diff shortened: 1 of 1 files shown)


ishitatsuyuki:
This doesn't look right, and even feels malicious. Can you check again what you were trying to do?

derekdreery:
So the expression `44100 * 2 / 3` evaluates to `0` becuase the interger division `2 / 3 = 0`. This means that the statement is equivalent to `Vec::new`. With `(44100 * 2) / 3` the expression correctly evaluates to (sample_rate * channels) / frequency

ishitatsuyuki:
Nah, that's not how math operators work. https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=29abc9ef177c7ae93e790b29aeae8631

derekdreery:
You are right.

Gosh I've been thinking that operators worked differently for decades!

https://en.wikipedia.org/wiki/Operators_in_C_and_C%2B%2B#Operator_precedence

PR #314, opened 2019-08-04
Update mod.rs

### LINKED ISSUE #309: error should cause panic/abort
https://github.com/tomaka/cpal/blob/a04d869eaa0790a739c6ebbe80aeee949e529060/src/host/alsa/mod.rs#L803

This should be a panic/abort, not an error. This is because if the user tries to carry on after this error then the counter will wrap around to 0 and begin overwriting existing channels, possibly leading to anything including UB.

It was a panic until the error refactoring in 0.9.

derekdreery:
Fixes #309


est31:
I'm not really sure whether a panic will suffice as the panic only ends the specific thread, not the entire process. But IDK what a good solution would be.

PR #321, opened 2019-08-12
Fix WASAPI capture logic

### LINKED ISSUE #299: Audio capture hangs on Windows 10
I made a modified version of record.rs example: https://gist.github.com/redbluescreen/ac5f0b1dde5318134ec601bec8b81d1f
It makes the recording duration longer and prints out an increasing counter.

For me the counter often stops at random. It seems to be easier to reproduce while there's also music playing at the same time.
Debugger shows that it's hanging in src/host/wasapi/stream.rs wait_for_handle_signal at WaitForMultipleObjectsEx
I reproduced it on both cpal 0.10.0 and 0.9.0.
Windows version: Windows 10 1903

ishitatsuyuki:
The buffer size handling differs from what render streams uses. Logic is based on https://github.com/microsoft/Windows-universal-samples/blob/master/Samples/WindowsAudioSession/cpp/WASAPICapture.cpp.

Fix #299

### DIFF
--- a/src/host/wasapi/stream.rs
+++ b/src/host/wasapi/stream.rs
@@ -502,73 +502,84 @@ impl EventLoop {
                 let stream_idx = handle_idx - 1;
                 let stream = &mut run_context.streams[stream_idx];
 
-                // The number of frames available for reading/writing.
-                let mut frames_available = match get_available_frames(stream) {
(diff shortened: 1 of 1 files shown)


ishitatsuyuki:
(the ci thing is because I somehow based my changes on an older commit, this should be good to merge.)

PR #326, opened 2019-08-31
Add the ability to read inner values from input and output buffers

MOZGIII:
I need the ability to get a slice with the proper lifespan (`'a`). In the next major release we'll be switching to plain slices instead of wrapper `InputBuffer`/`OutputBuffer` types, so it's not a problem.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -515,6 +515,15 @@ impl<'a, T> Deref for InputBuffer<'a, T>
     }
 }
 
+impl<'a, T> InputBuffer<'a, T>
+    where T: Sample
+{
+    #[inline]
+    pub fn inner(&'a self) -> &'a [T] {
+        self.buffer
+    }
+}
+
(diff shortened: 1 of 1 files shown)


MOZGIII:
Ok, this doesn't do what I expect it to do.

I pretty much just need to implement this:

```rust
fn convert<'a>(buf: &'a cpal::UnknownTypeInputBuffer<'_>) -> &'a [u16] {
    match buf {
        cpal::UnknownTypeInputBuffer::U16(buffer) => *buffer,
        _ => panic!("Unexpected buffer type"),
    }
}
```

How can I do that?

MOZGIII:
Nvm, I got confused with code surrounding this.

Correct solution:
```
fn convert<'a>(buf: cpal::UnknownTypeInputBuffer<'a>) -> &'a [u16] {
    match buf {
        cpal::UnknownTypeInputBuffer::U16(buffer) => &*buffer,
        _ => panic!("Unexpected buffer type"),
    }
}
```

PR #327, opened 2019-08-31
Alsa: Set buffer size near to 100ms

alexheretic:
Not setting the buffer size can cause an overlarge buffer to be used. This can causes noticeable audio delays. See #307 discussion.

### DIFF
--- a/src/host/alsa/mod.rs
+++ b/src/host/alsa/mod.rs
@@ -1116,6 +1116,16 @@ unsafe fn set_hw_params_from_format(
         return Err(format!("channel count could not be set: {}", e));
     }
 
+    // If this isn't set manually a overlarge buffer may be used causing audio delay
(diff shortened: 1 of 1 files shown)


alexheretic:
> Please use snd_pcm_hw_params_set_buffer_time_near instead.
> 
> 200ms is arbitrary; I will leave it up to you but for the reference, mpv uses 100ms.

Cool that usage seems cleaner. Actually using 100ms produces the same buffer as the previously calculated "200ms", so this isn't a change (and I guess it was 100 all along?).

alexheretic:
@derekdreery does this work for your use case that didn't with 'max'?

ishitatsuyuki:
I'm pretty sure it will. Thanks for the PR!

derekdreery:
I'll give it a go once I'm back home. It should do though.

PR #328, opened 2019-09-05
Travis: enable Cargo cache

ishitatsuyuki:
Faster is better, let's enable it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,8 @@
 language: rust
 sudo: false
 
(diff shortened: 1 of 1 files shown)


ishitatsuyuki:
These kind of redundant object files might increase when you change the dependency or feature set.

I don’t think it’s a big deal though, our compilation times are low and it’s unlikely that the size will grow out of control. Every Rust update also leads to a purge I think.

PR #330, opened 2019-09-25
Use apartment threading on Windows

RoyJacobs:
I'm trying to play back some music using rodio (and cpal) and create a window using winit.

I can choose to initialize rodio and winit in two ways:

### Initializing rodio first, then initializing winit
This gives me a panic in winit [here](https://github.com/rust-windowing/winit/blob/master/src/platform_impl/windows/window.rs#L85), because rodio/cpal already initialized COM.

### Initializing winit first, then initializing rodio
This gives me a panic in cpal [here](https://github.com/RustAudio/cpal/blob/master/src/host/wasapi/com.rs#L13), with the error `Os { code: -2147417850, kind: Other, message: "Cannot change thread mode after it is set." }`.

# Fix
The root cause seems to be that winit is initializing COM with `COINIT_APARTMENTTHREADED` whereas cpal initializes COM with `COINIT_MULTITHREADED`.

According to the [docs](https://docs.microsoft.com/en-us/windows/win32/learnwin32/initializing-the-com-library) apartment threading can work if multiple threads don't access the same COM object. I believe this is true in cpal? Using apartment threading effectively isolates the COM use from winit from the COM use in cpal, which seems to be what we want.

I can confirm that (in my tests, at least) after this change everything works correctly, regardless of the winit/rodio initialization order.

### DIFF
--- a/src/host/wasapi/com.rs
+++ b/src/host/wasapi/com.rs
@@ -3,14 +3,14 @@
 use super::check_result;
 use std::ptr;
 
-use super::winapi::um::objbase::{COINIT_MULTITHREADED};
+use super::winapi::um::objbase::{COINIT_APARTMENTTHREADED};
 use super::winapi::um::combaseapi::{CoInitializeEx, CoUninitialize};
 
 thread_local!(static COM_INITIALIZED: ComInitialized = {
     unsafe {
         // this call can fail if another library initialized COM in single-threaded mode
         // handling this situation properly would make the API more annoying, so we just don't care
-        check_result(CoInitializeEx(ptr::null_mut(), COINIT_MULTITHREADED)).unwrap();
+        check_result(CoInitializeEx(ptr::null_mut(), COINIT_APARTMENTTHREADED)).unwrap();
         ComInitialized(ptr::null_mut())
     }
 });


ishitatsuyuki:
I think this change is fair, although I have several concerns.

1. `cpal`'s `EventLoop` blocks the thread it's running on (and never returns), so people doesn't seem to have used `cpal` on their main thread. Although I think it's a valid use case to enumerate devices in those cases.
2. We're moving to an API where cpal spawns and manage threads (#301). If we're going with apartment threading, this means that we will need to implement marshaling and review all Send/Sync impls.

RoyJacobs:
As for point 1, it doesn't have to be on the main thread: Rodio also just uses cpal on a background thread. However, the underlying issue is that COM cannot be initialized in different threading modes.

For point 2, yes, any COM object access will need to happen from the same thread that created the object. But multiple threads can each get their own COM object, I believe, so you don't need to route all COM access to a single thread.

ishitatsuyuki:
> However, the underlying issue is that COM cannot be initialized in different threading modes.

This restriction applies per-thread. You can initialize the GUI thread with apartment threading while the rodio thread being MTA.

> But multiple threads can each get their own COM object, I believe, so you don't need to route all COM access to a single thread.

We initialize the audio objects on the main thread then pass it on to the worker thread, which is why it might be a problem.

RoyJacobs:
Hmm, okay, but if it currently fails it would mean that somehow rodio does not initialize cpal on a separate thread (and also winit is not initializing com on a separate thread then)?

ishitatsuyuki:
rodio spawns the worker thread then perform all cpal calls on it.

RoyJacobs:
But you wrote "You can initialize the GUI thread with apartment threading while the rodio thread being MTA", but isn't that then what is currently already happening (and failing)?

ishitatsuyuki:
I have a game that uses both winit and rodio and it runs without problems. It works when ran on separate threads, I guess you are misunderstanding something.

RoyJacobs:
Are you perhaps using a different version of winit? I am using winit 0.20.0 alpha 3, which seems to be quite heavily rewritten from earlier versions.

RoyJacobs:
It's indeed only an issue with the 0.20 alpha of winit. I've added a reproduction case here: https://github.com/RoyJacobs/cpal-winit-repro

I'll also open a ticket with winit, maybe the authors can shed their light on this as well.

ishitatsuyuki:
Oh, I think I know the cause - winit used to spawn a thread for event loop as a hack, but that's gone in 0.20. So it's now messing up then.

RoyJacobs:
Ah okay, I'll spawn my winit event loop in a separate thread then. Seems to work, thanks for the help!

PR #331, opened 2019-09-27
Uses ringbuffer in feedback example

mitchhentges:
As mentioned on the Rust Audio discord, the `feedback` example should use a ringbuffer instead of mpsc channels

### DIFF
--- a/examples/feedback.rs
+++ b/examples/feedback.rs
@@ -8,8 +8,10 @@
 
 extern crate cpal;
 extern crate failure;
+extern crate ringbuf;
 
 use cpal::traits::{DeviceTrait, EventLoopTrait, HostTrait};
+use ringbuf::RingBuffer;
 
 const LATENCY_MS: f32 = 150.0;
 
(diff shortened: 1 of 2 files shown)


examples/feedback.rs
```diff
@@ -37,12 +39,13 @@ fn main() -> Result<(), failure::Error> {
...
-    // The channel to share samples.
-    let (tx, rx) = std::sync::mpsc::sync_channel(latency_samples * 2);
+    // The buffer to share samples
+    let ring = RingBuffer::<f32>::new(latency_samples * 2);
+    let (mut producer, mut consumer) = ring.split();
 
     // Fill the samples with 0.0 equal to the length of the delay.
     for _ in 0..latency_samples {
-        tx.send(0.0)?;
+        producer.push(0.0).unwrap();
```
mitchhentges, line 50: I'd rather not unwrap this, but I'm not sure of a nice `failure`-happy way of converting `PushError`.
I tried:
```
producer.push(0.0).map_err(|push_error| {
    // Instantiate another error type here
})?;
```
But I wasn't sure what kind of `failure`-compatible error would be correct to create there.
Let me know what you think!

wrl: I'm not angry about `unwrap()` here. It's example code, and I think just adding a comment noting "we just created the ringbuffer with double the capacity we're pushing here, so if this fails, we have bigger problems" (or something to that effect) is sufficient.

Issue #318, opened 2019-08-10
Appveyor is broken


est31:
It seems that since the ownership transfer, appveyor has been broken.

Since Aug 10, these errors are being returned by the webhook:

```
{"message":"PR and project repositories do not match."}
```

See also https://github.com/appveyor/ci/issues/449

Not sure how to fix this.

est31:
We could maybe migrate to GitHub actions to mitigate this. It's enabled for the org, lewton uses it too. @luni-4 @ishitatsuyuki what do you think?

ishitatsuyuki:
Yes please, I don't have time to work on that but I was also planning to move to GitHub Actions.

Luni-4:
Yep, I agree. I can work on that, feel free to assign this task to me :)

est31:
Thanks @Luni-4 for working on this!

As for the config files, I guess appveyor can be removed as it is broken anyway. I'd say the other two can be removed as well as long as their functionality is replaced by the github actions config file. If you want to replace all in one go it's okay, just as replacing appveyor only and then tackling the other CIs later on.

ishitatsuyuki:
Thanks for working on this. Some things you’d want to keep a sight on:
- the circleci build files cover web platform, which might need some effort to set up.
- the Travis config contains an auto-publishing step for tags; you could keep that intact. If you are going to migrate it, ask @mitchmindtree to set the token.

Luni-4:
> As for the config files, I guess appveyor can be removed as it is broken anyway. I'd say the other two can be removed as well as long as their functionality is replaced by the github actions config file. If you want to replace all in one go it's okay, just as replacing appveyor only and then tackling the other CIs later on.

I think I will remove them in one go then, thanks!

> the circleci build files cover web platform, which might need some effort to set up.

Ok, I will maintain that, thanks for the advice!

> the Travis config contains an auto-publishing step for tags; you could keep that intact. If you are going to migrate it, ask @mitchmindtree to set the token.

IIRC it already exists an action for this step, but I need to check out. I will ask for sure if I need some help, thanks :)

PR #333, opened 2019-10-01
Added winapi feature "winbase" in windows dependencies

hatchling13:
When using cpal as dependency in other project, it was found that cpal
won't build due to unresolved import of winapi::um::winbase. Added the
feature at line 24 in Cargo.toml.

Edit) My OS was Windows 10, and Rust toolchain was like below:
stable-x86_64-pc-windows-msvc (default)
rustc 1.38.0 (625451e37 2019-09-23)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,7 +21,7 @@ hound = "3.4"
 ringbuf = "0.1.6"
 
 [target.'cfg(target_os = "windows")'.dependencies]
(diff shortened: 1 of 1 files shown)


MOZGIII:
Seems like appveyor integration is broken, it should've caught it.

est31:
@MOZGIII indeed it is, see #318

PR #336, opened 2019-10-04
Fix the clippy-action

Luni-4:
This PR fixes the `clippy-action` error

Thanks in advance for your review! :)


Luni-4:
@est31 
It seems there is an issue with PRs, the `clippy` error is gone though. I could disable it for PRs for now.

est31:
@Luni-4 yeah disabling would be cool.

Luni-4:
> @Luni-4 yeah disabling would be cool.

@est31 Done :)

PR #339, opened 2019-10-12
Implement WASAPI loopback support

### LINKED ISSUE #251: Support WASAPI Loopback
The WASAPI supports loopback, where you are provided a listening device for each output device.  There are a handful of libraries which support this, might be nice to have here?

I'm poking around at the source to see if I can figure it out, but I am quite inexperienced with Rust (and winapi, for that matter).  `winapi` provides a flag for this (used in other libraries I have found):  `AUDCLNT_STREAMFLAGS_LOOPBACK`.  Simply adding this doesn't work, at least for device enumeration.

Going to keep digging, but also curious if this is something that has been discussed/considered before.  Thanks!

Edit:  Oof, randomly found this while perusing some examples:  ```// note that AUDCLNT_STREAMFLAGS_LOOPBACK and AUDCLNT_STREAMFLAGS_EVENTCALLBACK
    // do not work together...``` so this may not be doable, since this lib is using that flag.

lucy:
This works by detecting output devices in build_input_stream() and
setting the AUDCLNT_STREAMFLAGS_LOOPBACK flag to enable loopback
recording.

closes #251

### DIFF
--- a/src/host/wasapi/device.rs
+++ b/src/host/wasapi/device.rs
@@ -552,7 +552,7 @@ impl Device {
         }
     }
 
-    fn data_flow(&self) -> EDataFlow {
+    pub(crate) fn data_flow(&self) -> EDataFlow {
         let endpoint = Endpoint::from(self.device as *const _);
         endpoint.data_flow()
     }
--- a/src/host/wasapi/mod.rs
+++ b/src/host/wasapi/mod.rs
@@ -21,6 +21,10 @@ mod device;
 mod stream;
 
 /// The WASAPI host, the default windows host type.
+///
+/// Note: If you use a WASAPI output device as an input device it will
+/// transparently enable loopback mode (see
+/// https://docs.microsoft.com/en-us/windows/win32/coreaudio/loopback-recording).
 #[derive(Debug)]
 pub struct Host;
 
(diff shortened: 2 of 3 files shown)


lucy:
Tested with:
```diff
diff --git a/examples/record_wav.rs b/examples/record_wav.rs
index 9127123..01f7c2d 100644
--- a/examples/record_wav.rs
+++ b/examples/record_wav.rs
@@ -13,9 +13,9 @@ fn main() -> Result<(), failure::Error> {
     let host = cpal::default_host();

     // Setup the default input device and stream with the default input format.
-    let device = host.default_input_device().expect("Failed to get default input device");
+    let device = host.default_output_device().expect("Failed to get default input device");
     println!("Default input device: {}", device.name()?);
-    let format = device.default_input_format().expect("Failed to get default input format");
+    let format = device.default_output_format().expect("Failed to get default input format");
     println!("Default input format: {:?}", format);
     let event_loop = host.event_loop();
     let stream_id = event_loop.build_input_stream(&device, &format)?;

```

ishitatsuyuki:
I think this is okay, but the current API design makes it a very hidden feature. Would you like to add some docs?

Though I think we might change the Device API to differ between input and output devices. At that time we might need to add a Windows-specific function to convert an output to "loopback" input.

lucy:
What's a good place to document this?

ishitatsuyuki:
I think you can document it as a part of `WasapiHost`.

memchk:
Is there a way I can help push this forward? I have a project [here](https://github.com/memchk/LedDesk), that would really benefit from this being merged. I currently have a vendored version of cpal with hacked together loopback support, it would be awesome if I could use a upstream version.

est31:
Merging as per @ishitatsuyuki 's approval, there seem to be docs now.

memchk:
Awesome! Thank y'all for the continued work on cpal!

open-trade:
Is there any document or example about how to use loopback in RustAudio/cpal?

SirJson:
It used to be as simple as just creating an input stream and sticking that into your output stream. The cpal API changed tho and I didn't had the time to wrap my head around it so I can't provide example code. But I guess the general idea didn't change here, just the function names.

PR #340, opened 2019-10-13
Switch to dtolnay's `thiserror` for handling error boilerplate

mitchmindtree:
Currently CPAL only really uses `failure` for its `derive` capabilities
and the ability to easily generate implementations for `Display`. That
said there are a few issues with using the `failure` crate:

- `failure` does not provie a `std::error::Error` implementation without
  first converting error types into `failure::Error`.
- It leaks significantly into the public API and expects downstream
  users to also depend on `failure` and the non-std `Fail` trait for
  their own error handling.
- Solved problems such as downcasting of causal errors which have since
  been addressed in `std`.
- Provides application-friendly `Fail` trait and `failure::Error` type,
  not particularly useful to libraries like CPAL.

The [`thiserror` crate](https://github.com/dtolnay/thiserror) is better
targeted towards libraries, does not leak into the public API while
providing easy generation of `Display`, `From` and `std::error::Error`
implementations including proper handling of the newish
`std::error::Error::source` method. Also slightly reduces the number
of dependencies.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,7 @@ lazy_static = "1.3"
 num-traits = "0.2.6"
 
 [dev-dependencies]
+anyhow = "1.0.12"
 hound = "3.4"
 ringbuf = "0.1.6"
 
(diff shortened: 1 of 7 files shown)


est31:
I think the examples still need porting.

mitchmindtree:
@est31 yep just noticed, I'm on it!

PR #345, opened 2019-10-26
Allow cargo-publish failure

Luni-4:
Allow `cargo-publish` to fail for now.

Looking for better solutions...


Luni-4:
@est31 

Now `cargo publish` shouldn't fail if the crate is already uploaded

PR #348, opened 2019-10-31
Don't fail if CoInitializeEx returns RPC_E_CHANGED_MODE. It's OK.

hrydgard:
As far as I can tell, cpal relies on no way on the COM mode being COINIT_MULTITHREADED.

So let's survive even if COM was already initialized in apartment-threaded mode (there is no such thing as single-threaded mode).

### DIFF
--- a/src/host/wasapi/com.rs
+++ b/src/host/wasapi/com.rs
@@ -1,30 +1,40 @@
 //! Handles COM initialization and cleanup.
 
-use super::check_result;
 use std::ptr;
 
 use super::winapi::um::objbase::{COINIT_MULTITHREADED};
 use super::winapi::um::combaseapi::{CoInitializeEx, CoUninitialize};
+use super::winapi::shared::winerror::{SUCCEEDED, HRESULT, RPC_E_CHANGED_MODE};
 
 thread_local!(static COM_INITIALIZED: ComInitialized = {
     unsafe {
-        // this call can fail if another library initialized COM in single-threaded mode
-        // handling this situation properly would make the API more annoying, so we just don't care
-        check_result(CoInitializeEx(ptr::null_mut(), COINIT_MULTITHREADED)).unwrap();
-        ComInitialized(ptr::null_mut())
+        // this call can fail with RPC_E_CHANGED_MODE if another library initialized COM
+        // in apartment-threaded mode. That's OK though, we don't care.
+        let result = CoInitializeEx(ptr::null_mut(), COINIT_MULTITHREADED);
+        if SUCCEEDED(result) || result == RPC_E_CHANGED_MODE {
+            ComInitialized {
+                result
+            }
+        } else {
+            // COM initialization failed in another way, something is really wrong.
+            panic!("Failed to initialize COM.");
+        }
     }
 });
 
 /// RAII object that guards the fact that COM is initialized.
-///
-// We store a raw pointer because it's the only way at the moment to remove `Send`/`Sync` from the
-// object.
-struct ComInitialized(*mut ());
+struct ComInitialized {
+    result: HRESULT,
+}
 
 impl Drop for ComInitialized {
     #[inline]
     fn drop(&mut self) {
-        unsafe { CoUninitialize() };
+        // Need to avoid calling CoUninitialize() if CoInitializeEx failed since it may have returned
+        // RPC_E_MODE_CHANGED - which is OK, see above.
+        if SUCCEEDED(self.result) {
+            unsafe { CoUninitialize() };
+        }
     }
 }
 


ishitatsuyuki:
cpal's thread safety assumptions might rely on MTA. We need an audit to back up the claim that threading mode doesn't matter.

hrydgard:
Empirically it works fine at least, and I don't see any complex COM/window message use that would be affected by the COM threading mode.

What prompted this was that the upcoming winit 0.20 only functions in apartment-threaded mode (see https://github.com/rust-windowing/winit/issues/1255), so this requirement makes it impossible to use cpal together with winit in the future, unless we can convince winit to lift that restriction.

ishitatsuyuki:
> makes it impossible to use cpal

It isn’t completely impossible, you could create a thread that perform all cpal API calls and that will not cause the conflict.

I wouldn't call "I ran it and it works" an audit, because the concern is about the Send and Sync implementation on most of the types, which is not always utilized but might become a soundness hole.

hrydgard:
Sent a pull request to winit to lift their restriction instead - if you don't care about OLE drag/drop there's no reason it shouldn't work with MULTITHREADED.

Creating an extra thread just for this reason is a last resort - not that threads are expensive, but having lots of mystery threads around isn't good for debuggability and profiling.

Still, at a first glance nothing in #301 seems to preclude apartment threaded use. The use of a sprinkling of RAII "com::com_initialized();" instead of carefully placed calls to CoInitializeEx/CoUninitialize makes a code audit difficult though.

hrydgard:
I still feel that this at least doesn't make anything worse. If it indeed somehow doesn't work with MULTITHREADED, then, well, without this pull request, it also didn't work, just crashed earlier.

Osspial:
Microsoft did a useful writeup on the difference between the two threading models, which you can read [here](https://blogs.msdn.microsoft.com/larryosterman/2004/04/28/what-are-these-threading-models-and-why-do-i-care/). TL;DR:

* `COINIT_MULTITHREADED` always runs COM methods in the thread in which they're invoked.
* `COINIT_APARTMENTTHREADED` sends a Windows message to the thread that created the COM object, and runs the method in that thread. Objects get destroyed when their host thread aborts. This requires that a Windows message loop be active in that thread, which Winit already manages.

The event-loop-must-be-running caveat shouldn't be an issue, since if another library already initialized COM to apartment-threaded that library should be running its own event loop anyway. I could potentially see there being problems if cpal sends COM objects across threads and the parent thread gets destroyed, but I'm not familiar with cpal's internal design so I don't know if it would trigger that case.

ishitatsuyuki:
I pretty much doubt that marshalling is feasible for cpal: audio loop can run much faster than the GUI's, and if methods are dispatched asynchronously then that mean this use case would just break horribly. An example of a method called inside critical section is https://docs.microsoft.com/en-us/windows/win32/api/audioclient/nf-audioclient-iaudiocaptureclient-getbuffer.

TBH I really couldn't figure out how the synchronization are supposed to be implemented, other than just using MTA so that we don't need to have to think about it.

Osspial:
Those are good points. In the meanwhile before #301 is finished, would it work to create a second thread, initialize that to `COINIT_MULTITHREADED`, use that thread to create COM objects, then send them back to the main thread?

Ralith:
The event loop is now dead. Is it time to revisit this?

ishitatsuyuki:
@Ralith nah, the story is mostly the same.

Rua:
Has this been addressed at all yet? I'm running into this issue and I'm not sure what the proper workaround is.

MCOfficer:
As a dumb rodio user, i'd mostly like to know how i can avoid this. I mean, i don't mind audio not working, i just don't want it to crash the entire application. I tried to wrap everything in `panic::catch_unwind`, and even that didn't work. Did anyone find a solution?

hrydgard:
Since apparently the cpal project isn't interested in fixing this, I just forked the whole thing. There's no simple solution, except possibly spawning an unnecessary extra thread and launching cpal from that.

Ralith:
cpal now spawns a thread internally; I'm not an expert on winapi or cpal's use thereof, but it might be worth checking whether this is still an issue on master, and if so, whether it might be easier to fix nondisruptively now.

ishitatsuyuki:
I'm sorry that this might be annoying, but after spending a while looking at Chromium source, it seems that the idea that Windows Audio APIs should be called on MTA threads is correct.

- [All audio engine initialization happens on a dedicated thread](https://source.chromium.org/chromium/chromium/src/+/master:media/audio/win/audio_manager_win.h;l=71;drc=1e7d9ec83d5f53454cec0f9abf5d09114322450b?originalUrl=https:%2F%2Fcs.chromium.org%2F)
- [Internal audio headers explicit states that it should be called on MTA thread or Chrome's Audio Thread](https://source.chromium.org/chromium/chromium/src/+/master:media/audio/win/core_audio_util_win.h;l=8;drc=c88fe6b8ac0d4fc5ff03863235acd37eab7937ae?originalUrl=https:%2F%2Fcs.chromium.org%2F)

Ralith:
Having to spin up a dedicated thread to make CPAL API calls if you're on windows is a massive footgun. What do other cross-platform audio libraries do? I haven't found any such limitation mentioned in portaudio's docs or discussions, for example.

> All audio engine initialization happens on a dedicated thread

The comment on that line suggests that this is only done to avoid blocking the UI:

```c
  // Helper methods for performing expensive initialization tasks on the audio
  // thread instead of on the UI thread which AudioManager is constructed on.
```

> Internal audio headers explicit states that it should be called on MTA thread or Chrome's Audio Thread

I'm not much of a winapi expert, but the comment seems to only say "all methods must be called on a valid COM thread," not what threading mode is in use.

ishitatsuyuki:
Fair, I think I had some oversights when looking at the comments.

Regarding the latter comment, [this one](https://source.chromium.org/chromium/chromium/src/+/master:media/audio/win/device_enumeration_win.h;l=21;drc=8fca5bbd18caffd5d30753dc4326cf264254728a?originalUrl=https:%2F%2Fcs.chromium.org%2F) explicitly states MTA. As you said in other places it's stated as just COM initialization required, but FWIW all such initialization on audio threads are MTA (which can be confirmed through a code search).

JoshuaBatty:
Hey I managed to make a super small example based off the current master of cpal that replicates this behavior in case it's helpful for someone looking to try and solve this. The test example can be [found here](https://github.com/JoshuaBatty/cpal/blob/winit_conflict/examples/beep.rs)

The example kicks off a winit events loop, and when the key R is pressed, it runs the regular cpal beep example. If the code to launch cpal in a thread is uncommented then winit and cpal live happily side by side with each other but as said above, this is not a very nice solution.

Herschel:
My previous workaround for this was to create the stream on another thread and move it back to the main thread, but this no longer works as 0.12 marks `Stream` as `!Send` on all platforms; what's the best solution now?

I do not interact with the `Stream` once it starts playing, so I suppose I could just `std::mem::forget` instead of returning it to keep it alive, or toss an `unsafe impl Send` on the structure when I return it.

Imberflur:
@Herschel it looks like this can be avoided now if you don't need drag and drop support https://github.com/rust-windowing/winit/pull/1524

Rua:
What is the solution when you do need drag and drop support?

Imberflur:
@Rua my previous workaround (using rodio) was to spawn a new thread for every operation that would eventually hit this and then immediately `.join()` to block until the operation was complete. This seemed okayish because most of it was just initialization and initial creation of `Sink`s iirc.

dheijl:
As it looks like this PR isn't going to be accepted anytime soon: an easy workaround is to make sure that you initialize cpal before you do anything else, so that cpal is the first library to initialize COM, this way you can prevent this unnecessary panic.

PR #350, opened 2019-11-09
wasapi: emphasize eventloop abstraction vs stream handling

msiglreith:
Shifting code around to better mark the difference between the typesafe eventloop abstraction provided by cpal and the wasapi specific stream handling.
This should give more room for potentially lifting the current eventloop into a separate layer and/or exposing backend internals.

Changes:
- Rename `StreamInner` into `Stream` and split it into `StreamInner` (wasapi specific) and `StreamMeta` (event loop relevant) components
- Unify the internal event loop to streamline the API of both capture and render clients. Now they are pretty compact and concise IMO

### DIFF
--- a/src/host/wasapi/stream.rs
+++ b/src/host/wasapi/stream.rs
@@ -55,7 +55,7 @@ pub struct EventLoop {
 
 struct RunContext {
     // Streams that have been created in this event loop.
-    streams: Vec<StreamInner>,
+    streams: Vec<Stream>,
 
     // Handles corresponding to the `event` field of each element of `voices`. Must always be in
     // sync with `voices`, except that the first element is always `pending_scheduled_event`.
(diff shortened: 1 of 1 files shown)


mitchmindtree:
Thanks @msiglreith! Is it safe to close this now that #354 has landed, or does this PR include some other changes you would like to see land? Feel free to rebase if so and I'll take a closer look!

msiglreith:
My longer term goals were to slowly transition towards a low level api with exposed RunContext and feature levels (Vulkan level) (first iteration: https://github.com/norse-rs/audir/blob/master/examples/music/src/main.rs#L94-L123) and more easy to approach high level as it's currently with basic stream management on top.

As with the removal of the eventloop API I have to reassess a bit as right now this PR doesn't provide a lot of value by itself :D

-> closing for now

PR #351, opened 2019-11-15
Move error files into another directory

Luni-4:
In this way, `cargo-publish` shouldn't view them

### DIFF
--- a/.github/workflows/cpal.yml
+++ b/.github/workflows/cpal.yml
@@ -30,8 +30,6 @@ jobs:
 
     env:
       CRATESIO_TOKEN: ${{ secrets.CRATESIO_TOKEN }}
-      ALSA: alsa-sys.txt
-      CPAL: cpal.txt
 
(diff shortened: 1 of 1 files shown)


Luni-4:
> Could you use `mktemp` for this instead of the home directory?

Hmm, I'm not sure about that. Why do you think that `cargo-publish` couldn't find them?

est31:
mktemp is the utility for creating temporary files. They should usually be inside the /tmp dir and have random names. It's a better solution than to hack it via using the home dir.

Luni-4:
> mktemp is the utility for creating temporary files. They should usually be inside the /tmp dir and have random names. It's a better solution than to hack it via using the home dir.



> mktemp is the utility for creating temporary files. They should usually be inside the /tmp dir and have random names. It's a better solution than to hack it via using the home dir.

Ok then, now it's much clearer, you want the files to be in the `/tmp` dir. Ok, let me try :)

PR #354, opened 2019-12-13
Removing the `EventLoop` - rebased

mitchmindtree:
This PR is a rebase of @ishitatsuyuki's work in #301. Originally opened at ishitatsuyuki/cpal#3 but closed in favour opening a new PR here.

### Backends

- [x] null
- [x] ALSA
- [x] CoreAudio
- [x] WASAPI
- [x] Emscripten
- [x] ASIO

### Other TODO

- [x] Polymorphic struct versions of Traits will be explicitly !Sync, !Send
- [x] Docs

### ALSA multiple streams bug

*Edit: This bug seems to have been addressed in the rebase*.

After testing each of the examples with the ALSA backend on my Arch Linux system, I noticed that the `feedback.rs` example no longer works. No audio output was occurring and after some investigation it seems the output stream callback only seems to get called once. `beep.rs` and `record_wav.rs` both work fine, which leads me to wonder if there's an issue with the way multiple streams are being handled. `feedback.rs` also works fine on `master`, so this has likely been introduced in this PR.

@ishitatsuyuki does anything come to mind as a potential cause for this? I did notice that we're now spawning a whole OS thread per stream. I wonder if there's some kind of race going on in the stream loop? I might see if I can find anything about thread safety in the ALSA docs. I also wonder if we can reduce the number of threads necessary by just using a single thread on which all of the streams run, similar to the old `EventLoop::run` impl. If there is some race going on, maybe using a single thread for all streams would help to avoid this.

Also, I was reading through PortAudio's non-blocking ALSA code and noticed that, when spinning up a thread for a new stream, they also attempt to set the thread's priority to the highest possible. It might be worth seeing if there is some way we can do this too, though this work can probably wait for a follow-up PR.

### DIFF
--- a/examples/feedback.rs
+++ b/examples/feedback.rs
@@ -49,15 +49,7 @@ fn main() -> Result<(), anyhow::Error> {
 
     // Build streams.
     println!("Attempting to build both streams with `{:?}`.", format);
-    let input_stream = input_device.build_input_stream(&format, move |result| {
-        let data = match result {
-            Ok(data) => data,
-            Err(err) => {
-                eprintln!("an error occurred on input stream: {}", err);
-                return;
-            },
-        };
-
+    let input_stream = input_device.build_input_stream(&format, move |data| {
         match data {
             cpal::StreamData::Input {
                 buffer: cpal::UnknownTypeInputBuffer::F32(buffer),
@@ -74,15 +66,10 @@ fn main() -> Result<(), anyhow::Error> {
             },
             _ => panic!("Expected input with f32 data"),
         }
+    }, move |err| {
+        eprintln!("an error occurred on input stream: {}", err);
     })?;
-    let output_stream = output_device.build_output_stream(&format, move |result| {
-        let data = match result {
-            Ok(data) => data,
-            Err(err) => {
-                eprintln!("an error occurred on output stream: {}", err);
-                return;
-            },
-        };
+    let output_stream = output_device.build_output_stream(&format, move |data| {
         match data {
             cpal::StreamData::Output {
                 buffer: cpal::UnknownTypeOutputBuffer::F32(mut buffer),
@@ -103,6 +90,8 @@ fn main() -> Result<(), anyhow::Error> {
             },
             _ => panic!("Expected output with f32 data"),
         }
+    }, move |err| {
+        eprintln!("an error occurred on output stream: {}", err);
     })?;
     println!("Successfully built streams.");
 
--- a/examples/feedback.rs
+++ b/examples/feedback.rs
@@ -10,30 +10,27 @@ extern crate anyhow;
 extern crate cpal;
 extern crate ringbuf;
 
-use cpal::traits::{DeviceTrait, EventLoopTrait, HostTrait};
+use cpal::traits::{DeviceTrait, HostTrait, StreamTrait};
 use ringbuf::RingBuffer;
 
 const LATENCY_MS: f32 = 150.0;
 
 fn main() -> Result<(), anyhow::Error> {
     let host = cpal::default_host();
-    let event_loop = host.event_loop();
 
     // Default devices.
-    let input_device = host.default_input_device().expect("failed to get default input device");
-    let output_device = host.default_output_device().expect("failed to get default output device");
+    let input_device = host
+        .default_input_device()
+        .expect("failed to get default input device");
+    let output_device = host
+        .default_output_device()
+        .expect("failed to get default output device");
     println!("Using default input device: \"{}\"", input_device.name()?);
     println!("Using default output device: \"{}\"", output_device.name()?);
 
     // We'll try and use the same format between streams to keep it simple
     let mut format = input_device.default_input_format()?;
     format.data_type = cpal::SampleFormat::F32;
-
-    // Build streams.
-    println!("Attempting to build both streams with `{:?}`.", format);
-    let input_stream_id = event_loop.build_input_stream(&input_device, &format)?;
-    let output_stream_id = event_loop.build_output_stream(&output_device, &format)?;
-    println!("Successfully built streams.");
 
     // Create a delay in case the input and output devices aren't synced.
     let latency_frames = (LATENCY_MS / 1_000.0) * format.sample_rate.0 as f32;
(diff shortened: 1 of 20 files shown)


ishitatsuyuki:
some quick facts:

- libalsa is thread safe with a flag, which is enabled by default on most distros. In my implementation I assume it is enabled.
- there are two ways to configure real-time privileges:
  * with ulimits. This is simpler and apparently more popular.
  * with RealtimeKit. It provides additional monitoring to ensure a process doesn't over-use realtime budgets, but requires D-Bus, and because it's not very popular it has been almost a sibling of PulseAudio.

ishitatsuyuki:
Oh hey, and I tested the examples and apparently it works. I'm running PulseAudio with Arch if that matters.

mitchmindtree:
Oh strange, it's working for me too now! Even switching through all my different available audio devices seems to work.

Ahh, I can recreate the error by switching back to the non-rebased branch. It looks like one of the ALSA patches introduced in the rebase must have addressed whatever was going on here.

ishitatsuyuki:
Feels like #327.

MOZGIII:
> Polymorphic struct versions of Traits will be explicitly !Sync, !Send
What does this mean (compared to the currently existing event loop APIs)? And why not make them Send + Sync?

ishitatsuyuki:
>> Polymorphic struct versions of Traits will be explicitly !Sync, !Send ...

@MOZGIII Because it’s not portable. Android is a clear obstacle in this case, and there are also some platforms where thread safety is unknown.

If you need the Send/Sync version, then you need to use the concrete types.

MOZGIII:
Oh, ok. I'll still be able to use generics, right?

Are the structs this is about just to abstract away things to dynamic dispatch while hiding it as an implementation detail?

MOZGIII:
Actually, it'd be perfect if some lower-level tweaking abilities were exposed if I were to use concrete classes - like buffer sizes etc. Any plans on enabling that?

ishitatsuyuki:
I was thinking of exposing those platform specific options, but I'm not in charge of development at the moment. It's completely possible and a matter of being implemented or not.

MOZGIII:
Sounds good. I'd look into exposing the things I need, but there's way too much turbulence going on with the code base, so I'll just wait for this to land first.

mitchmindtree:
@msiglreith I just had a read through your ASIO update and it looks good to me!

From my understanding reading through the thread at ishitatsuyuki/cpal#2 it looks like the remaining steps are:

- **Ensure `Stream`s don't become invalid if the `Device` is dropped first.** It looks like we can solve this by sharing the `Device`'s `driver` and `asio_streams` `Arc`s with the `Stream`s themselves.
- **Ensure that when a `Stream` is dropped that its callback is also removed from the main ASIO loop.** We could possibly achieve this by having the `asio::Driver::set_callback` method return a unique `Id` associated with the callback which can be stored by the `Stream`. The `Id` can then be used by the `Stream` upon `drop` to remove its callback from the list, perhaps via some new `asio::Driver::remove_callback(callback_id)` method. It might be worth changing the `asio::Driver::set_callback` method name to `push_callback` or `add_callback` to make its behaviour clearer.
- **Work out some way to stop stream creation/removal from blocking the processing of audio**. As there's not an easy, clear solution for this just yet I'm happy to leave this for a future PR, particularly seeing as this is how it was behaving before anyway. Landing this PR is probably a bigger priority as it's a blocker for quite a few other issues.

Does this sound about right? Would you be interested in finishing this off? If not, let me know and I'll rebase and make these changes myself next time I get access to a Windows machine.

msiglreith:
Hi 👋,

1st and 2nd point are rather design decision imo. It was not clear to me how the lifetime of the objects are coupled in the API. The suggested approach looks viable.

For the 3rd one, there are two aspects:
- Recreation of the buffers in the asio driver layer when adding a new stream, which could be addressed on the cpal API side
- Adding/removing streams in the internal event loop, which is not really an issue in my opinion. Using lightweight mutexes should be enough

> Would you be interested in finishing this off? If not, let me know and I'll rebase and make these changes myself next time I get access to a Windows machine.

The earliest date for me to start working on it again would most likely be at end of the year.

mitchmindtree:
Thanks for the quick response!

> The earliest date for me to start working on it again would most likely be at end of the year.

No worries! I'll try to get onto this in the next day or two.

examples/feedback.rs
```diff
@@ -50,59 +47,67 @@ fn main() -> Result<(), anyhow::Error> {
...
+        "Starting the input and output streams with `{}` milliseconds of latency.",
+        LATENCY_MS
+    );
+    input_stream.play()?;
+    output_stream.play()?;
 
     // Run for 3 seconds before closing.
     println!("Playing for 3 seconds... ");
     std::thread::sleep(std::time::Duration::from_secs(3));
+    drop(input_stream);
```
msiglreith, line 109: why an explicit drop here?

src/host/wasapi/stream.rs
```diff
@@ -16,33 +10,24 @@ use super::winapi::um::winnt;
...
-    // `run()`. This ensures that two `run()` invocations can't run at the same time, and also
-    // means that we shouldn't try to lock this field from anywhere else but `run()`.
-    run_context: Mutex<RunContext>,
-
-    // Identifier of the next stream to create. Each new stream increases this counter. If the
-    // counter overflows, there's a panic.
-    // TODO: use AtomicU64 instead
-    next_stream_id: AtomicUsize,
+pub struct Stream {
+    /// The high-priority audio processing thread calling callbacks.
```
msiglreith, line 28: These comments feel kind of misleading as priority for these threads are never set

src/lib.rs
```diff
@@ -56,87 +52,96 @@
...
+//! that was passed to the function. The callback is passed an instance of type `StreamData` that
+//! represents the data that must be read from or written to. The inner `UnknownTypeOutputBuffer`
+//! can be one of `I16`, `U16` or `F32` depending on the format that was passed to
+//! `build_output_stream`.
 //!
-//! Now we must start the stream. This is done with the `play_stream()` method on the event loop.
+//! > **Note**: Creating and running a stream will *not* block the thread. On modern platforms, the
+//! > given callback is called by a dedicated, high-priority thread responsible for delivering
+//! > audio data to the system's audio device in a timely manner. On older platforms that only
+//! > provide a blocking API (e.g. ALSA), CPAL will create a thread in order to consistently
```
msiglreith, line 82: should be noted that it's currently one thread per stream

examples/feedback.rs
ishitatsuyuki: The stream terminates when it's dropped, and this makes it explicit.

mitchmindtree:
Thanks for the review @msiglreith!

I think this is just about ready to go now other than:

> Polymorphic struct versions of Traits will be explicitly !Sync, !Send

@ishitatsuyuki off the top of your head, do you remember if this is necessary for *all* of those types? Or is it e.g. just the `Host` that requires this? Or just the `Stream`? It would be nice if we can make this restriction only where absolutely necessary as it is a bit of an inconvenience to have on the API.

ishitatsuyuki:
My original intention was AAudio, which is not only not thread safe but also prohibits calling certain functions within a callback. Therefore I went for setting the `!Send` requirement. (Apparently I can't confirm the callback limits anymore, which used to be explicitly checked in the library code - maybe I'm just missing where it's implemented or Google might have patched this.)

AAudio on the other hand is stateless for stream creation, so the `!Send` and `!Sync` bound is only required for `Stream`.

Though, I think ASIO is not thread safe at all? Therefore we might also need `!Sync` (but not `!Send`) on `Host`.

src/host/wasapi/stream.rs
```diff
@@ -81,569 +70,49 @@ enum AudioClientFlow {
...
+            handles: vec![pending_scheduled_event, stream_inner.event],
+            stream: stream_inner,
+            commands: rx,
+        };
 
-                        debug_assert!(!buffer.is_null());
-                        let buffer_len = frames_available as usize
-                            * stream.bytes_per_frame as usize / sample_size;
+        let thread =
+            thread::spawn(move || run_inner(run_context, &mut data_callback, &mut error_callback));
```
TannerRogalsky, line 109: Should this thread initialize the COM library to COINIT_MULTITHREADED? And, if so, does this mean that the calling thread can have it's requirement on that threading model relaxed since the device setup is less timing-sensitive?

ishitatsuyuki: COM is initialised lazily using thread_local.

Threading model will not be relaxed AFAIK because what matters is the thread that created the handle. As long as the handle is created before the thread spawns, we'll need to use MTA when doing it.

mitchmindtree:
OK this should be good to go!

MOZGIII:
Any plans on publishing a new version to crates.io in the near future?

Ralith:
Thank you for your hard work in bringing this to completion!

PR #358, opened 2020-01-07
DragonFly support

jbeich:
DragonFly has the same audio system as FreeBSD.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,7 +25,7 @@ ringbuf = "0.1.6"
 winapi = { version = "0.3", features = ["audiosessiontypes", "audioclient", "coml2api", "combaseapi", "debug", "devpkey", "handleapi", "ksmedia", "mmdeviceapi", "objbase", "std", "synchapi", "winbase", "winuser"] }
 asio-sys = { version = "0.1", path = "asio-sys", optional = true }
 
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -25,7 +25,7 @@ ringbuf = "0.1.6"
 winapi = { version = "0.3", features = ["audiosessiontypes", "audioclient", "coml2api", "combaseapi", "debug", "devpkey", "handleapi", "ksmedia", "mmdeviceapi", "objbase", "std", "synchapi", "winbase", "winuser"] }
 asio-sys = { version = "0.1", path = "asio-sys", optional = true }
 
-[target.'cfg(any(target_os = "linux", target_os = "dragonfly", target_os = "freebsd", target_os = "openbsd"))'.dependencies]
+[target.'cfg(any(target_os = "linux", target_os = "dragonfly", target_os = "freebsd"))'.dependencies]
```
mitchmindtree, line 28: Thanks for the PR!

This all seems fine to me, though would you mind explaining why you removed the `target_os = "openbsd"` here? I've never used or tested cpal on openbsd - would this affect any of those users?

mitchmindtree: Ahh my apologies, I just noticed your reasoning in the relevant commit message!

PR #359, opened 2020-01-14
Remove `UnknownTypeBuffer` in favour of specifying sample type.

### LINKED ISSUE #119: Move `SampleFormat` check out of the main stream loop
Currently CPAL requires matching on an `UnknownTypeBuffer` enum upon every iteration of the audio loop, despite being able to query compatible sample formats prior to running the loop using `Voice::format`.

The `beep` example demonstrates how each format might be handled in this manner:

``` rust
    loop {
        match channel.append_data(32768) {
            cpal::UnknownTypeBuffer::U16(mut buffer) => {
                for (sample, value) in buffer.chunks_mut(format.channels.len()).zip(&mut data_source) {
                    let value = ((value * 0.5 + 0.5) * std::u16::MAX as f32) as u16;
                    for out in sample.iter_mut() { *out = value; }
                }
            },

            cpal::UnknownTypeBuffer::I16(mut buffer) => {
                for (sample, value) in buffer.chunks_mut(format.channels.len()).zip(&mut data_source) {
                    let value = (value * std::i16::MAX as f32) as i16;
                    for out in sample.iter_mut() { *out = value; }
                }
            },

            cpal::UnknownTypeBuffer::F32(mut buffer) => {
                for (sample, value) in buffer.chunks_mut(format.channels.len()).zip(&mut data_source) {
                    for out in sample.iter_mut() { *out = value; }
                }
            },
        }

        channel.play();
    }
```

In any serious audio applications, the work contained in each branch would likely be required to be abstracted into some function, where the application either:
- Does all its work in one format and then converts the final buffer to the stream's sample format or
- Remains entirely generic over the sample type, using some `Sample` trait.

In either case, we likely end up using some generic function within the branching like so:

``` rust
    loop {
        match channel.append_data(32768) {
            cpal::UnknownTypeBuffer::U16(mut buffer) =>
                fill_buffer_with_data::<u16>(&mut buffer, &mut data_source),
            cpal::UnknownTypeBuffer::I16(mut buffer) =>
                fill_buffer_with_data::<i16>(&mut buffer, &mut data_source),
            cpal::UnknownTypeBuffer::F32(mut buffer) =>
                fill_buffer_with_data::<f32>(&mut buffer, &mut data_source),
        }

        channel.play();
    }
```

Considering we already know whether or not certain formats are supported before we start the stream, we should be able to move this branching to occur before the stream even begins:

``` rust
fn main() {
    let endpoint = cpal::get_default_endpoint().expect("Failed to get default endpoint");

    if run_stream::<f32>(&end_point, &mut data_source).is_ok() {}
    else if run_stream::<i32>(&end_point, &mut data_source).is_ok() {}
    else if run_stream::<i16>(&end_point, &mut data_source).is_ok() {}
    ...
    else {
        panic!("No compatible audio stream formats found for the device");
    }
}
```

This could of course be refactored to only try different sample formats if some specific `UnsupportedFormat` error is returned, or by iterating on the `EndPoint`'s supported formats, matching on them and attempting to run the stream that way.

I'm curious to get your thoughts on this, as the current in-loop matching seems both unnecessary and to question whether the `sample_format` field in the `Format` struct yielded by `endpoint.get_supported_formats_list()` is useful at all?

### LINKED ISSUE #260: UnknownTypeBuffer does not feel rusty
I'm new to rust, but `UnknownTypeOutputBuffer`/`UnknownTypeInputBuffer` in the event loop API does not feel rusty at all.

I think it could be implemented differently, for example using traits with bound types.

After playing around a bit, I was able to pin point my discomfort to the following problem: enum types are used for encoding type information, which breaks a lot of flexibility that rust provides. As I've seen in the cpal implementation itself, macros are used to build the `UnknownTypeBuffer` result types, which tells me something's definitely wrong.

I'd like to discuss possible ways of improving this API.
Disclaimer: I'm still learning though, so my suggestions may be horribly bad.

So, why enum types, when we could use traits with bound types?

mitchmindtree:
This is an implementation of the planned changes described in #119.

For a quick overview of how the API has changed, check out the updated
examples.

**TODO:**

- [x] Update API.
- [x] Update examples.
- ~Remove `data_type` field from `Format`? See [here](https://github.com/RustAudio/cpal/issues/119#issuecomment-573788380).~
- [x] Update docs.
- Update backends:
  - [x] null
  - [x] ALSA
  - [x] ASIO
  - [x] WASAPI
  - [x] CoreAudio
  - [x] Emscripten

Closes #119
Closes #260

### DIFF
--- a/asio-sys/Cargo.toml
+++ b/asio-sys/Cargo.toml
@@ -10,7 +10,7 @@ keywords = ["audio", "sound", "asio", "steinberg"]
 build = "build.rs"
 
 [target.'cfg(any(target_os = "windows"))'.build-dependencies]
-bindgen = "0.42.0"
+bindgen = "0.51.0"
 walkdir = "2"
 cc = "1.0.25"
 
--- a/asio-sys/src/bindings/mod.rs
+++ b/asio-sys/src/bindings/mod.rs
@@ -319,9 +319,7 @@ impl Asio {
         // Make owned CString to send to load driver
         let driver_name_cstring = CString::new(driver_name)
             .expect("failed to create `CString` from driver name");
-        let mut driver_info = ai::ASIODriverInfo {
-            _bindgen_opaque_blob: [0u32; 43],
-        };
+        let mut driver_info = std::mem::MaybeUninit::<ai::ASIODriverInfo>::uninit();
 
         unsafe {
             // TODO: Check that a driver of the same name does not already exist?
@@ -329,7 +327,8 @@ impl Asio {
                 false => Err(LoadDriverError::LoadDriverFailed),
                 true => {
                     // Initialize ASIO.
-                    asio_result!(ai::ASIOInit(&mut driver_info))?;
+                    asio_result!(ai::ASIOInit(driver_info.as_mut_ptr()))?;
+                    let _driver_info = driver_info.assume_init();
                     let state = Mutex::new(DriverState::Initialized);
                     let name = driver_name.to_string();
                     let destroyed = false;
--- a/examples/beep.rs
+++ b/examples/beep.rs
@@ -1,56 +1,61 @@
 extern crate anyhow;
 extern crate cpal;
 
-use cpal::traits::{DeviceTrait, StreamTrait, HostTrait};
+use cpal::traits::{DeviceTrait, HostTrait, StreamTrait};
 
 fn main() -> Result<(), anyhow::Error> {
     let host = cpal::default_host();
-    let device = host.default_output_device().expect("failed to find a default output device");
+    let device = host
+        .default_output_device()
+        .expect("failed to find a default output device");
     let format = device.default_output_format()?;
     let sample_rate = format.sample_rate.0 as f32;
-    let channels = format.channels;
-    let mut sample_clock = 0f32;
+    let channels = format.channels as usize;
 
     // Produce a sinusoid of maximum amplitude.
+    let mut sample_clock = 0f32;
     let mut next_value = move || {
         sample_clock = (sample_clock + 1.0) % sample_rate;
         (sample_clock * 440.0 * 2.0 * 3.141592 / sample_rate).sin()
     };
 
-    let stream = device.build_output_stream(&format, move |data| {
-        match data {
-            cpal::StreamData::Output { buffer: cpal::UnknownTypeOutputBuffer::U16(mut buffer) } => {
-                for sample in buffer.chunks_mut(channels as usize) {
-                    let value = ((next_value() * 0.5 + 0.5) * std::u16::MAX as f32) as u16;
-                    for out in sample.iter_mut() {
-                        *out = value;
-                    }
-                }
-            },
-            cpal::StreamData::Output { buffer: cpal::UnknownTypeOutputBuffer::I16(mut buffer) } => {
-                for sample in buffer.chunks_mut(channels as usize) {
-                    let value = (next_value() * std::i16::MAX as f32) as i16;
-                    for out in sample.iter_mut() {
-                        *out = value;
-                    }
-                }
-            },
-            cpal::StreamData::Output { buffer: cpal::UnknownTypeOutputBuffer::F32(mut buffer) } => {
-                for sample in buffer.chunks_mut(channels as usize) {
-                    let value = next_value();
-                    for out in sample.iter_mut() {
-                        *out = value;
-                    }
-                }
-            },
-            _ => (),
-        }
-    }, move |err| {
+    let err_fn = |err| {
         eprintln!("an error occurred on stream: {}", err);
-    })?;
+    };
+
+    let stream = match format.data_type {
+        cpal::SampleFormat::F32 => device.build_output_stream(
+            &format,
+            move |mut data| write_data::<f32>(&mut *data, channels, &mut next_value),
+            err_fn,
+        ),
+        cpal::SampleFormat::I16 => device.build_output_stream(
+            &format,
+            move |mut data| write_data::<i16>(&mut *data, channels, &mut next_value),
+            err_fn,
+        ),
+        cpal::SampleFormat::U16 => device.build_output_stream(
+            &format,
+            move |mut data| write_data::<u16>(&mut *data, channels, &mut next_value),
+            err_fn,
+        ),
+    }?;
+
     stream.play()?;
 
     std::thread::sleep(std::time::Duration::from_millis(1000));
 
     Ok(())
+}
+
+fn write_data<T>(output: &mut [T], channels: usize, next_sample: &mut dyn FnMut() -> f32)
+where
+    T: cpal::Sample,
+{
+    for frame in output.chunks_mut(channels) {
+        let value: T = cpal::Sample::from::<f32>(&next_sample());
+        for sample in frame.iter_mut() {
+            *sample = value;
+        }
+    }
 }
--- a/examples/feedback.rs
+++ b/examples/feedback.rs
@@ -49,50 +49,44 @@ fn main() -> Result<(), anyhow::Error> {
 
     // Build streams.
     println!("Attempting to build both streams with `{:?}`.", format);
-    let input_stream = input_device.build_input_stream(&format, move |data| {
-        match data {
-            cpal::StreamData::Input {
-                buffer: cpal::UnknownTypeInputBuffer::F32(buffer),
-            } => {
-                let mut output_fell_behind = false;
-                for &sample in buffer.iter() {
-                    if producer.push(sample).is_err() {
-                        output_fell_behind = true;
-                    }
+    let input_stream = input_device.build_input_stream(
+        &format,
+        move |data| {
+            let mut output_fell_behind = false;
+            for &sample in data.iter() {
+                if producer.push(sample).is_err() {
+                    output_fell_behind = true;
                 }
-                if output_fell_behind {
-                    eprintln!("output stream fell behind: try increasing latency");
-                }
-            },
-            _ => panic!("Expected input with f32 data"),
-        }
-    }, move |err| {
-        eprintln!("an error occurred on input stream: {}", err);
-    })?;
-    let output_stream = output_device.build_output_stream(&format, move |data| {
-        match data {
-            cpal::StreamData::Output {
-                buffer: cpal::UnknownTypeOutputBuffer::F32(mut buffer),
-            } => {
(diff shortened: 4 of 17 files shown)


mitchmindtree:
OK this should be good to go!

I'll leave it up for another couple days before merging in case anyone has any feedback.

msiglreith:
probably a good point in the timeline to get CI rustfmt checks in 👀

mitchmindtree:
Thanks for the review @msiglreith!

I've been reflecting on this PR a little more the past couple hours. One issue to note about this approach is that it will likely require generating a fair bit more code, as the entire call to `build_input/output_stream` must be generated for each of the unique sample formats the user wishes to support.

One potential alternative could be to revert back to a dynamically dispatched approach, though rather than the unwieldy `Unknown{Input/Output}BufferType`s we instead pass `&Data`/`&mut Data` to the callback, where `Data` provides a method for safely casting to a slice of the expected sample type.

Building the output stream in the beep.rs example might look something like this instead:

```rust
let data_fn = move |data| match sample_format {
    cpal::SampleFormat::F32 => write_data::<f32>(data, channels, &mut next_value),
    cpal::SampleFormat::I16 => write_data::<i16>(data, channels, &mut next_value),
    cpal::SampleFormat::U16 => write_data::<u16>(data, channels, &mut next_value),
};
let stream = device.build_output_stream(&format, data_fn, err_fn)?;

// ...

fn write_data<T>(output: &mut cpal::Data, channels: usize, next_sample: &mut dyn FnMut() -> f32)
where
    T: cpal::Sample,
{
    let output = output.as_slice_mut::<T>().expect("unexpected sample type");
        let value: T = cpal::Sample::from(&next_sample());
            *sample = value;
        }
    }
}
```

Thoughts?

msiglreith:
I'm generally in favor of more (kinda) unopinionated APIs at the abstraction layer (down to just passing memory ranges around), tends to be easier to agree on 👍
On open point for me is how to potentially distinguish between de/interleaved formats? IIRC the ASIO backend manually does these deinterleaving operations right now.

mitchmindtree:
> On open point for me is how to potentially distinguish between de/interleaved formats?

Aye this has been crossing my mind too, I think coreaudio is also interleaving/de-interleaving behind the scenes atm. It would be nice to know whether or not all platforms providing non-interleaved data do so with all channels in contiguous memory (one after the other). I think I'll leave adding an API to distinguish between non/interleaved for a future PR and look into this some more then. I agree it's worth keeping in mind.

mitchmindtree:
@msiglreith (and any other reviewers), I'm leaning towards going down this route mitchmindtree/cpal#3.

Ralith:
I'm disappointed by the way this turned out. The new approach strikes me as being semantically equivalent to the original `UnknownBufferType`, but more error-prone/less idiomatic, since now instead of pattern-matching to get a slice that's guaranteed to be of the right type, you have to make sure you get your `SampleFormat` and `as_slice_mut` uses matched up by hand. The `SampleFormat` check is still stuck in the main loop, so I'm not sure what problem this has solved.

mitchmindtree:
@Ralith originally I had opted for passing the sample type as a type parameter via the `build_input/output_stream` methods as in the original issue I opened. You can see how the API looked at this point by checking out the commit before [this comment](https://github.com/RustAudio/cpal/pull/359#issuecomment-575919009). However, soon after I wrote the comment, I realised that this would result in a much larger amount of code generation in each of the host implementations as a result for little to no practical performance or safety gain beyond what landed in this PR. Please see [this comment](https://github.com/RustAudio/cpal/pull/359#issuecomment-575931461) and [this PR](https://github.com/mitchmindtree/cpal/pull/3) where I shared my thoughts on this.

With the changes in this PR, the user can still move the check outside of the main loop - here's an example of an alternative way the beep.rs example could be written that moves the `SampleFormat` match outside of the main audio loop:

```rust
// ...

let stream = match format.data_type {
    cpal::SampleFormat::F32 => device.build_output_stream(
        &format,
        move |data| write_data::<f32>(data, channels, &mut next_value),
        err_fn,
    ),
    cpal::SampleFormat::I16 => device.build_output_stream(
        &format,
        move |data| write_data::<i16>(data, channels, &mut next_value),
        err_fn,
    ),
    cpal::SampleFormat::U16 => device.build_output_stream(
        &format,
        move |data| write_data::<u16>(data, channels, &mut next_value),
        err_fn,
    ),
}?;

// ...

fn write_data<T>(output: &mut cpal::Data, channels: usize, next_sample: &mut dyn FnMut() -> f32)
where
    T: cpal::Sample,
{
    let output = output.as_slice_mut::<T>().expect("unexpected sample type");
        let value: T = cpal::Sample::from(&next_sample());
            *sample = value;
        }
    }
}
```

If you're concerned about the necessity for this line:
```rust
    let output = output.as_slice_mut::<T>().expect("unexpected sample type");
```
Perhaps we could add an unsafe, unchecked version of this method that omits the check and returns the slice directly rather than an `Option`?

I hope this helps to clarify my thinking a little. If not let me know, and perhaps share a rough code snippet of what you would have liked to have seen instead.

Also I'm trying to leave a couple of days or so for review in my bigger PRs atm as there are quite a few things I'd like to address while I have the chance, but perhaps this isn't long enough. Let me know if you think there should be more time left for review.

Ralith:
> However, soon after I wrote the comment, I realised that this would result in a much larger amount of code generation in each of the host implementations as a result for little to no practical performance or safety gain beyond what landed in this PR. Please see this comment and this PR where I shared my thoughts on this.

I reviewed those prior to commenting. I agree that causing large amounts of monomorphization is not something that should be done lightly. A statically-typed (i.e. generic-based) API could be provided which did not incur significant monomorphization by, for example, constructing user-facing generic functions as trivial wrappers around internal monomorphic functions; this is sometimes referred to as an "outlining". I'm willing to attempt a draft of this approach if there's interest.

> With the changes in this PR, the user can still move the check outside of the main loop

The apparent removal of the check is only cosmetic; each iteration still must internally perform a type-check via `expect`, and worse still, all static typechecking has been lost so e.g. a rarely-used format combination in more complex code could easily grow a lurking panic.

> Perhaps we could add an unsafe, unchecked version of this method that omits the check and returns the slice directly rather than an Option?

I'm not really concerned about performance, but rather robustness and how easy it is to use the API incorrectly. This change looks like a performance noop and a step backwards in those other regards.

> Also I'm trying to leave a couple of days or so for review in my bigger PRs atm as there are quite a few things I'd like to address while I have the chance, but perhaps this isn't long enough

I would have liked to weigh in before the merge, and two days would have been adequate time to do so had I been aware of the PR. I'm not subscribed to every change that's made in the repository, only individual issues that I care about. As a subscriber to #119, I was only notified of the PR's existence when it was merged; a comment there when the PR was opened would have been helpful.

mitchmindtree:
> A statically-typed (i.e. generic-based) API could be provided which did not incur significant monomorphization by, for example, constructing user-facing generic functions as trivial wrappers around internal monomorphic functions; this is sometimes referred to as an "outlining". I'm willing to attempt a draft of this approach if there's interest.

That would be greatly appreciated if you get the chance! I spent a while trying to work out how to allow for a user to submit a generic function that could be called by the Host implementation based on the sample format, though could not quite manage to find a nice solution that did not require a fancier type system. Perhaps I was getting too caught up on trying to keep the API that allows the user to pass a closure or function - are you thinking perhaps the user passes a type implementing some trait that allows for processing the data instead?

> As a subscriber to #119, I was only notified of the PR's existence when it was merged; a comment there when the PR was opened would have been helpful.

That's a great point - I'll make a note to ping the relevant issues as I open each PR in the future, sorry for the oversight!

Ralith:
> That would be greatly appreciated if you get the chance!

Will do. I appreciate that rich polymorphism is sometimes more easily said than done. I'll definitely be trying something closurey first, as an explicit trait implementation is certainly much less ergonomic.

MOZGIII:
Hey! I love this, had pretty much the same issue with not being notified, but I think it turned up good.
I think we should consider using `sample` crate and it's `Format` and `Sample` types - or at least provide good compatibility for using it.

MOZGIII:
The patch's a bit unreadable with all the formatting changes, we really need to enforce `cargo fmt` everywhere once and for all.

Ralith:
I noticed there's a .rustfmt.toml that thoroughly breaks my stable rustfmt, I'm guessing due to nightly options.

mitchmindtree:
> I think we should consider using sample crate and it's Format and Sample types - or at least provide good compatibility for using it.

@MOZGIII I tend to use the `sample` crate in all my downstream audio stuff, I'm glad to see it suggested as a useful tool for this sort of thing!

That said, CPAL feels like a special case where we want to have the minimal set of dependencies possible and only the absolutely necessary abstractions to provide a steady cross-platform experience, as I imagine it will serve as the foundation for a lot of different kinds of audio applications/projects.

As a side note - there has been some talk of stripping down the `sample` crate a bit further lately, splitting it into multiple crates and feature-gating the different features. Perhaps it might become a better candidate for CPAL after this process, though even then I'm not sure CPAL would gain much. Users should still be able to use the `sample` crate to work pretty seemlessly with the audio data slices provided by CPAL, as `sample::Sample` is quite a vast superset of `cpal::Sample`.

MOZGIII:
> That said, CPAL feels like a special case where we want to have the minimal set of dependencies possible and only the absolutely necessary abstractions to provide a steady cross-platform experience, as I imagine it will serve as the foundation for a lot of different kinds of audio applications/projects.

Oh, yeah, I remember this was discussed earlier. So, it's still the same.

> As a side note - there has been some talk of stripping down the `sample` crate a bit further lately, splitting it into multiple crates and feature-gating the different features. Perhaps it might become a better candidate for CPAL after this process, though even then I'm not sure CPAL would gain much. Users should still be able to use the `sample` crate to work pretty seemlessly with the audio data slices provided by CPAL, as `sample::Sample` is quite a vast superset of `cpal::Sample`.

I like this. I wasn't aware of this, but this sort of things is why I'm bringing up using `sample` crate again.
I'm also using it downstream. I didn't work on my project for quite a while now, I'll need to get up to date with the latest changes with cpal and sample crate.

MOZGIII:
The compelling reason to use `sample` crate for `cpal` is `sample::Format` rather than `sample::Sample` - but I'd say usability experiments have to be conducted before this can be confirmed as a good (or bad :smile:) idea.

PR #365, opened 2020-01-20
Remove old `.rustfmt.toml` config. Run default `cargo fmt` on repo. Add formatting check github action.

mitchmindtree:
Seeing as a few large refactors have landed recently, I thought I'd take this opportunity to do a `cargo fmt` run and standardise on the default rustfmt settings.

This also adds an action to check the formatting though I'm uncertain if it's worth it? I wonder if there's a github action out there that automatically adds a commit or opens a PR against the user's branch with formatting fixes in the case that the formatting check fails?


MOZGIII:
Finally! Looking forward to it, it'll make reviewing the changes so much easier!

MOZGIII:
> This also adds an action to check the formatting though I'm uncertain if it's worth it? I wonder if there's a github action out there that automatically adds a commit or opens a PR against the user's branch with formatting fixes in the case that the formatting check fails?

An action to check the formatting is very useful to communicate issues with the incoming code to the authors. It's a great thing, I hope we had it sooner.
To automatically create a fixup PR to misformatted branches we could use this: https://github.com/marketplace/actions/create-pull-request
This can implemented quite trivially using that action - we just format the worktree in an action step, and instruct a PR creation - it should handle all the lifecycle stuff - like updating the previously created PR if needed - automatically.

Ralith:
Certainly *some* sort of automation is necessary for formatting to not inevitably drift.

mitchmindtree:
OK great!

@MOZGIII I'm inclined to merge this as is for now and save the auto-PR steps for a future PR. Is there any chance you would be interested in submitting a patch to address this after this lands? It would be a great help! I'd likely use it as an example of how to do so in many of my other repos too :)

@Ralith are you happy for this to land now? Or would you prefer we resolve #366 first?

Oh also I just realised this only formats `cpal` itself and not the other crates in the repo. I'll add a commit that adds the `--all` flag now.

MOZGIII:
> @MOZGIII I'm inclined to merge this as is for now and save the auto-PR steps for a future PR. Is there any chance you would be interested in submitting a patch to address this after this lands? It would be a great help! I'd likely use it as an example of how to do so in many of my other repos too :)

Possibly, but no guarantees :)

Ralith:
@mitchmindtree go nuts, as far as I'm concerned the current form of #366 is disposable proof-of-concept code. Per the TODO list, I planned to replace everything that's currently there regardless.

PR #366, opened 2020-01-21
Polymorphic streams

### LINKED ISSUE #119: Move `SampleFormat` check out of the main stream loop

Ralith:
Fixes #119. ~Seeking feedback before proceeding.~ As discussed in #359, this provides a statically-typed buffer to the user without incurring additional monomorphization cost.

TODO:
- [x] Consider a richer buffer type than a slice: leaving this for future work
- [x] Bundle channels and sample rate into a single argument
- [x] Input streams
- ~Privatize `Data`~
- [x] Update all backends
- [x] Update examples and docs

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -59,10 +59,10 @@
 //! use cpal::traits::{DeviceTrait, HostTrait, StreamTrait};
 //! # let host = cpal::default_host();
 //! # let device = host.default_output_device().unwrap();
-//! # let format = device.default_output_format().unwrap();
-//! let stream = device.build_output_stream_raw(
-//!     &format,
-//!     move |data: &mut Data| {
+//! # let shape = device.default_output_format().unwrap().shape();
+//! let stream = device.build_output_stream(
+//!     &shape,
+//!     move |data: &mut [f32]| {
 //!         // react to stream events and read or write stream data here.
 //!     },
 //!     move |err| {
@@ -93,15 +93,14 @@
 //! # let device = host.default_output_device().unwrap();
 //! # let format = device.default_output_format().unwrap();
 //! let err_fn = |err| eprintln!("an error occurred on the output audio stream: {}", err);
-//! let data_fn = move |data: &mut Data| match data.sample_format() {
-//!     SampleFormat::F32 => write_silence::<f32>(data),
-//!     SampleFormat::I16 => write_silence::<i16>(data),
-//!     SampleFormat::U16 => write_silence::<u16>(data),
-//! };
-//! let stream = device.build_output_stream_raw(&format, data_fn, err_fn).unwrap();
+//! let shape = format.shape();
+//! let stream = match format.data_type {
+//!     SampleFormat::F32 => device.build_output_stream(&shape, write_silence::<f32>, err_fn),
+//!     SampleFormat::I16 => device.build_output_stream(&shape, write_silence::<i16>, err_fn),
+//!     SampleFormat::U16 => device.build_output_stream(&shape, write_silence::<u16>, err_fn),
+//! }.unwrap();
 //!
-//! fn write_silence<T: Sample>(data: &mut Data) {
-//!     let data = data.as_slice_mut::<T>().unwrap();
+//! fn write_silence<T: Sample>(data: &mut [T]) {
 //!     for sample in data.iter_mut() {
 //!         *sample = Sample::from(&0.0);
 //!     }
@@ -181,6 +180,38 @@ pub struct Format {
     pub data_type: SampleFormat,
 }
 
+impl Format {
+    /// Construct a format having a particular `shape` and `data_type`.
+    pub fn with_shape(shape: &Shape, data_type: SampleFormat) -> Self {
+        Self {
+            channels: shape.channels,
+            sample_rate: shape.sample_rate,
+            data_type,
+        }
+    }
+
+    /// Extract aspects of the format independent of the data type.
+    pub fn shape(&self) -> Shape {
+        self.clone().into()
+    }
+}
+
+/// The properties of an input or output audio stream excluding its data type.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Shape {
+    pub channels: ChannelCount,
+    pub sample_rate: SampleRate,
+}
+
+impl From<Format> for Shape {
+    fn from(x: Format) -> Self {
+        Self {
+            channels: x.channels,
+            sample_rate: x.sample_rate,
+        }
+    }
+}
+
 /// Describes a range of supported stream formats.
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct SupportedFormat {
@@ -193,9 +224,10 @@ pub struct SupportedFormat {
     pub data_type: SampleFormat,
 }
 
-/// Represents a buffer of audio data, delivered via a user's stream data callback function.
+/// A buffer of dynamically typed audio data, passed to raw stream callbacks.
 ///
-/// Input stream callbacks receive `&Data`, while output stream callbacks expect `&mut Data`.
+/// Raw input stream callbacks receive `&Data`, while raw output stream callbacks expect `&mut
+/// Data`.
 #[derive(Debug)]
 pub struct Data {
     data: *mut (),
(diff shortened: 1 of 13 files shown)


msiglreith:
Would it be fine to provide this on top of the current raw `Data` api interface instead of replacing it?

I believe there is a benefit in splitting the problem spaces of designing a user-friendly API and an abstraction layer in the hope of being able to stabilizing the abstraction layer faster by providing a lot of flexibilty (ie full format specification w/ data layout with sight on more exotic format and de/interleaved support). This would, in theory, allow to quicker iterate on the primarily user facing API as well as providing the option to go down the unsafe rabbit hole.

mitchmindtree:
Ahh my apologies! On closer inspection I realise you're not passing the sample type parameter down throughout the entire host implementation, but rather using the `Data` type internally for the deeper implementation details so that not nearly as much code generation is required. This seems like a good middle ground to me!

mitchmindtree:
@msiglreith I see where you're both coming from - on one hand desiring easier API iteration and flexibility both for cpal and downstream crates, and on the other taking advantage of rust to remove the chance for incorrectness where possible.

One approach that comes to mind is to have what is currently the `build_input/output_stream` methods on master become `build_input/output_stream_raw` methods (or some better name), and then add a new, thin `build_input/output_stream` method to the `Device` trait with a provided implementation that takes the sample type as a parameter and calls `build_input/output_stream_raw` internally? Perhaps the `raw` method can be more reliably expected to remain unchanged and serve as a basis for us to refine the type-safe method as we iterate on its design, work out how to handle different data layouts in a nice manner, etc? That said, the raw `Data` type will likely need to be revisited when handling different layouts too. Currently it expects a single contiguous slice of data, however from memory not all backends providing non-interleaved data guarantee that each channel is contiguous in memory.

Side note: I'll open a dedicated issue where we can discuss handling different data layouts.

mitchmindtree:
I've opened an issue here w.r.t. alternate data layouts #367.

Ralith:
> Would it be fine to provide this on top of the current raw Data api interface instead of replacing it?

This is how it works internally, and exposing that to the user seems reasonable to me; sometimes dynamic typing is more convenient.

> This approach of specifying the sample type in the build_input/output_stream method instead does not avoid the chance that a user might accidentally specify the wrong sample type for the matched sample format - the same issue exists where a panic! may lie hidden for a rarely used sample format in the case that the user has made a mistake

The `unwrap` introduced in this patch will never panic, because it is statically guaranteed that the type parameter we supply to `as_slice_mut` matches the format we constructed the stream with. We could even replace it with an unsafe cast if desired, but I don't expect the cost of an impossible branch traversed once per callback to be significant. If the user supplies an unsupported sample type, they will receive an `Err` from stream construction instead. This is easier to handle gracefully, because you don't need to route error information between threads through the callback and check it asynchronously. That's the core win of this change.

> On closer inspection I realise you're not passing the sample type parameter down throughout the entire host implementation

Right. This change does not introduce a type parameter to any formerly monomorphic functions.

> add a new, thin build_input/output_stream method to the Device trait with a provided implementation that takes the sample type as a parameter and calls build_input/output_stream_raw internally

I think this is reasonable, not necessarily for reasons of stability (since modifying any trait method is a breaking change) but because the polymorphic interface may not be convenient for all applications. As a bonus, it should simplify implementation of and iteration on the polymorphic interface, since we won't need to update every backend.

Thanks to both of you for having a look! I'll make a second draft soon.

Ralith:
I've fleshed out what I think is the minimal form of this change based on the above discussion. These interfaces will need to evolve substantially due to #367, but that's best handled separately.

src/lib.rs
```diff
@@ -181,6 +180,38 @@ pub struct Format {
...
+
+    /// Extract aspects of the format independent of the data type.
+    pub fn shape(&self) -> Shape {
+        self.clone().into()
+    }
+}
+
+/// The properties of an input or output audio stream excluding its data type.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct Shape {
```
Ralith, line 201: Bikeshedding welcome on this name, though I think this one suits the expected addition of interleaving information per #367.

Ralith:
The examples illustrate that the new interfaces makes some code less concise and other code more concise. This illustrates some of the benefit of retaining the dynamically typed APIs as well. However, in all cases it removes opportunity for error, so I think it's appropriate to prefer the polymorphic APIs in examples and documentation.

src/lib.rs
Ralith: Perhaps `Format` should become `(SampleFormat, Shape)`, or a struct with those members?

Ralith: On further consideration I like the previous proposal of

> remove the data_type field and have the default_input/output_format return a (Format, SampleFormat)?

but I'm not sure whether I should bake that into this PR or save it for future work.

ishitatsuyuki: Shape sounds really unfamiliar so I'm afraid it could cause confusion for users. Ideally we should copy a term from other libraries, but so far I don't know one.

Ralith: I'm happy to make the change into `Format` if that sounds good to people.

mitchmindtree: I agree `Format` seems nicer than `Shape`. I've always felt `Format` is also a little odd considering the fields, but I'll leave this bikeshed for another issue (#370) and future PR :)

PR #371, opened 2020-01-28
Rename stream `Format` types to `StreamConfig` and other related renamings.

### LINKED ISSUE #370: Change `Format` to something more suitable/flexible
It has always seemed a little odd to me to include the sample rate within the stream's "format" type. Fields like `channels` and `interleaved` make sense as these describe the arrangement/layout of the streams' buffers, but sample rate seems a little out of place in a `Format` type.

Some other more general/flexible names that come to mind:

- `StreamParams/Parameters` It seems PortAudio [use this](http://portaudio.com/docs/v19-doxydocs/structPaStreamParameters.html#aa1e80ac0551162fd091db8936ccbe9a0).
- `StreamDesc/Description`
- `StreamConfig/Configuration`

I quite like `StreamParams` and `StreamConfig`.

These more general names might also make more sense for potential future fields like `desired_buffer_len` or others we haven't considered just yet.

Perhaps we could keep `StreamFormat` around for grouping the `channels` and `interleaved` (#367) fields, and have `StreamFormat` as a field within `StreamParams`/`StreamConfig`?

mitchmindtree:
This implements the changes described at #370.

Closes #370.

Backends implemented:

- [x] null
- [x] emscripten
- [x] asio
- [x] wasapi
- [x] alsa
- [x] coreaudio
- [x] docs updated
- [x] examples updated

Edit: here is the full list of renamings to make it easier for downstream crates to adapt:

- `Format` type's `data_type` field renamed to `sample_format`.
- `Shape` -> `StreamConfig` - The configuration input required to build a stream.
- `Format` -> `SupportedStreamConfig` - Describes a single supported stream configuration.
- `SupportedFormat` -> `SupportedStreamConfigRange` - Describes a range of supported configurations.
- `Device::default_input/output_format` -> `Device::default_input/output_config`.
- `Device::supported_input/output_formats` -> `Device::supported_input/output_configs`.
- `Device::SupportedInput/OutputFormats` -> `Device::SupportedInput/OutputConfigs`.
- `SupportedFormatsError` -> `SupportedStreamConfigsError`
- `DefaultFormatError` -> `DefaultStreamConfigError`
- `BuildStreamError::FormatNotSupported` -> `BuildStreamError::StreamConfigNotSupported`

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -31,24 +31,25 @@
 //! let device = host.default_output_device().expect("no output device available");
 //! ```
 //!
-//! Before we can create a stream, we must decide what the format of the audio samples is going to
-//! be. You can query all the supported formats with the `supported_input_formats()` and
-//! `supported_output_formats()` methods. These produce a list of `SupportedFormat` structs which
-//! can later be turned into actual `Format` structs. If you don't want to query the list of
-//! formats, you can also build your own `Format` manually, but doing so could lead to an error
-//! when building the stream if the format is not supported by the device.
+//! Before we can create a stream, we must decide what the configuration of the audio stream is
+//! going to be. You can query all the supported configurations with the
+//! `supported_input_configs()` and `supported_output_configs()` methods. These produce a list of
+//! `SupportedStreamConfigRange` structs which can later be turned into actual
+//! `SupportedStreamConfig` structs. If you don't want to query the list of configs, you can also
+//! build your own `StreamConfig` manually, but doing so could lead to an error when building the
+//! stream if the config is not supported by the device.
 //!
-//! > **Note**: the `supported_formats()` method could return an error for example if the device
-//! > has been disconnected.
+//! > **Note**: the `supported_input/output_configs()` methods could return an error for example if
+//! > the device has been disconnected.
 //!
 //! ```no_run
 //! use cpal::traits::{DeviceTrait, HostTrait};
 //! # let host = cpal::default_host();
 //! # let device = host.default_output_device().unwrap();
-//! let mut supported_formats_range = device.supported_output_formats()
-//!     .expect("error while querying formats");
-//! let format = supported_formats_range.next()
-//!     .expect("no supported format?!")
+//! let mut supported_configs_range = device.supported_output_configs()
+//!     .expect("error while querying configs");
+//! let supported_config = supported_configs_range.next()
+//!     .expect("no supported config?!")
 //!     .with_max_sample_rate();
 //! ```
 //!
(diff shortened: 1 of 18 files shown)


mitchmindtree:
I've also added a
```rust
impl From<SupportedStreamConfig> for StreamConfig
```
implementation, but I'm unsure if it would be better to have an inherent impl on `SupportedStreamConfig` along the lines of
```rust
fn config(&self) -> StreamConfig
```
instead? Perhaps both? My thinking is the inherent impl may 1. make it easier to avoid type inference ambiguity and 2. reduce the need for `.clone()` calls on `SupportedStreamConfig` seeing as it does not impl `Copy` (even though it could in theory).

src/traits.rs
```diff
@@ -84,9 +84,9 @@ pub trait HostTrait {
 /// methods that involve a device return a `Result` allowing the user to handle this case.
 pub trait DeviceTrait {
     /// The iterator type yielding supported input stream formats.
```
msiglreith, line 86: `format` should be replaced, same below

src/lib.rs
```diff
@@ -235,6 +218,17 @@ pub struct Data {
     sample_format: SampleFormat,
 }
 
+impl SupportedStreamConfig {
+    /// Construct a `SupportedStreamConfig` from an existing `StreamConfig`.
+    pub fn from_config(conf: &StreamConfig, fmt: SampleFormat) -> Self {
```
msiglreith: `supported` confuses me a bit here as is it guarenteed that it is actually supported by the backend if I can just craft it myself?

mitchmindtree: Good point! I remember this crossing my mind also now that you mention it, but I didn't get around to addressing it.

Perhaps we *should* make it so that `SupportedStreamConfig`'s can only be constructed via either the `Device` or `SupportedStreamConfigRange`, make this `from_config` constructor `pub(crate)`, and expose the fields via methods instead? This way any instance of a `SupportedStreamConfig` can assumed to actually be supported (edit: from the `Device` that created it that is... nothing stopping the user using it with a non-supporting device)?

It's worth keeping in mind that users intending to use `build_input/output_stream_raw` currently must use a `SupportedStreamConfig`, so it would be especially important that `Device::supported_input/output_formats` is accurate and thorough to ensure we aren't restricting users from opening streams that are in fact supported by the device and host. Alternatively, to avoid this issue we could change the `build_input/output_stream_raw` API to accept a `StreamConfig` and `SampleFormat` as separate arguments, rather than expecting a `SupportedStreamConfig`.

src/traits.rs
mitchmindtree: Agh I knew I'd miss one somewhere

src/lib.rs
Ralith: The proposed change to the raw stream APIs seems natural, but bear in mind that this is a type with public fields, which can therefore be constructed by hand regardless.

Issue #363, opened 2020-01-17
Proposal for minimal timestamp API to allow for synchronising media with CPAL streams


mitchmindtree:
This is a proposal to begin addressing #279 with the most minimal API necessary.

## Background

The most seemingly accurate and thorough research I could come across on this topic is Ross Bencina's excellent paper *PortAudio and Media Synchronisation - It's All in the Timing*. It contains an overview of the media synchronisation problem with example scenarios, visual diagrams, etc that make it more intuitive.

http://www.portaudio.com/docs/portaudio_sync_acmc2003.pdf

The first few sections of the paper describe some hypothetical scenarios and different techniques for synchronising audio with some other kind of media. A MIDI clock is the primary example used in the paper, but the same techniques apply to presenting frames of graphics and other forms of media sync.

Section 6 describes the minimal set of information necessary in order to make these synchronisation techniques possible:

- **Sample rate**. We already have this.
- **Buffer start times**. We do not yet provide this. This refers to the most accurate form of monotonic clock time available on the system. It is also essential that users have access to the same source of time that provides this value in order to timestamp their media events. This means 1. describing the exact source for each host in the docs and possibly 2. providing a function for easily retrieving this value (portaudio do so via a `GetStreamTime(Stream* s)` function).

PortAudio decided to provide this monotonic time in seconds using a double-precision floating-point data type:

> The double   data   type   was   chosen   after   considerable deliberation because  it  provides  sufficient  resolution to   represent   time   with high-precision,   may   be manipulated   using   numerical   operators, and is   a standard  part  of  the  C  and  C++  languages.

Section 7 also describes implementation issues. They can be roughly summed up as follows:

- 7.1 **Sample rates**: Subtle variations between the nominal sample rate and the observed sample rate occur between sound card / chipsets, resulting in subtle inaccuracies occurring within the aforementioned synchronisation techniques. PortAudio provides an actual sample rate via its stream info parameters. Calculating this requires a high-resolution system clock, though this isn't always available.
- 7.2 **One Shared Time-base**: The time source of timestamps provided via audio callbacks sometimes differ from the source used to provide timestamps for other media events (e.g. Windows' MIDI API). This is why PortAudio found it necessary to provide a function for easy access to the correct source (`GetStreamTime`).
[...]

PR #375, opened 2020-02-08
Use bindgen to generate alsa-sys bindings

mitchmindtree:
While working on #363 I noticed I was missing a few items from the ALSA
lib. Rather than adding them by hand, I thought I would update the crate
using the latest version of bindgen. I'm on the NixOS unstable channel
and have used the `alsaLib.dev` package headers as the source for
bindings generation.

I noticed that bindgen also generated the parts of `libc` that we were
using, so I removed the `libc` dep in favour of the generated bindings.
I'm not sure what the best-practise approach is here - that is, whether
it is better to use `libc` where possible, or to use what bindgen gives
us. rust-bindgen itself doesn't give us an option to use `libc`, so I
think you'd have to use some extensive black/white-listing to get it
working.

There are a few warnings in the generated bindings related to u128 types
having an unstable ABI. This seems to be a bug in
rust-lang/rust-bindgen#1560. Rather than fixing this by hand I'd prefer
to regenerate the bindings once the bug has been fixed.


mitchmindtree:
I've [posted to the rust forums](https://users.rust-lang.org/t/best-practice-when-working-with-rust-bindgen-and-libc/38071) in order to get some feedback on best practice when working with `rust-bindgen` and `libc`.

cc @Ralith from memory you're also on Linux - would you mind having a quick look / sharing your thoughts on this PR?

Ralith:
I think best practice here is to use whitelisting and rely on std or libc for definitions originating outside the lib being bound. This helps keep the total amount of code rustc has to digest down. It seems like `snd.*` should get us pretty far whitelistwise. I also find it pretty confusing to see things like `alsa::close` when it's actually not an alsa function.

mitchmindtree:
Thanks for the feedback @Ralith! I've have a go at your suggestions tomorrow. I think I'll include the script used to generate the bindings in the `alsa-sys` directory as well to make it a little easier for folks to keep up to date in the future too.

mitchmindtree:
Coming back to this after a long-winded side-track! I think for now it's best to halt work on this until #386 is resolved.

mitchmindtree:
Closing in favour of #386.

PR #379, opened 2020-02-24
Remove redundant 'static from &'static str

est31:
Specifying the static lifetime in constants and static variables
is redundant since Rust 1.17.0

https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1170-2017-04-27


mitchmindtree:
Much nicer, thanks!

PR #393, opened 2020-04-15
Recover if hw_params_set_buffer_time failed.

derekdreery:
This PR leaves the buffer time unrestricted if restricting it failed. Failure will only happen with buggy alsa drivers, so this allows users with those drivers (e.g. me :P) to use projects built on `cpal`. It should have no impact on users where the function succeeds.

### DIFF
--- a/src/host/alsa/mod.rs
+++ b/src/host/alsa/mod.rs
@@ -8,10 +8,10 @@ use crate::{
     StreamConfig, StreamError, SupportedStreamConfig, SupportedStreamConfigRange,
     SupportedStreamConfigsError,
 };
-use std::{cmp, mem};
 use std::sync::Arc;
 use std::thread::{self, JoinHandle};
 use std::vec::IntoIter as VecIntoIter;
+use std::{cmp, mem};
 use traits::{DeviceTrait, HostTrait, StreamTrait};
 
 pub use self::enumerate::{default_input_device, default_output_device, Devices};
--- a/src/host/alsa/mod.rs
+++ b/src/host/alsa/mod.rs
@@ -799,6 +799,7 @@ fn set_hw_params_from_format<'a>(
     // If this isn't set manually a overlarge buffer may be used causing audio delay
     let mut hw_params_copy = hw_params.clone();
     if let Err(_) = hw_params.set_buffer_time_near(100_000, alsa::ValueOr::Nearest) {
+        // Swap out the params with errors for a snapshot taken before the error was introduced.
         mem::swap(&mut hw_params_copy, &mut hw_params);
     }
 
--- a/src/host/alsa/mod.rs
+++ b/src/host/alsa/mod.rs
@@ -8,7 +8,7 @@ use crate::{
     StreamConfig, StreamError, SupportedStreamConfig, SupportedStreamConfigRange,
     SupportedStreamConfigsError,
 };
-use std::cmp;
+use std::{cmp, mem};
 use std::sync::Arc;
 use std::thread::{self, JoinHandle};
 use std::vec::IntoIter as VecIntoIter;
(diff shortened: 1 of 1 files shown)


src/host/alsa/mod.rs
```diff
@@ -962,14 +962,21 @@ unsafe fn set_hw_params_from_format(
...
         &mut 0,
     )) {
-        return Err(format!("buffer time could not be set: {}", e));
+        // Go back to copy from before setting buffer time if there was an error.
+        //
+        // This is a workaround for buggy alsa drivers where calling
+        // `snd_pcm_hw_params_set_buffer_time_near` errors, even though calling the function should
+        // not restrict the params space to the empty set.
+        std::mem::swap(&mut hw_params_copy, &mut hw_params);
+        //eprintln!("warning: buffer time could not be set: {}", e);
```
richard-uk1: I left this in to let you decide whether any warning should be emitted. It should be uncommented or deleted before any merge.

derekdreery:
Aah bad timing - I'll rebase to use alsa. :)

Ralith:
Should cpal really be clobbering the platform's defaults like this to begin with? If an application requires low latency, there should be a way for them to set buffer sizes as appropriate on all platforms, and for those that don't, the platform defaults may be more appropriate. I wonder if this might even be the root cause for https://github.com/RustAudio/cpal/issues/272.

ishitatsuyuki:
@Ralith Setting buffer size is not super portable (not configurable for JACK, whether this is possible is unknown for Apple platforms), so I'm not in favor in exposing it on the polymorphic Host. (Exposing them as options for individual Host is fine.) Instead I think we could probably expose a low-latency hint, so it would be passed on for APIs that accept it, or otherwise choose a decently short buffer size.

Talking about ALSA, the 100ms default was copied from mpv and the reason we don't use the defaults is because PulseAudio will actually give you an insanely long buffer if you don't specify.

mitchmindtree:
> so I'm not in favor in exposing it on the polymorphic Host

@ishitatsuyuki I tend to agree with @Ralith, I think we should at least provide an option for *requesting* buffer sizes of a specific length, with a large note in the docs indicating that the user should not expect the slice delivered to the audio callbacks to actually have this size. PortAudio takes this approach and documents the platform-specific caveats.

A little more context: Having the ability to change the buffer size is essential to a lot of pro-audio applications like DAWs, where you request a buffer size that suits the latency requirements of your project. E.g. during recording sessions low-latency is a must, on the other hand during the mixing and sound design stage effects processing can be highly intensive and increasing buffer size can reduce glitching and improve CPU performance. Taking a look at #205 it seems there is quite a bit of interest in this. 

@JoshuaBatty is currently investigating what APIs are exposed on each platform. Once I finish my work on #363 I will assist him with proposing an API at #205 and landing a PR.

derekdreery:
Maybe you could call the generic buffer size method `buffer_size_hint` - I think this is how some GUIs name their size functions where the window manager may ignore or change the requested size.

I think this PR is orthogonal to any decisions about exposing buffer sizing - it would still be nice for alsa implementations that don't support setting the buffer size to degrade to their default rather than fall over.

On my setup the driver causing problems is the jack alsa driver. I'm not 100% how it works but I think jack intercepts the alsa client request and creates a corresponding jack client.

mitchmindtree:
I agree @derekdreery, I was imagining something similar, e.g. a `desired_buffer_size` field in the `StreamConfig` struct, but `buffer_size_hint` is also nice.

W.r.t. this PR, I think I'm OK with this landing as a temporary workaround if it allows some folks to at least get a stream running in the meantime. @Ralith @ishitatsuyuki  does that sound OK with you?

derekdreery:
@mitchmindtree you may want to keep it even with a `buffer_size` method - all it does is change from failure when setting the buffer size doesn't work to using the default buffer size. :)

ishitatsuyuki:
It seems that this PR is all about JACK rejecting buffer size changes - so providing a hint is out of this PR's scope. (I'm fine with adding it in another PR.)

It would be good to see some references on how other applications deal with this as ALSA doesn't really have concrete rules for this kind of things.

Ralith:
I'm not equipped to judge this PR in and of itself, but I have no objection to it preserving pre-existing behavior, even if we want to replace that behavior in the future.

PR #394, opened 2020-04-15
Add noise example

derekdreery:
An example using random noise.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,7 @@ num-traits = "0.2.6"
 anyhow = "1.0.12"
 hound = "3.4"
 ringbuf = "0.1.6"
+rand = { version = "0.7.3", features = ["small_rng"] }
 
 [target.'cfg(target_os = "windows")'.dependencies]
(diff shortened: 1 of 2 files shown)


mitchmindtree:
Thanks for the PR @derekdreery !

I'm a little wary of adding this - while it's nice for users to see how to generate noise, doing so is not specific to CPAL, and apart from this the example does not demonstrate any parts of CPAL's API that are not already demonstrated within the other examples. In particular, it looks like this is a copy of the `beep.rs` example, where the only change is the [function for producing samples](https://github.com/RustAudio/cpal/pull/394/files#diff-b8af1ea3f8ad5bb48d2fabe0222b4f72R31) generates random values rather than a sine wave. I'm not sure we want to set the precedent for adding examples demonstrating more general DSP stuff when CPAL is already a little tight on maintenance. What are your thoughts @derekdreery ?

derekdreery:
Yeah I'm happy for you to take or leave this PR, just thought I'd offer it. The reason I use it is that because of resonance from a particular frequency (beep) it's hard to tell if my speakers are working properly (equal volume etc.), whereas noise doesn't have any particular frequency.

So yeah feel free to close this if you prefer. :)

PR #397, opened 2020-04-27
[WIP] Timestamp and `StreamInstant` APIs

mitchmindtree:
This PR addresses #363. 

Currently, the PR only contains the top-level API but does not yet provide an implementation for each of the hosts. I'm posting this early in order to allow for feedback before diving too deeply into the implementation for each host.

- [x] API
- Platforms:
  - ~null~ N/A
  - [x] alsa - mitchmindtree/cpal#5
  - [x] coreaudio
  - [x] wasapi
  - [x] asio
  - [x] emscripten
- [x] Document `StreamInstant` source for each platform
- [ ] `stream.now()`? Possibly future PR.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,6 +34,7 @@ libc = "0.2.65"
 [target.'cfg(any(target_os = "macos", target_os = "ios"))'.dependencies]
 coreaudio-rs = { version = "0.9.1", default-features = false, features = ["audio_unit", "core_audio"] }
 core-foundation-sys = "0.6.2" # For linking to CoreFoundation.framework and handling device name `CFString`s.
+mach = "0.3" # For access to mach_timebase type.
 
 [target.'cfg(target_os = "emscripten")'.dependencies]
 stdweb = { version = "0.1.3", default-features = false }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ hound = "3.4"
 ringbuf = "0.1.6"
 
 [target.'cfg(target_os = "windows")'.dependencies]
-winapi = { version = "0.3", features = ["audiosessiontypes", "audioclient", "coml2api", "combaseapi", "debug", "devpkey", "handleapi", "ksmedia", "mmdeviceapi", "objbase", "std", "synchapi", "winbase", "winuser"] }
+winapi = { version = "0.3", features = ["audiosessiontypes", "audioclient", "coml2api", "combaseapi", "debug", "devpkey", "handleapi", "ksmedia", "mmdeviceapi", "objbase", "profileapi", "std", "synchapi", "winbase", "winuser"] }
 asio-sys = { version = "0.1", path = "asio-sys", optional = true }
 parking_lot = "0.9"
 
--- a/asio-sys/src/bindings/mod.rs
+++ b/asio-sys/src/bindings/mod.rs
@@ -84,8 +84,15 @@ pub struct SampleRate {
     pub rate: u32,
 }
 
+/// Information provided to the BufferCallback.
+#[derive(Debug)]
+pub struct CallbackInfo {
+    pub buffer_index: i32,
+    pub system_time: ai::ASIOTimeStamp,
+}
+
 /// Holds the pointer to the callbacks that come from cpal
-struct BufferCallback(Box<dyn FnMut(i32) + Send>);
+struct BufferCallback(Box<dyn FnMut(&CallbackInfo) + Send>);
 
 /// Input and Output streams.
 ///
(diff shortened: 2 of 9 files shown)


mitchmindtree:
The ALSA implementation is just about ready [here](https://github.com/mitchmindtree/cpal/pull/5), however there are two issues to address before merging that branch into this one. The issues are mentioned in the linked PR.

The missing ALSA API is an easy fix, but I'm less sure how to address the issue where ALSA does not seem to report valid timestamps for all devices. cc @diwic  @Ralith @ishitatsuyuki + any other linux audio devs - any ideas on how to handle this? I'm going to take a break from ALSA for the moment and move onto another backend, but will come back to this soon.

diwic:
@mitchmindtree Looking at the alsa-lib code, I think we should just implement it without worrying too much about it. The API was added in 2014 so if you're happy with not supporting ALSA earlier than 2015ish then go for it.

mitchmindtree:
@diwic Thanks for sharing! My concern is not so much with backwards compatibility, but rather that ALSA's `get_htstamp` API doesn't return a valid value when targeting the "pulse" device.

Here's the line: https://github.com/mitchmindtree/cpal/pull/5/files#diff-5befd78a846aaf969baa67643878cd6aR726 - it returns a `timespec` with zeroed `tv_sec` and `tv_nsec` fields on my system. I'm running GNOME3 on NixOS and so the default ALSA device is pulse.

Seeing as currently ALSA is the only Host supported on Linux (pending #389), I think many linux CPAL users would likely run into this issue. I'm also not sure if there is a reliable way to test whether or not `get_htstamp` will be valid for a given device (without calling it and testing for `0`, `0` that is).

mitchmindtree:
It looks like the portaudio devs ran into this same problem and consider it a bug in the alsa-pulse plugin https://portaudio.music.columbia.narkive.com/cJ4dWvTd/getstreamtime-with-alsa-hostapi

It's fine for portaudio to ignore this as most folks on a pulseaudio system would just use the proper pulseaudio host provided by portaudio. However, we don't have a proper pulseaudio host and as such can't really rely on this. I would wager the vast majority of linux cpal users are on a distro that specifies pulseaudio as the default alsa device and as such, the vast majority of linux cpal users would have broken timestamps if we let this slide.  

I guess one possible solution would be to call `get_htstamp` while building the stream to check whether or not the device is returning zeroed timestamps. If so, we can instead just use the duration since the stream started for timestamps via `std::time::Instant::now` before each callback? Generally any kind of unnecessary I/O should be avoided in the audio callback, but the cost of `std::time::Instant` should really be negligible (especially using pulseaudio via alsa) and surely better than producing invalid (or no) timestamps?

Ralith:
`Instant::now` is pretty complex due to all the effort needed to guarantee monotonicity, but I think hacking around alsa bugs is in general acceptable, given that we'll eventually want native pulse support anyway. Might also be worth checking how hard upstreaming a fix would be.

diwic:
@mitchmindtree I'm not an expert on clocks and timestamping, but it sounds fine to me. As you say, if the only one not providing this is the pulse api, then the overhead of timestamping should be negligble.

mitchmindtree:
OK, this PR now has the implementations for alsa and emscripten. I have local branches with implementations for the remaining branches (asio, coreaudio, wasapi), however I still need to test them on their respective platforms before merging them into this PR.

diwic:
@mitchmindtree Btw, if your problem is that pulse is not producing timestamps and you want to test timestamping, then here are four ways to disable pulseaudio, in order of preference/simplicity:

 * If CPAL allows you do use another ALSA device, such as `plughw:0` (where `0` is your sound card, see `/proc/asound/cards` for a list), then just wait a few seconds to have pulseaudio release the sound card and then start CPAL
 * If CPAL does not support that, run `pasuspender cmd` where `cmd` could be either the shell or the CPAL app.
 * If that does not work either, you should be able to create your own `.asoundrc` and override `pcm.!default` there.
 * And if that does not work, then in my case (Ubuntu), there is a `/usr/share/alsa/alsa.conf.d/pulse.conf` file that checks if pulseaudio is running, and if so includes `/usr/share/alsa/pulse.conf` that sets `pcm.!default`. So modifying/disabling these should do the trick. I'm guessing things look the same on NixOS.

mitchmindtree:
OK just tested this on windows and macos with @JoshuaBatty and it seems to be working nicely across all platforms!

The `Stream::now` method is still missing, however this should be easy enough to add in a future PR as the process for producing `StreamInstant`s has already been nicely isolated into small functions within each backend implementation.

I'll leave this open over the weekend for review, then we'll likely dive into #205.

Issue #398, opened 2020-05-01
macOS: A backend-specific error has occurred: core foundation unexpectedly returned null string


thavlik:
Mac Mini 2018 (Catalina 10.15.4)

Calling `device.name()` returns the error string `A backend-specific error has occurred: core foundation unexpectedly returned null string` for my audio interfaces. I am opening this issue to document its existence. Will make audio hardware specs known if it ends up being necessary, but I'm assuming this hardware-specific since the issue doesn't exist yet.

Thanks a ton for cpal. Everything else works really well :)

thavlik:
Looks like there is a slow route available for getting the device name that can serve as a fallback. I'm guessing my driver doesn't want to give up its name without you supplying the memory, and that makes sense. Now it'll use the existing (fast) code path and fallback to creating a buffer on the stack to call CFStringGetCString if CFStringGetCStringPtr returns nil.

PR #399, opened 2020-05-01
macOS: fallback to CFStringGetCString if CFStringGetCStringPtr returns nil

thavlik:
Fixes https://github.com/RustAudio/cpal/issues/398


thavlik:
Thanks! Here's an explanation of why this works on Apple docs https://developer.apple.com/documentation/corefoundation/1542133-cfstringgetcstringptr

hiteshjoshi:
Yes, I used 0.11.0

PR #416, opened 2020-05-25
Add ishitatsuyuki's wasm-beep example

mitchmindtree:
This adds the `wasm-beep` example @ishitatsuyuki put together for
testing webaudio support. The example helped me to better understand how
to use the webaudio host and provided an easy way to test it, I'm sure
other contributors/users might benefit in the same way.

I guess in the future it would be nice if we could have some way of
sharing the same code between wasm-target examples and native target
examples. Still, I think it's beneficial to include this wasm-beep
example in the meantime to have at least some kind of demonstration of
how to use cpal with wasm.

Are you happy for this to be added as is @ishitatsuyuki?


ishitatsuyuki:
Sure, why not!

PR #419, opened 2020-05-25
Fix CoreAudio warnings.

ely-uf:
Hi, discovered 7 warnings issued by `rustc 1.42.0 (b8cedc004 2020-03-09)`.

`cargo build` and `cargo test` run successfully with the changes in this PR.

The list of warnings fixed:
1. warning: unnecessary `unsafe` block [L184]
2. warning: use of deprecated item 'std::error::Error::description': use the Display impl or to_string() [L330, L817, L831, L846]
3. warning: field is never read: `device_id` [L411]


mitchmindtree:
Thanks so much!

Issue #322, opened 2019-08-23
ALSA cpu usage


alexheretic:
Hello, I recently had a look at Robo Instructus' cpu usage and noticed the audio usage seems high. It's not a huge problem, but I thought I'd raise this in case it tickles the imagination.

I profiled the game without any music, so just occasional sounds and mostly silence. Rodio maintains a single stream and (I presume) will continually write silent bytes -> output.

cpal event loop `run_inner` is the majority of the game's sampled usage.

Full: [robo-instructus-flamegraph.zip](https://github.com/RustAudio/cpal/files/3534694/robo-instructus-flamegraph.zip)

![](https://user-images.githubusercontent.com/2331607/63596203-98356f00-c5b2-11e9-8794-55fbe1affac9.png)

balping:
Is it better with this: https://github.com/RustAudio/cpal/pull/301 ?

alexheretic:
> Is it better with this: #301 ?

Maybe. It doesn't look like that change will be easy for me to swap in, it'll require a bunch of rodio changes etc.

alexheretic:
Ok I've reworked rodio to #301 locally, it actually makes rodio simpler.

In terms of CPU usage the news is even better as rodio & cpal are no longer visible on the flamegraph. So #301 should solve this issue.
[robo-instructus-cpal-pr301-flame.zip](https://github.com/RustAudio/cpal/files/3535801/robo-instructus-cpal-pr301-flame.zip)
[robo-instructus-cpal-0.10-flame.zip](https://github.com/RustAudio/cpal/files/3535802/robo-instructus-cpal-0.10-flame.zip)

### cpal 0.10
![](https://user-images.githubusercontent.com/2331607/63616829-528f9b00-c5e0-11e9-99fd-9100fa0f6c54.png)

### cpal #301
![](https://user-images.githubusercontent.com/2331607/63616834-56bbb880-c5e0-11e9-90a3-381e141f6fba.png)

ishitatsuyuki:
First I doubt that #301 would make a difference. It can optimize lock handling for other platforms, but for ALSA there's already built-in synchronization and we don't have contention with it.

So please try retaking your profiles.

If that's actually being a bottleneck, though, I think [this](https://github.com/RustAudio/cpal/blob/master/src/host/alsa/mod.rs#L1057) might be an issue. I'll file a PR after testing that it works. It showed no CPU usage difference (with the beep example), though.

alexheretic:
[...]

PR #431, opened 2020-06-18
[Alsa Host] Add a default period time

### LINKED ISSUE #369: Problem with playback under arch linux.
Hi

When trying out the latest master I get.

```
cargo run --example beep
warning: field is never used: `can_pause`
   --> src/host/alsa/mod.rs:501:5
    |
501 |     can_pause: bool,
    |     ^^^^^^^^^^^^^^^
    |

    Finished dev [unoptimized + debuginfo] target(s) in 0.31s
     Running `target/debug/examples/beep`
Error: A backend-specific error has occurred: buffer time could not be set: Invalid argument
```
I'm on arch Linux using Jack, all packages recently updated and other sound applications (Jack, pulseaudio, alsa) works as expected.

```
cargo run --example enumerate
warning: field is never used: `can_pause`
   --> src/host/alsa/mod.rs:501:5
    |
501 |     can_pause: bool,
    |     ^^^^^^^^^^^^^^^
    |

   Compiling cpal v0.11.0 (/home/pln/rust/adder/cpal)
    Finished dev [unoptimized + debuginfo] target(s) in 1.92s
     Running `target/debug/examples/enumerate`
Supported hosts:
  [Alsa]
Available hosts:
  [Alsa]
ALSA
  Default Input Device:
    Some("default")
  Default Output Device:
    Some("default")
  Devices: 
  1. "jack"
    Default input stream format:
      Format { channels: 2, sample_rate: SampleRate(44100), data_type: F32 }
    All supported input stream formats:
      1.1. SupportedFormat { channels: 2, min_sample_rate: SampleRate(44100), max_sample_rate: SampleRate(44100), data_type: F32 }
    Default output stream format:
      Format { channels: 2, sample_rate: SampleRate(44100), data_type: F32 }
    All supported output stream formats:
      1.1. SupportedFormat { channels: 2, min_sample_rate: SampleRate(44100), max_sample_rate: SampleRate(44100), data_type: F32 }
ALSA lib pcm_dmix.c:1089:(snd_pcm_dmix_open) unable to open slave
ALSA lib pcm_dsnoop.c:641:(snd_pcm_dsnoop_open) unable to open slave
ALSA lib pcm_usb_stream.c:508:(_snd_pcm_usb_stream_open) Unknown field hint
ALSA lib pcm_usb_stream.c:508:(_snd_pcm_usb_stream_open) Unknown field hint
  2. "hdmi:CARD=HDMI,DEV=0"
Error: DeviceNotAvailable
ALSA lib pcm_usb_stream.c:508:(_snd_pcm_usb_stream_open) Unknown field hint
ALSA lib pcm_usb_stream.c:508:(_snd_pcm_usb_stream_open) Unknown field hint
```

I have a similar setup on my laptop, arch, Jack etc. And there it works, so either my setup is wrong, or I hit a bug in `cpal`.

Any thoughts/ideas. What to try next?

### LINKED ISSUE #391: Problem when setting buffer time (alsa backend)
Continuing from #247, I'm still getting the problem (the story is I removed the set_buffer call but it was re-introduced as it created unacceptable latency). I've implemented dumping the `HwParams` (code below) and to be honest it looks like a bug in the ALSA driver for my sound card. I have an idea to make it work on my setup - that is to try to set the buffer time, and just to leave it unconstrained if there is an error setting it. I'll knock up a PR with the idea in it.

<details>
<summary>Code to implement `Debug` for `HwParams`</summary>

```rust
impl fmt::Debug for HwParams {
    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
        unsafe {
            match dump_alsa(|outputp| alsa::snd_pcm_hw_params_dump(self.0, outputp) == 0) {
                Some(msg) => write!(f, "HwParams({})", msg),
                None => f.write_str("HwParams")
            }
        }
    }
}

/// output of dump_fn should be whether the function ran successfully (false means an error)
unsafe fn dump_alsa(dump_fn: impl FnOnce(*mut alsa::snd_output_t) -> bool) -> Option<String> {
    let mut outputp: *mut alsa::snd_output_t = ptr::null_mut();
    if let Err(e) = check_errors(alsa::snd_output_buffer_open(&mut outputp)) {
        return None;
    }
    if ! dump_fn(outputp) {
        return None;
    }
    let mut output_cstr: *mut libc::c_char = ptr::null_mut();
    let len = alsa::snd_output_buffer_string(outputp, &mut output_cstr);
    // be careful to drop this before any mutating calls to outputp
    let output_cstr: &[u8] = std::slice::from_raw_parts(output_cstr as *const u8, len);
    Some(String::from_utf8_lossy(output_cstr).into_owned())
}
```
</details>

sniperrifle2004:
This will accomplish two things, in addition to what the buffer
time already provides:
  - With a latency of 100ms it doesn't make sense to wake up the
    cpu for new audio data more often than this. Maintaining a lower
    period time will create unnecessary load. This is an effect
    that stacks if the alsa device is virtual (Like the pulse plugin,
    which happens to have a low default period time, which stands
    in stark contrast to its reportedly high buffer). Most of the
    time 50ms would also work, but its better to have some fallout
    protection in the default scenario
  - It prevents errors trying to set the buffer time.

I removed the workaround introduced earlier for the second issue. While it's still not entirely clear why it should actually raise an error there if you do not first set the period time, there is something to be said for first choosing the period time in general, since it does restrict the possible buffer sizes.

Fixes #369, Fixes #391 (probably)

Now I also strongly suspect a good period time will have a positive effect for #322, but that should be verified. Since pulse appears involved I am optimistic.

EDIT: I now think that period size first is a requirement that comes with devices that require buffer sizes to be multiples of the period size. If that period size is not defined it is not yet possible to set a buffer size even though the minimum and maximum are well defined. That is probably why `set_buffer_time_near` fails.

### DIFF
--- a/src/host/alsa/mod.rs
+++ b/src/host/alsa/mod.rs
@@ -869,7 +869,7 @@ fn set_hw_params_from_format<'a>(
     config: &StreamConfig,
     sample_format: SampleFormat,
 ) -> Result<alsa::pcm::HwParams<'a>, BackendSpecificError> {
-    let mut hw_params = alsa::pcm::HwParams::any(pcm_handle)?;
+    let hw_params = alsa::pcm::HwParams::any(pcm_handle)?;
     hw_params.set_access(alsa::pcm::Access::RWInterleaved)?;
 
     let sample_format = if cfg!(target_endian = "big") {
--- a/src/host/alsa/mod.rs
+++ b/src/host/alsa/mod.rs
@@ -890,12 +890,10 @@ fn set_hw_params_from_format<'a>(
     hw_params.set_rate(config.sample_rate.0, alsa::ValueOr::Nearest)?;
     hw_params.set_channels(config.channels as u32)?;
 
-    // If this isn't set manually a overlarge buffer may be used causing audio delay
-    let mut hw_params_copy = hw_params.clone();
-    if let Err(_) = hw_params.set_buffer_time_near(100_000, alsa::ValueOr::Nearest) {
-        // Swap out the params with errors for a snapshot taken before the error was introduced.
-        mem::swap(&mut hw_params_copy, &mut hw_params);
-    }
+    // These values together represent a moderate latency and wakeup interval.
+    // Without them we are at the mercy of the device
+    hw_params.set_period_time_near(25_000, alsa::ValueOr::Nearest);
+    hw_params.set_buffer_time_near(100_000, alsa::ValueOr::Nearest);
 
     pcm_handle.hw_params(&hw_params)?;
 
--- a/src/host/alsa/mod.rs
+++ b/src/host/alsa/mod.rs
@@ -892,8 +892,8 @@ fn set_hw_params_from_format<'a>(
 
     // These values together represent a moderate latency and wakeup interval.
     // Without them we are at the mercy of the device
-    hw_params.set_period_time_near(25_000, alsa::ValueOr::Nearest);
-    hw_params.set_buffer_time_near(100_000, alsa::ValueOr::Nearest);
+    hw_params.set_period_time_near(25_000, alsa::ValueOr::Nearest)?;
+    hw_params.set_buffer_time_near(100_000, alsa::ValueOr::Nearest)?;
 
     pcm_handle.hw_params(&hw_params)?;
 


derekdreery:
Can confirm that it works for me! Awesome!

rfwatson:
I tested this change locally after also experiencing high CPU load in a simple ALSA event loop.

Nothing scientific, but at a glance building CPAL from this branch reduced CPU usage for the process from around 10% to ~2%. :rocket:

derekdreery:
Really excited about this landing - I will be able to use cpal again! :)

sniperrifle2004:
Superseded by #520.

PR #433, opened 2020-06-19
[Alsa Host] Set appropriate start threshold depending on the direction

### LINKED ISSUE #432: [Alsa Host] Start threshold of 0
I noticed that currently the start threshold is set to 0 frames:

https://github.com/RustAudio/cpal/blob/b177bcdf22be548763c6bab2ffc83415eb343151/src/host/alsa/mod.rs#L910

Now this doesn't seem like a good idea to me. I don't think this has well defined behaviour across devices and it also seems counter intuitive, to start recording/playing before there are any frames. For playback that threshold should probably be the buffer size, while for recording a single frame would work and the resulting semantics would be the same.

Using significantly less than the buffer size for playback could result in underruns (Which lead to all kinds of fun), though with the alsa host's current approach to buffering - "Just use whatever the device uses or well 100ms is okay I guess" - that can still happen for certain applications. Luckily #401 will change that). Now a period size (Or even a single frame) would also work, but it's risky. If an application relies on the buffer, because the source is bursting or can just about make the period time, not filling up the buffer before starting will result in underruns and no amount of changing the buffer size will fix that if the start_threshold is not appropriately set.

EDIT: The alsa pcm example uses a period less then a full buffer. I guess that would work as well.

sniperrifle2004:
I have taken the alsa pcm's approach of a period of the buffer size for playback. Hopefully that fills up the buffer since the next period is coming shortly and can probably be written well within the time it takes to play the first period in the buffer. I don't know if this assumption holds in extremely low latency situations, but if it doesn't I suspect underruns would start happening anyway.

Fixes #432

### DIFF
--- a/src/host/alsa/mod.rs
+++ b/src/host/alsa/mod.rs
@@ -211,7 +211,12 @@ impl Device {
             _ => None,
         };
 
-        handle.start()?;
+        // For capture streams, starts can't be delayed until the
+        // read calls since the input_stream_worker won't call those
+        // if the pcm is not started
+        if stream_type == alsa::Direction::Capture {
+            handle.start()?
+        }
 
         let stream_inner = StreamInner {
             channel: handle,
@@ -918,9 +923,11 @@ fn set_sw_params_from_format(
         }
         match stream_type {
             alsa::Direction::Playback => {
+                // Don't start playback until the buffer is filled
                 sw_params.set_start_threshold((buffer - period) as alsa::pcm::Frames)?
             }
-            alsa::Direction::Capture => sw_params.set_start_threshold(1)?,
+            // For capture streams rely on an explicit start
+            alsa::Direction::Capture => sw_params.set_start_threshold(i32::MAX.into())?,
         }
         sw_params.set_avail_min(period as alsa::pcm::Frames)?;
         let buffer = buffer as usize * config.channels as usize;
--- a/src/host/alsa/mod.rs
+++ b/src/host/alsa/mod.rs
@@ -189,7 +189,7 @@ impl Device {
             let hw_params = set_hw_params_from_format(&handle, conf, sample_format)?;
             hw_params.can_pause()
         };
-        let (_buffer_len, period_len) = set_sw_params_from_format(&handle, conf)?;
+        let (_buffer_len, period_len) = set_sw_params_from_format(&handle, conf, stream_type)?;
 
         handle.prepare()?;
 
@@ -905,9 +905,9 @@ fn set_hw_params_from_format<'a>(
 fn set_sw_params_from_format(
     pcm_handle: &alsa::pcm::PCM,
     config: &StreamConfig,
+    stream_type: alsa::Direction,
 ) -> Result<(usize, usize), BackendSpecificError> {
     let sw_params = pcm_handle.sw_params_current()?;
-    sw_params.set_start_threshold(0)?;
 
     let (buffer_len, period_len) = {
         let (buffer, period) = pcm_handle.get_params()?;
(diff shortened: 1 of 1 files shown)


mitchmindtree:
Thanks for diving into this! And thanks for taking a close look into the ALSA backend in general - I've already learned a few new things reading through your recent comments/PRs.

Perhaps we should land #401 first and then base this work on top of that?

sniperrifle2004:
I'm fine with that. I am still doing some research into this at the moment. It appears that on my machine the fact that `poll_descriptors_and_prepare_buffer` initially requests a (nearly) full buffer of frames the first time is sufficient to prevent buffer underruns even with the pcm starting while no data is available in the buffer. However whether this will always be the case is harder to say. It's not something I personally would want to rely on given my current understanding of the semantics, but it might also mean that I have misunderstood something in the semantics and that this is actually perfectly fine.

sniperrifle2004:
> 
> 
> It appears that on my machine the fact that `poll_descriptors_and_prepare_buffer` requests a (nearly) full buffer of frames the first time is sufficient to prevent buffer underruns even with the pcm starting while no data is available in the buffer.

That hypothesis can be scrapped. My understanding of the semantics involved was correct. I can now confirm that this will definitely result in under-runs. In fact it happens immediately at the `handle.start()` if you talk directly to a hardware device. So it's likely that features on the software level have been helpfully hiding this problem to some extend. Plugins like pulse/dmix/plug a lot of cards have for the `(sys)default` configuration. That might also explain why my less powerful system was stuttering despite not showing under-runs: Under-runs were occurring just not at the cpal level. I have now also reproduced this by talking to the `hw:0,0` device on my regular system: Immediate under-run.

src/host/alsa/mod.rs
```diff
@@ -211,7 +211,12 @@ impl Device {
             _ => None,
         };
 
-        handle.start()?;
+        // For capture streams, starts can't be delayed until the
+        // read calls since the input_stream_worker won't call those
+        // if the pcm is not started
+        if stream_type == alsa::Direction::Capture {
+            handle.start()?
+        }
```
alexmoon, line 219: If you set the `start_threshold` for capture streams to 1, I think you would not have to call `start()` here at all.

mitchmindtree: @alexmoon any chance you might be interested in getting this work over the line in a new PR? My understanding on this is a bit vague, but it sounds like this might help with some of the initial underruns that can occur when kicking off an ALSA stream?

alexmoon: @mitchmindtree yes, I can work on that. I'd probably like to combine this change with #431 since they're related and wait until #506 is merged so we don't have too many outstanding PRs touching the same files.

sniperrifle2004: @alexmoon You may be right. I vaguely recall trying `1` and it not working. I then hypothesized that it was due to `readi` not getting called at all if the pcm is not running and the `start_threshold` for recording relying on that to switch the pcm to running. But that memory is shaky at best and I am not an expert.

sniperrifle2004: @alexmoon Well it seems like I actually have some evidence to back up that memory in commit 3d1ecbe1eb2572d1bef656a464e2eb977807938c . The log message also goes into the technical details of the hypothesis. Note also that I thought it worked in the earlier commit since I had not actually removed the `handle.start()?` yet.

alexmoon: Yes, that seems to be correct. In my other implementation I was using readi() and a start_threshold of 1 worked correctly.

sniperrifle2004:
Superseded by #520

PR #437, opened 2020-06-29
Add some details to doc

HEnquist:
When I started reading on how to use cpal I was a little confused by the docs. This PR adds some details that would have helped me. This is work in progress and I will probably add more when I get further.

### DIFF
--- a/asio-sys/examples/enumerate.rs
+++ b/asio-sys/examples/enumerate.rs
@@ -61,6 +61,7 @@ fn main() {
                 };
                 println!("  Input {:?}", in_fmt);
                 println!("  Output {:?}", out_fmt);
+            }
             Err(err) => {
                 println!("Failed to load driver: {:?}, error: {}", name, err);
             }
--- a/asio-sys/examples/enumerate.rs
+++ b/asio-sys/examples/enumerate.rs
@@ -41,27 +41,29 @@ fn main() {
     let asio = sys::Asio::new();
     for name in asio.driver_names() {
         println!("Driver: {:?}", name);
-        if let Ok(driver) = asio.load_driver(&name) {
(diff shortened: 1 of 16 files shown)


asio-sys/examples/enumerate.rs
```diff
@@ -37,24 +41,27 @@ fn main() {
...
+            let out_fmt = Format {
+                channels: channels.outs as _,
+                sample_rate: SampleRate(sample_rate as _),
+                data_type: SampleFormat::F32,
+            };
+            println!("  Input {:?}", in_fmt);
+            println!("  Output {:?}", out_fmt);
+        } else {
+            println!("Failed to load driver: {:?}", name);
+        }
```
mitchmindtree, line 68: It would be nice if we could also include the error in the `"Failed to load driver..."` message, perhaps a `match` would be better suited here? E.g.

```rust
match asio.load_driver(&name) {
    Ok(driver) => {
        // ...
    }
    Err(err) => println!("Failed to load driver {:?}: {}", name, err),
}
```

If you're worried about indentation, you could do something like this instead:

```rust
let driver = match asio.load_driver(&name) {
    Ok(d) => d,
    Err(err) => {
        println!("Failed to load driver: {:?}: {}", name, err);
        continue;
    }
};
// ...
```

HEnquist: Good idea! I have updated as you suggested.

HEnquist:
There was a merge conflict so I rebased on the current master.

HEnquist:
I had completely forgotten about this PR. What is needed to finish this up?

HEnquist:
Happy birthday PR 437!
I'm closing since there seems to be very little interest in this.

Issue #284, opened 2019-06-13
Pause semantics


ishitatsuyuki:
Each platform has its own way of abstracting pauses, and cpal currently only expose a simple API (which probably has inconsistent behavior wrt following description).

Many driver seems to implement hardware level pauses, which IIUC can stop processing even in the middle of the buffer. When paused though there are two choices regarding the remaining buffer: flush (drain) until they're fully played, drop them completely (resetting the stream), or just keep them until the stream is resumed.

Interaction with OS mixers, though, is unknown as in theory hardware pause won't be used if another application is playing audio.

- On Windows, only `Start` and `Stop` is exposed, with a `Reset` function that clears playback position, achieving the "drop" behavior above.
- ALSA exposes a variety of APIs, but portability of those are unknown. The exposed API covers all of the behavior listed above.
- Android (AAudio) has a rather strange API that distinguish between a "pause" operation (output stream only) and a "stop" operation. Neither of them closes (destroys) the stream. The paused state has an additional option of getting flushed.

How should we expose these behaviors? Or if we decide to not do so, which behavior should we consistently achieve on every platform?

Johannesd3:
I'm experiencing the issue that some samples that are pulled before pausing are played after resuming (using ALSA). That's exactly what this issue is about?

roderickvd:
#1258 introduces `start`/`pause`/`stop` semantics with draining stop.

LastExceed:
#1258 is merged, so this can be closed as complete

Issue #197, opened 2018-02-01
ALSA: Provide info wether HW provides pause/play


Chief-Detektor:
Hi!
Thanks for this great library! :)
I was a little bit confused that a call to event_loop.pause and event_loop.pause had no effect on my system because I obviously does not have this capability on my HW. This is not a big deal but I thought I would be nice if cpal notifies the users that this feature is not available when calling it in debug mode.
I could prevent others from being confused or even open a bug ticket.

tomaka:
I agree that there should definitely be a way to report that. Probably one method for the capability, plus make the pause function return a result.

Chief-Detektor:
Switched my HW and now everything works as expected.
But I recognized that in some situations play and pause (alsa::snd_pcm_pause) might fail returning -77 c_int as error code. It might be desirable to pass the error to the user to indicate if the call failed or succeeded.

mitchmindtree:
@Chief-Detektor I'm currently thinking a bit about `pause` behaviour and I'm curious about your case. What device were you using that did not support `pause`?

I wonder what the effective difference is between creating and destroying the stream vs playing and pausing it. I also can't help but wonder how common it is that devices support "pause" and relative to how often they do not. In downstream projects that require playing/pausing streams frequently (e.g. in a DAW), is it simply up to the user to check for hw pause capabilities, pause if possible, otherwise just destroy their stream and then recreate it when they need to play again? Or perhaps they swap out the callback with one that just writes silence instead?

What got me wondering about this was how to handle streams that are still playing before returning from `EventLoop::run` #245. It's not super clear to me yet what the expected behaviour across all platforms.

It could be worth checking how JACK, PortAudio and others handle pausing.

The ALSA functions for reference:

- [snd_pcm_pause](http://www.alsa-project.org/alsa-doc/alsa-lib/group___p_c_m.html#gad711b582c7066bd4fffa1d08a04316b5)
- [snd_pcm_hw_params_can_pause](http://www.alsa-project.org/alsa-doc/alsa-lib/group___p_c_m___h_w___params.html#gaad4ba98b2f66baa5360b9a85f064a3e9)

kolbma:
There would be snd_pcm_drop/prepare.  
But the 1st question is, why pause() returns successful result when it doesn't pause and uses ~100% CPU in thread.  

I got a problem with this, because dropping and creating new streams in cpal is no option when using pipewire-alsa.  
The stream setup is so slow with pipewire in its graph.

[...]

PR #438, opened 2020-06-29
Add test coverage for cmp_default_heuristics

rfwatson:
Hi there,

This patch adds some basic test coverage for the above function.

It also includes a tiny change adding an `#[allow(dead_code)]` to kill a compilation warning in the ALSA module.

### DIFF
--- a/src/host/alsa/mod.rs
+++ b/src/host/alsa/mod.rs
@@ -9,11 +9,11 @@ use crate::{
     SupportedBufferSize, SupportedStreamConfig, SupportedStreamConfigRange,
     SupportedStreamConfigsError,
 };
+use std::cmp;
 use std::convert::TryInto;
 use std::sync::Arc;
 use std::thread::{self, JoinHandle};
 use std::vec::IntoIter as VecIntoIter;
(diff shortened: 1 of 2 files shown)


src/host/alsa/mod.rs
```diff
@@ -436,6 +436,7 @@ struct StreamInner {
     // Minimum number of samples to put in the buffer.
     period_len: usize,
 
+    #[allow(dead_code)]
     // Whether or not the hardware supports pausing the stream.
     can_pause: bool,
```
mitchmindtree, line 452: Could you possibly add a `// TODO: We need an API to expose this. See #197, #284.`

rfwatson:
Done! Thanks.

mitchmindtree:
Ahhhh #401 just landed today - it looks like this might need to be updated to account for it?

rfwatson:
> Ahhhh #401 just landed today - it looks like this might need to be updated to account for it?

Good shout. I'll take a look later.

rfwatson:
Should be passing again now :slightly_smiling_face:

PR #441, opened 2020-07-04
Retry COM initialization

HEnquist:
I noticed that running the enumerate example with ASIO enabled listed ASIO devices fine, and then panic:ed on Wasapi. Reversing the order to do Wasapi first made it work fine.
I'm thinking that the same situation could happen in an application that supports switching between different hosts.
This PR adds a second try to initialize COM if the first try fails. This fixes the enumerate example, but I guess there could be side effects I'm not aware of.

### DIFF
--- a/src/host/wasapi/com.rs
+++ b/src/host/wasapi/com.rs
@@ -10,7 +10,11 @@ thread_local!(static COM_INITIALIZED: ComInitialized = {
     unsafe {
         // this call can fail if another library initialized COM in single-threaded mode
         // handling this situation properly would make the API more annoying, so we just don't care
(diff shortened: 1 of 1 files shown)


mitchmindtree:
Thanks for the PR @HEnquist!

Unfortunately I'm also not savvy enough with the Windows platform to know what other implications this might have. If anyone on Windows would like to test this or offer any thoughts/input that would be greatly appreciated!

HEnquist:
Yes there might be some problem with this approach. This change only comes into play when there would otherwise be a panic, and my thinking was that it might be better to risk having some problem in the future than a exiting with a panic right now.

HEnquist:
Soon a year.. I'll close it now.

PR #455, opened 2020-07-22
Release asio-sys 0.2.0

### LINKED ISSUE #450: [asio] Asio does not work in Cpal 0.12.0 from crates.io
There seems to be an issue with the packaging of the asio-sys crate. When building a project that uses the Cpal crate from crates.io, it also downloads a quite outdated version of asio-sys from crates.io. It seems the version that is included in the cpal repository doesn't get included in the cpal crate. The result is that the build fails because the old asio-sys is missing things.
Shouldn't the asio-sys crate be broken out to a separate repo, with a proper version number?

est31:
Fixes #450


HEnquist:
Wouldnt it be better to bump to 0.2.0? Asio-sys version 0.1.0 really doesn't work with cpal 0.12.0, and I'm also guessing cpal 0.11.0 doesn't work with this new asio-sys (but haven't checked). 
Then this needs to be changed here too: https://github.com/RustAudio/cpal/blob/master/Cargo.toml#L24
That would of course also need a new release of cpal, 0.12.1 perhaps?

est31:
Hmm yeah good point

PR #458, opened 2020-07-27
Use the given sample rate in `SupportedStreamConfigRange::with_sample_rate()`

richardmitic:
`with_sample_rate` previously returned the same as `with_max_sample_rate`, which seems like a typo.
Additionally, ALSA reports that some soundcards have a maximum sample rate of `INT_MAX` in which case there will be an error when opening the stream.


mitchmindtree:
Thanks for the PR! I came across #462 and merged it before coming across this one, my apologies! Closing now that #462 is merged.

PR #464, opened 2020-08-02
Fix broken Stream concept link in the docs.

ely-uf:
Affects the "How to use cpal" section on the main page of the docs.

Point the `Stream` link to the `struct Stream` following the example
of the `Device` and `Host` links from the same section.


ely-uf:
Current documentation page: https://docs.rs/cpal/0.12.1/cpal/
The broken link: https://docs.rs/cpal/0.12.1/cpal/trait.Stream.html

mitchmindtree:
Thanks!

PR #468, opened 2020-08-23
Android support via Oboe

endragor:
This code compiles (assuming my PR into `oboe-rs` gets merged), but has not been tested yet, because there are problems with including multiple .so in a single .apk (https://github.com/katyo/oboe-rs/issues/5). In the meantime I'd like to ask for clarifications for a couple of questions that occurred during implementation:

1. `StreamInstant` for callbacks.

It's unclear whether CPAL guarantees any relations between different `StreamInstant` values. For example, should `callback` and `capture` fields of `InputStreamTimestamp` be comparable in any way? Should `InputStreamTimestamp` values for different streams be comparable?

Oboe provides only the number of frames read/written from/to the stream. Based on the number of frames and sample rate it's possible to compute the duration. This is how I provide values for `capture` and `playback` fields of `InputStreamTimestamp` and `OutputStreamTimestamp`. For `callback` fields I simply use `std::time::Instant`, providing the time elapsed since the stream creation. Is that OK?

2. There seems to be no good way to query maximum buffer size from Android, but methods for querying minimum buffer size exist. So I decided to return `i32::MAX` as the maximum buffer size. Providing this value should theoretically work, but the implementation may reduce the value to something more meaningful. Again, is this fine or is there a better approach?

### DIFF
--- a/.github/workflows/cpal.yml
+++ b/.github/workflows/cpal.yml
@@ -18,11 +18,17 @@ jobs:
         toolchain: stable
         override: true
         components: clippy
+        target: armv7-linux-androideabi
     - name: Run clippy
       uses: actions-rs/clippy-check@v1
       with:
         token: ${{ secrets.GITHUB_TOKEN }}
         args: --all --all-features
+    - name: Run clippy for Android target
+      uses: actions-rs/clippy-check@v1
+      with:
+        token: ${{ secrets.GITHUB_TOKEN }}
+        args: --all --all-features --target armv7-linux-androideabi
 
   rustfmt-check:
     runs-on: ubuntu-latest
@@ -177,3 +183,23 @@ jobs:
       run: cargo test --all --no-default-features --verbose
     - name: Run all features
       run: cargo test --all --all-features --verbose
+
+  android-check:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Install stable
+      uses: actions-rs/toolchain@v1
+      with:
+        profile: minimal
+        toolchain: stable
+        override: true
+        target: armv7-linux-androideabi
+    - name: Check beep
+      run: cargo check --example beep --target armv7-linux-androideabi --verbose
+    - name: Check enumerate
+      run: cargo check --example enumerate --target armv7-linux-androideabi --verbose
+    - name: Check feedback
+      run: cargo check --example feedback --target armv7-linux-androideabi --verbose
+    - name: Check record_wav
+      run: cargo check --example record_wav --target armv7-linux-androideabi --verbose
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,3 +43,9 @@ stdweb = { version = "0.1.3", default-features = false }
 wasm-bindgen = { version = "0.2.58", optional = true }
 js-sys = { version = "0.3.35" }
 web-sys = { version = "0.3.35", features = [ "AudioContext", "AudioContextOptions", "AudioBuffer", "AudioBufferSourceNode", "AudioNode",  "AudioDestinationNode", "Window", "AudioContextState"] }
+
+[target.'cfg(target_os = "android")'.dependencies]
+oboe = { version = "0.2.1", features = [ "java-interface" ] }
+ndk = "0.2"
+ndk-glue = "0.2"
+jni = "0.17"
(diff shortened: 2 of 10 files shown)


enfipy:
What status of Android support here? Why this PR closed?

endragor:
Still need to figure out a way to include oboe library to the .apk builds that depend on it.

zarik5:
Is there any update on this?

enfipy:
Any updates on this? I think this will be very useful in the nearest future.

endragor:
This should in theory work and be ready for merge. I tested it in the past on Android 10 with the older version of `oboe-rs`. I'd appreciate if someone could perform more testing and see if this works on older Android version - ideally before Android 8.0, where AAudio was introduced.

enfipy:
@endragor I tried to add this to Bevy and I was able to build an app for Android. But when I add this system:
```rust
fn audio(asset_server: Res<AssetServer>, audio_output: Res<AudioOutput>) {
    let music = asset_server.load("sounds/Windless Slopes.mp3").unwrap();
    audio_output.play(music);
}
```

It breaks after rendering some Bevy graphics with the `ndk_glue` error here:
```rust
pub fn native_activity() -> &'static NativeActivity {
    unsafe { NATIVE_ACTIVITY.as_ref().unwrap() }
}
```

I'm not sure but maybe something wrong with [get_activity](https://github.com/katyo/oboe-rs/search?q=get_activity) or where it gets called. Any ideas?

mitchmindtree:
Hey @endragor, thanks for such an epic PR!

As I don't have an Android device available to me, it's difficult to test this out locally. It would be great to have at list once other contributor be able to confirm this is running before landing this.

Also, it would be great if a CI job could be added for building and testing this backend (see `.github/workflows/cpal.yml`). This would greatly assist maintainers when they don't have a device to test on locally, and would help to reduce the chance of accidentally introducing errors. CI scripts can also act as a nice reference for maintainers or other users to see how they can setup an environment in which they can build for the desired target.

enfipy:
I'm trying to run it on my Mi Mix 2 and getting pretty weird errors. First, I wasn't able to proceed without updating `ndk` & `ndk-glue` in `cpal` to version 0.2 (was getting an error with `native_activity()`). But after I updated - I receive the following 2 kinds of errors randomly!

Screen Shot 2020-10-21 at 11 04 57 PM

I don't know when exactly errors appear but I'm sure that it panics during playback!

1. `FORTIFY: pthread_mutex_lock called on a destroyed mutex (0x77f01fe4d0)`
2. `thread '<unnamed>' panicked at 'supplied instant is later than self', library/std/src/time.rs:285:48`

Any ideas?

enfipy:
I placed my logs in Pastebin analog [here](https://0bin.net/paste/DH7lOKnu#7F0cvVJHjuUrhtgvbYp6zCtD0OLDxlvose98yWIG2kW).

I tried to clear them from gfx & bevy spam logs and it clearly shows what happens. But I'm not sure how to fix it.

endragor:
@enfipy I'm planning to do some testing on the weekend. At a glance the stack trace looks like `rodio` stream handle is being used when the stream itself was destroyed. So that is more about correct `rodio` usage; things look fine on `CPAL` side - the internal stream was opened and started successfully.

endragor:
`ndk` and `ndk-glue` versions indeed needed bumping, but otherwise this branch works for me.

I've added a repo for others to quickly test Android support: https://github.com/endragor/cpal-android-test
Here is a direct link to the prebuilt APK that you can share with others so they can test it: https://github.com/endragor/cpal-android-test/raw/master/CPALTest.apk

I've tested it on Honor 10i running Android 10. It'd be great to test on older Android versions. If Android emulator works for you (it does not for me), you may test it there, too.

@mitchmindtree I've added CI jobs to check clippy errors and compilation of examples for Android target. We can't actually run anything on an Android device from CI, so I don't think there is much else to do there. By the way, I noticed there are clippy warnings (unrelated to this PR) and yet the clippy job finishes successfully. Is that expected?

enfipy (approved):
Works amazingly! Superb work!

enfipy:
@est31 Hey, may you consider to review this one so we can proceed with this?

mitchmindtree:
Awesome stuff @endragor, looking good to me! And thanks @enfipy for testing.

Maybe it would also be useful to have a `cargo apk build` step in the CI too, just to check that all the build time linking and packaging stays in a working state too? I think it's fine for this to be added in a follow-up PR though.

Happy to merge!

PR #469, opened 2020-08-30
Fix compilation error when using null host

exrook:
Also remove other references to EventLoop

### DIFF
--- a/.github/workflows/cpal.yml
+++ b/.github/workflows/cpal.yml
@@ -127,6 +127,25 @@ jobs:
       - name: Build beep example
         run: cargo build --example beep --target ${{ matrix.target }} --features=wasm-bindgen
 
+  wasm32-wasi-test:
+
+    strategy:
+      matrix:
+        target: [wasm32-wasi]
+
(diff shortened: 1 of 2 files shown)


est31:
I wonder if it would be possible to set up CI for the null host? Only to do a cargo check whether it builds.

exrook:
If each backend was a default feature flag, then compiling without default features would accomplish that (which it looks like the current CI already does). Otherwise I think we would have to have CI cross compile to a platform with no backends yet. (That's how I ran into this error, I'm looking into adding an Android backend)

est31:
> I think we would have to have CI cross compile to a platform with no backends yet.

Sounds good! Any ideas for a platform that has an easy to set up cross compilation toolchain?

exrook:
I think `wasm32-wasi` should work well, it doesn't pull in any of the existing backends, and the tests already run for the other wasm targets.

est31:
@exrook could you add `wasm-wasi` CI code to this PR? That would be neat!

est31 (approved):
Thanks!

PR #470, opened 2020-08-31
Handle ALSA overruns/underruns by resetting the PCM device

exrook:
This fixes an issue I was running into where when an underrun happened on the output buffer, the output stream worker would get stuck in a loop using up 100% cpu polling the descriptors and audio playback would stop.


est31 (approved):
Thanks!

PR #474, opened 2020-09-10
webaudio: Fix default_output_config

rfwatson:
Fix the behaviour of wasm `default_output_device()` based on the documentation [here](https://github.com/RustAudio/cpal/blob/master/src/lib.rs#L570), and in-browser testing

### DIFF
--- a/src/host/emscripten/mod.rs
+++ b/src/host/emscripten/mod.rs
@@ -107,7 +107,7 @@ impl Device {
         configs.sort_by(|a, b| a.cmp_default_heuristics(b));
         let config = configs
             .into_iter()
-            .next()
+            .last()
             .expect(EXPECT)
             .with_sample_rate(DEFAULT_SAMPLE_RATE);
         Ok(config)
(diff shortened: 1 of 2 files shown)


est31:
Good catch!

It's indeed the case that the configs with the highest priority are ordered *last*. https://github.com/RustAudio/cpal/blob/bfffbb524103fd8c6b81cdd588f28595d811df72/src/lib.rs#L637

I've found the same issue in the emscripten code which also uses the first element.

Furthermore, maybe it would be helpful to use the `max_by` iterator combinator? Then we don't need to store all configs into a vec and sort them (sorting is `o(n log n)` while finding maximum is `o(n)`).

Do you want to adjust the code in the webaudio and emscripten backends to do that?

rfwatson:
Good idea, updated!

As a separate issue, do you think there's any scope to improve the readability of the [`cmp_default_heuristics`](https://github.com/rfwatson/cpal/blob/784489dd61f39150eaf2f2e6d988b442ce1f1869/src/lib.rs#L589) function itself? This isn't the first time I've struggled to understand its inner workings (although I'm fairly new to Rust which probably doesn't help)

est31 (approved):
Thanks!

Issue #476, opened 2020-09-12
Would it be possible to re-enable WASAPI loopback?


dheijl:
Contrary to the current documentation in wasapi/mod.rs the loopback facility is **not** present :
```rust
/// The WASAPI host, the default windows host type.
///
/// Note: If you use a WASAPI output device as an input device it will
/// transparently enable loopback mode (see
/// https://docs.microsoft.com/en-us/windows/win32/coreaudio/loopback-recording).
```

It was added in PR #339 []((https://github.com/RustAudio/cpal/pull/339)) but seems to have been removed.
Would it be possible to please add this again in device.rs? I suppose it got lost when code was moved from stream.rs to device.rs? 
I can provide a PR if you prefer.
```diff
@@ -33,8 +33,7 @@ use super::winapi::um::audioclient::{
    self, IAudioClient, IID_IAudioClient, AUDCLNT_E_DEVICE_INVALIDATED,
};
use super::winapi::um::audiosessiontypes::{
-    AUDCLNT_SHAREMODE_SHARED, AUDCLNT_STREAMFLAGS_EVENTCALLBACK,
-};
+    AUDCLNT_SHAREMODE_SHARED, AUDCLNT_STREAMFLAGS_EVENTCALLBACK, AUDCLNT_STREAMFLAGS_LOOPBACK
+};
use super::winapi::um::combaseapi::{
    CoCreateInstance, CoTaskMemFree, PropVariantClear, CLSCTX_ALL,
};
@ -657,6 +656,13 @@ impl Device {
                BufferSize::Default => (),
            };

+           let mut stream_flags: DWORD = AUDCLNT_STREAMFLAGS_EVENTCALLBACK;
+
+           if self.data_flow() == eRender {
+               stream_flags |= AUDCLNT_STREAMFLAGS_LOOPBACK;
+           }

            // Computing the format and initializing the device.	
            // Computing the format and initializing the device.
            let waveformatex = {
                let format_attempt = config_to_waveformatextensible(config, sample_format)
@ -673,7 +679,7 @@ impl Device {
                // finally initializing the audio client
                let hresult = (*audio_client).Initialize(
                    share_mode,
-                   AUDCLNT_STREAMFLAGS_EVENTCALLBACK,
+                   stream_flags,
                    0,
                    0,
                    &format_attempt.Format,```

nyanpasu64:
I just downloaded cpal, and was trying to use WASAPI loopback output. I'm looking forward to the pull request.

[...]

PR #477, opened 2020-09-20
Re-enable WasApi loopback capture

dheijl:
See issue https://github.com/RustAudio/cpal/issues/476.

This is essentially PR #339  again which was merged but was later (accidentally?) removed again.

It allows loopback capture from a WasApi output device.


dheijl:
will corrrect format error first

PR #483, opened 2020-10-04
SampleRate: Implement std::ops::{Div, Mul}

### LINKED ISSUE #482: SampleRate could implement core::ops::{Div, Mul}
Currently, it is somewhat inconvenient to multiply/divide a `SampleRate`, as it involves something like `SampleRate(k * rate.0)`; it would make sense to implement the `core::ops::{Div, Mul}` traits to allow multiplication and division by integers.

Those operations are often useful when resampling (I actually ran into this when working on `rodio`'s resampler), or when computing the maximum frequency that can be reproduced (the Nyquist-Shannon sampling theorem states that's half the sample rate), etc...

nbraud:
Closes #482

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -159,6 +159,7 @@ pub use platform::{
 };
 pub use samples_formats::{Sample, SampleFormat};
 use std::convert::TryInto;
+use std::ops::{Div, Mul};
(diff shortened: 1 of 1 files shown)


est31 (approved):
Thanks!

PR #493, opened 2020-10-30
Add support for OpenBSD platform via sndio host

AaronM04:
This adds support for the [sndio](https://sndio.org/) host which is the default and only supported host on OpenBSD. I have tested all the examples for cpal and rodio, with no issues observed. If there are any improvements needed to bring this PR up to the standards of the cpal project, please let me know. :-)


mitchmindtree:
Thanks for the epic PR @AaronM04!

I don't have time to dig into this myself at the moment, not to mention my lack of experience with sndio & OpenBSD, however two steps that would help to land this are:

1. Having one or more other OpenBSD users review and test the examples. I don't know of any other OpenBSD users personally, but if any are reading this it would help a lot if you could test and review!
2. Adding a job to the Github CI workflow for OpenBSD + sndio. I'm not sure off the top of my head if it's possible to spin up a OpenBSD instance as easily as it is say Ubuntu. If it turns out not to be feasible with the github workflow, perhaps you have another recommendation for OpenBSD-specific CI? That said, I'd prefer we get this working with the github workflow in order to keep managing CI as simple as possible for maintainers.

src/host/sndio/adapters.rs
```diff
@@ -0,0 +1,100 @@
...
+
+    Box::new(move |data: &Data, info: &InputCallbackInfo| {
+        // Note: we construct adapted_data here instead of in the parent function because buf needs
+        // to be owned by the closure.
+        let mut adapted_data =
+            unsafe { Data::from_parts(buf.as_mut_ptr() as *mut _, buffer_size, sample_format) };
+        let data_slice: &[i16] = data.as_slice().unwrap(); // unwrap OK because data is always i16
+        match sample_format {
+            SampleFormat::F32 => {
+                let adapted_slice: &mut [f32] = adapted_data.as_slice_mut().unwrap(); // unwrap OK because of the match
```
blackgnezdo: This may be outside the scope of your patch, but it feels wrong that adapted_data requires an unwrap when match above uses the same sample_format that was just baked into adapted_data.

src/host/sndio/adapters.rs
```diff
@@ -0,0 +1,100 @@
...
+                }
+            }
+            SampleFormat::U16 => {
+                let adapted_slice: &mut [u16] = adapted_data.as_slice_mut().unwrap(); // unwrap OK because of the match
+                assert_eq!(data_slice.len(), adapted_slice.len());
+                for (i, adapted_ref) in adapted_slice.iter_mut().enumerate() {
+                    *adapted_ref = data_slice[i].to_u16();
+                }
+            }
+            SampleFormat::I16 => {
```
blackgnezdo: What stops the code from changing to be a top-level match which covers 3 different cases exactly once?
This unreachable branch is only unreachable because L14 above covered it. It would seem the two other branches should be possible to move into a couple distinct Box's that potentially can use a common type-parameterized function.

src/host/sndio/endian.rs
```diff
@@ -0,0 +1,14 @@
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub(super) enum Endian {
+    BE,
+    LE,
+}
+
+pub(super) fn get_endianness() -> Endian {
```
blackgnezdo: There's a config to determine this at build time instead.
https://doc.rust-lang.org/reference/conditional-compilation.html#target_endian

src/host/sndio/mod.rs
```diff
@@ -0,0 +1,901 @@
...
+            SndioError::DeviceNotAvailable => SupportedStreamConfigsError::DeviceNotAvailable,
+            SndioError::BackendSpecific(bse) => {
+                SupportedStreamConfigsError::BackendSpecific { err: bse }
+            }
+        }
+    }
+}
+
+pub struct Devices {
+    returned: bool,
```
blackgnezdo: This struct (and the field) could use a comment.

src/host/sndio/mod.rs
```diff
@@ -0,0 +1,901 @@
...
+}
+
+impl Iterator for Devices {
+    type Item = Device;
+    fn next(&mut self) -> Option<Device> {
```
blackgnezdo: This looks like a singleton iterator. Can you cook something up with a vec of one element instead of building a separate gizmo?

src/host/sndio/mod.rs
```diff
@@ -0,0 +1,901 @@
...
+impl Devices {
+    fn new() -> Devices {
+        Devices { returned: false }
+    }
+}
+
+/// The shared state between Device and Stream. Responsible for closing handle when dropped.
+struct InnerState {
+    /// If device has been open with sio_open, contains a handle. Note that even though this is a
+    /// pointer type and so doesn't follow Rust's borrowing rules, we should be careful not to copy
```
blackgnezdo: It is not immediately obvious why it has to be a pointer rather than a reference. Two layers of optionality aren't immediately intuitive. If the construct is required for some reason, it's worth documenting such unusual choice.

src/host/sndio/mod.rs
```diff
@@ -0,0 +1,901 @@
...
+    /// it out because that may render Mutex<InnerState> ineffective in enforcing exclusive access.
+    hdl: Option<*mut sndio_sys::sio_hdl>,
+
+    /// Buffer overrun/underrun behavior -- ignore/sync/error?
+    behavior: BufferXrunBehavior,
+
+    /// If a buffer size was chosen, contains that value.
```
blackgnezdo: The abundance of Optional suggests a huge state space. I suspect it can be reduced by refactoring `InnerState` type into an enum with a pruned set of fields that apply to specific stages of of the state machine.

src/host/sndio/mod.rs
```diff
@@ -0,0 +1,901 @@
...
+    data_callback: Box<dyn FnMut(&Data, &InputCallbackInfo) + Send + 'static>,
+    error_callback: Box<dyn FnMut(StreamError) + Send + 'static>,
+}
+
+struct OutputCallbacks {
+    data_callback: Box<dyn FnMut(&mut Data, &OutputCallbackInfo) + Send + 'static>,
+    error_callback: Box<dyn FnMut(StreamError) + Send + 'static>,
+}
+
+unsafe impl Send for InnerState {}
```
blackgnezdo: This looks very scary. Could you add some comments explaining why this type is safe to Send? Maybe the type should be wrapped into an Arc instead?

src/host/sndio/mod.rs
```diff
@@ -0,0 +1,901 @@
...
+    fn open(&mut self) -> Result<(), SndioError> {
+        if self.hdl.is_some() {
+            // Already open
+            return Ok(());
+        }
+
+        let hdl = unsafe {
+            // The transmute is needed because this C string is *const u8 in one place but *const i8 in another place.
+            let devany_ptr = mem::transmute::<_, *const i8>(sndio_sys::SIO_DEVANY as *const _);
+            let nonblocking = true as i32;
```
blackgnezdo: This might as we be written as 1 with an increase in readability, what do you think?

src/host/sndio/mod.rs
```diff
@@ -0,0 +1,901 @@
...
+            sndio_sys::sio_open(
+                devany_ptr,
+                sndio_sys::SIO_PLAY | sndio_sys::SIO_REC,
+                nonblocking,
+            )
+        };
+        if hdl.is_null() {
+            return Err(SndioError::DeviceNotAvailable);
```
blackgnezdo: Can you do better and somehow report the errno? Probably applies everywhere.

src/host/sndio/mod.rs
```diff
@@ -0,0 +1,901 @@
...
+        if inner_state.sample_rate_to_par.is_none() {
+            inner_state.open()?;
+        }
+
+        if inner_state.sample_rate_to_par.is_none() {
+            return Err(backend_specific_error("no sample rate map!").into());
+        }
+
+        let mut config_ranges = vec![];
+        // unwrap OK because of the check at the top of this function.
```
blackgnezdo: Why should it not be combined with the above to avoid the unwrap?
```
if Some(r) = inner_state.sample_rate_to_par {
  // for loop
} else {
  return Err(backend_specific_error("no sample rate map!").into());
}
```
Since it applies in many cases below, it should be factored into a function that takes a continuation for the `Ok` branch. Something like `with_open_sample_rate`.

blackgnezdo (commented):
Responding to [review request](https://marc.info/?l=openbsd-tech&m=160428302106558&w=2) since I happen to care about OpenBSD and Rust :)

The change is fairly massive so I didn't make it all the way to the end. Depending on your motivation and the prevailing conventions of RustAudio, some comments may be too nit-picky.

blackgnezdo:
> 2. Adding a job to the Github CI workflow for OpenBSD + sndio. I'm not sure off the top of my head if it's possible to spin up a OpenBSD instance as easily as it is say Ubuntu. If it turns out not to be feasible with the github workflow, perhaps you have another recommendation for OpenBSD-specific CI? That said, I'd prefer we get this working with the github workflow in order to keep managing CI as simple as possible for maintainers.

I don't know what it would take to get an OpenBSD CI on github, but as a point of reference, here's what it takes to get a GCE image of OpenBSD.
https://github.com/google/syzkaller/blob/master/tools/create-openbsd-gce-ci.sh

src/host/sndio/mod.rs
AaronM04: I can try that here but when I checked `errno` for an error from `sio_setpar` it didn't set it.

src/host/sndio/mod.rs
AaronM04: I see your point -- scary indeed! IIRC without this, `InnerState` would be `!Send` because it contains a `*mut sndio_sys::sio_hdl` pointer.  I'll see if it'll compile with `unsafe impl Send for Arc<Mutex<InnerState>> {}` instead.

src/host/sndio/mod.rs
AaronM04: Maybe. I tend to favor boolean values as Rust `bool`s, though the cast may be too confusing here.

grayed:
> 2. Adding a job to the Github CI workflow for OpenBSD + sndio. I'm not sure off the top of my head if it's possible to spin up a OpenBSD instance as easily as it is say Ubuntu. If it turns out not to be feasible with the github workflow, perhaps you have another recommendation for OpenBSD-specific CI? That said, I'd prefer we get this working with the github workflow in order to keep managing CI as simple as possible for maintainers.

FWIW, the sndio library is available on Linux, so you don't need OpenBSD to run CI for it: https://sndio.org/

AaronM04:
For CI, I think it will be more practical to use sndio on Linux (as @grayed mentioned) than to setup an OpenBSD VM as part of the github workflow. The one potential complication is getting `sndio-sys` as a dependency on linux, but only conditionally. Currently the dependency is conditional on `cfg(target_os = "openbsd")`. I am not very familiar with this part of Cargo/Rust so I would be very grateful for any assistance with this.

As for testing the workflow, should I just push to this branch? (it could be a lot of commits if it's trial and error that way)

src/host/sndio/adapters.rs
AaronM04: I did move it into distinct Boxes to get rid of the unreachable and some unwraps. I didn't do the type-parameterized function because it would only be used in two places and it didn't seem like that level of DRY was warranted here...

Let me know if you think the new version is better.

src/host/sndio/mod.rs
AaronM04: I'll address this with the `InnerState` refactoring.

mitchmindtree:
> The one potential complication is getting sndio-sys as a dependency on linux, but only conditionally.

Perhaps we can add a `linux-sndio` cargo feature, and then conditionally include the sndio host with a `cfg` that looks something like:

```rust
#[cfg(or(target_os = "openbsd", and(target_os = "linux", feature = "linux-sndio")))]
```

@RustAudio/cpal-maintainers does this sound OK?

> As for testing the workflow, should I just push to this branch? (it could be a lot of commits if it's trial and error that way)

Yeah that's totally fine, though perhaps rather than continuously adding new commits you could use `git commit --amend` on the last commit that adds the workflow and then force push it on each attempt? This is what I normally do when attempting to add a new workflow at least, not sure if there's a nicer way!

src/host/sndio/mod.rs
AaronM04: The gizmo has been vanquished!

src/host/sndio/mod.rs
AaronM04: I have replaced the `unsafe impl Send for InnerState` with an `unsafe impl Send for SioHdl`, and added `SioHdl` as a wrapper on `*mut sndio_sys::sio_hdl`.

This seems better to me in that one of the other parts of `InnerState` could be `!Send` and this won't mask the problem resulting in unsound code.

src/host/sndio/mod.rs
AaronM04: @blackgnezdo I have fixed this. Please take a look when you get a chance. Thanks!

src/host/sndio/mod.rs
```diff
@@ -0,0 +1,1077 @@
...
+impl InnerState {
+    fn new() -> Self {
+        InnerState::Init {
+            behavior: BufferXrunBehavior::Sync,
+        }
+    }
+
+    fn open(&mut self) -> Result<(), SndioError> {
+        match self {
+            InnerState::Opened { .. } | InnerState::Running { .. } => Ok(()), // Already opened
```
blackgnezdo: Why is it a good idea to permit a duplicate open? I'd complain and return an SndioError.

src/host/sndio/mod.rs
```diff
@@ -0,0 +1,1077 @@
...
+                    ))
+                    .into());
+                }
+
+                let buffer_size = determine_buffer_size(&config.buffer_size, par.round, None)?;
+
+                // Transition to running
+                par.appbufsz = buffer_size as u32;
+                hdl.set_params(&par)?;
+                let mut tmp: HashMap<u32, sndio_sys::sio_par> = HashMap::new();
```
blackgnezdo: Does turbo-fish not work here? It should be shorter.

src/host/sndio/mod.rs
```diff
@@ -0,0 +1,1077 @@
...
+        behavior: BufferXrunBehavior,
+    },
+    Opened {
+        /// Contains a handle returned from sio_open. Note that even though this is a pointer type
+        /// and so doesn't follow Rust's borrowing rules, we should be careful not to copy it out
+        /// because that may render Mutex<InnerState> ineffective in enforcing exclusive access.
+        hdl: SioHdl,
+
+        /// Map of sample rate to parameters.
+        sample_rate_to_par: HashMap<u32, sndio_sys::sio_par>,
```
blackgnezdo: `HashMap<u32, sndio_sys::sio_par>` type seems pervasive. It probably deserves a well-named alias or even a newtype. I'm also partial to creating aliases for numeric types such that a sample rate and every other kind of u32 aren't easy to mix-up. Between these two, I'd probably have here:
```
rate_parameters: RateParameterMap,
```

src/host/sndio/mod.rs
```diff
@@ -0,0 +1,1077 @@
...
+            return Err(backend_specific_error("failed to set parameters with sio_setpar").into());
+        }
+        Ok(())
+    }
+}
+
+unsafe impl Send for SioHdl {}
+
+/// The shared state between Device and Stream. Responsible for closing handle when dropped.
+enum InnerState {
```
blackgnezdo, line 317: This is so much more descriptive than a bunch of Options!

src/host/sndio/mod.rs
```diff
@@ -0,0 +1,1077 @@
...
+        if status != 1 {
+            return Err(backend_specific_error("failed to set parameters with sio_setpar").into());
+        }
+        Ok(())
+    }
+}
+
+unsafe impl Send for SioHdl {}
+
+/// The shared state between Device and Stream. Responsible for closing handle when dropped.
```
blackgnezdo: Maybe this should also list the permitted transitions unless they are fully controlled by the type impl?

src/host/sndio/mod.rs
```diff
@@ -0,0 +1,1077 @@
...
+}
+
+impl Iterator for Devices {
+    type Item = Device;
+    fn next(&mut self) -> Option<Self::Item> {
+        self.0.take()
+    }
+}
+
+struct SioHdl(*mut sndio_sys::sio_hdl);
```
blackgnezdo, line 116: This probably deserves a comment.

src/host/sndio/mod.rs
```diff
@@ -0,0 +1,1077 @@
...
+            sndio_sys::sio_setpar(self.0, &mut newpar as *mut _)
+        };
+        if status != 1 {
+            return Err(backend_specific_error("failed to set parameters with sio_setpar").into());
+        }
+        Ok(())
+    }
+}
+
+unsafe impl Send for SioHdl {}
```
blackgnezdo, line 246: Probably deserves a comment explaining `unsafe`: why it has to be unsafe, also why and under which assumptions it's OK.

src/host/sndio/mod.rs
```diff
@@ -0,0 +1,1077 @@
...
+        D: FnMut(&Data, &InputCallbackInfo) + Send + 'static,
+        E: FnMut(StreamError) + Send + 'static,
+    {
+        let inner_state_arc = self.inner_state.clone();
+
+        let mut inner_state = self.inner_state.lock().unwrap();
+
+        inner_state.setup_stream(config)?;
+
+        let idx;
```
blackgnezdo, line 848: Combine declaration with assignment.

src/host/sndio/mod.rs
```diff
@@ -0,0 +1,1077 @@
...
+
+impl Device {
+    pub fn new() -> Self {
+        Device {
+            inner_state: Arc::new(Mutex::new(InnerState::new())),
+        }
+    }
+
+    pub fn set_xrun_behavior(&mut self, b: BufferXrunBehavior) -> Result<(), SndioError> {
+        let mut inner_state = self.inner_state.lock().unwrap();
```
blackgnezdo: Since there's a way to return an error, why unwrap rather than complain? Seems to apply in all the cases below.

src/host/sndio/mod.rs
```diff
@@ -0,0 +1,1077 @@
...
+        drop(inner_state); // Unlock
+        Ok(Stream {
+            inner_state: self.inner_state.clone(),
+            is_output: false,
+            index: idx,
+        })
+    }
+
+    /// Create an output stream.
+    fn build_output_stream_raw<D, E>(
```
blackgnezdo, line 896: Looks like 50 lines of code largely similar to `build_input_stream_raw`, any chance you can figure out the common parts and de-dup?

src/host/sndio/mod.rs
```diff
@@ -0,0 +1,1077 @@
...
+            inner_state: self.inner_state.clone(),
+            is_output: true,
+            index: idx,
+        })
+    }
+}
+
+fn supported_config_from_par(par: &sndio_sys::sio_par, num_channels: u32) -> SupportedStreamConfig {
+    SupportedStreamConfig {
+        channels: num_channels as u16,
```
blackgnezdo: Would this be a true comment to add? Otherwise it's not clear why u32 suddenly becomes u16.
// Conversion is courtesy of type mismatch between sndio and RustAudio.

blackgnezdo (commented):
I noted a few more things, but the whole change is too big to read in one sitting :(

Also, there's a bunch of CI generated complaints that probably need to be addressed.

src/host/sndio/mod.rs
AaronM04: I did it this way because only panics on the previous thread to have locked the mutex can cause an error to be returned from lock. I suppose it still makes sense to return a "thread panicked" error here.

src/host/sndio/mod.rs
AaronM04: I spent a lot of time trying this out and unfortunately I don't think this can be de-duped without adding another layer of `Box` around the callbacks, which would impact performance (albeit not by much).

src/host/sndio/mod.rs
AaronM04: Yes, added verbatim!

src/host/sndio/mod.rs
AaronM04: (well, there are macros but that feels almost like obfuscating the code to me)

AaronM04:
@blackgnezdo I finally was able to return to this. Please review when you have time. Thanks in advance! :smile: 

BTW I don't know what CI complaints you are referring to.

src/host/sndio/mod.rs
```diff
@@ -0,0 +1,1144 @@
...
+                // Nothing to do -- device is not open.
+                Ok(())
+            }
+        }
+    }
+
+    // TODO: make these 4 methods generic (new CallbackSet<T> where T is either InputCallbacks or OutputCallbacks)
+    /// Puts the supplied callbacks into the vector in the first free position, or at the end. The
+    /// index of insertion is returned.
+    fn add_output_callbacks(&mut self, callbacks: OutputCallbacks) -> Result<usize, SndioError> {
```
blackgnezdo, line 462: How important is the choice of Vec as callbacks storage? I see a significant code reduction if a pair `(usize, Hash<usize,T>)` were to be as for storage with usize getting unconditionally incremented on each insertion.

blackgnezdo (commented):
@AaronM04, I wanted to play with the code but `cargo build` fails for me on openbsd-amd64-current:
```
   Compiling sndio-sys v0.0.1
error: failed to run custom build command for `sndio-sys v0.0.1`

Caused by:
  process didn't exit successfully: `/home/greg/s/cpal/target/debug/build/sndio-sys-c905c01512cca331/build-script-build` (exit code: 101)
  --- stdout
  cargo:rustc-link-lib=sndio
  cargo:rerun-if-changed=wrapper.h

  --- stderr
  thread 'main' panicked at 'Unable to find libclang: "couldn\'t find any valid shared libraries matching: [\'libclang.so\', \'libclang.so.*\'], set the `LIBCLANG_PATH` environment variable to a path where one of these files can be found (invalid: [])"', /home/greg/.cargo/registry/src/github.com-1ecc6299db9ec823/bindgen-0.53.3/src/lib.rs:1956:31
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
Any clues?

AaronM04:
Somehow I missed the notification for this. :/

> @AaronM04, I wanted to play with the code but `cargo build` fails for me on openbsd-amd64-current:
> 
> ```
>    Compiling sndio-sys v0.0.1
> error: failed to run custom build command for `sndio-sys v0.0.1`
> 
> Caused by:
>   process didn't exit successfully: `/home/greg/s/cpal/target/debug/build/sndio-sys-c905c01512cca331/build-script-build` (exit code: 101)
>   --- stdout
>   cargo:rustc-link-lib=sndio
>   cargo:rerun-if-changed=wrapper.h
> 
>   --- stderr
>   thread 'main' panicked at 'Unable to find libclang: "couldn\'t find any valid shared libraries matching: [\'libclang.so\', \'libclang.so.*\'], set the `LIBCLANG_PATH` environment variable to a path where one of these files can be found (invalid: [])"', /home/greg/.cargo/registry/src/github.com-1ecc6299db9ec823/bindgen-0.53.3/src/lib.rs:1956:31
>   note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
> ```
> 
> Any clues?

Did it work back in November but not anymore? If so, I have no idea what the issue is.

If this is the first time you're trying to build it, run:
```
doas pkg_add llvm
```

I added this to https://github.com/mjkillough/sioctl-rs#installation

src/host/sndio/mod.rs
AaronM04: Not that important. I'll give your idea a shot.

AaronM04: Fixed. I think I like it this way!

AaronM04:
The iOS build is failing but I suspect this is not due to any of my commits.

blackgnezdo:
> Somehow I missed the notification for this. :/
> ...
> > Any clues?
> 
> Did it work back in November but not anymore? If so, I have no idea what the issue is.

No, I had never tried to build your code before the first time I complained.

> If this is the first time you're trying to build it, run:
> 
> ```
> doas pkg_add llvm
> ```
> 
> I added this to https://github.com/mjkillough/sioctl-rs#installation

I do have llvm package:
```
% pkg_info llvm
Information for inst:llvm-10.0.1p8
...
```

The library is available on the system:
```
% pkg_info -L llvm | grep 'libclang.*so'
...
/usr/local/lib/libclang-cpp.so.0.0
/usr/local/lib/libclang.so.8.1
```

Yet `cargo build` reports (for d419f0fa5cdbaf47ec2e127d9fbe86a471627bf5):
```
  Compiling sndio-sys v0.0.1
error: failed to run custom build command for `sndio-sys v0.0.1`

Caused by:
  process didn't exit successfully: `/home/greg/s/cpal/target/debug/build/sndio-sys-cbc09558b43ecb1b/build-script-build` (exit code: 101)
  --- stdout
  cargo:rustc-link-lib=sndio
  cargo:rerun-if-changed=wrapper.h

  --- stderr
  thread 'main' panicked at 'Unable to find libclang: "couldn\'t find any valid shared libraries matching: [\'libclang.so\', \'libclang.so.*\'], set the `LIBCLANG_PATH` environment variable to a path where one of these files can be found (invalid: [])"', /home/greg/.cargo/registry/src/github.com-1ecc6299db9ec823/bindgen-0.53.3/src/lib.rs:1956:31
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Now, I bothered to read the error message and the resolution of my problem is:

```
% LIBCLANG_PATH=/usr/local/lib/libclang.so.8.1 cargo build
   Compiling sndio-sys v0.0.1
   Compiling cpal v0.12.1 (/home/greg/s/cpal)
    Finished dev [unoptimized + debuginfo] target(s) in 7.29s

```

I wonder why you don't have to set `LIBCLANG_PATH`.

AaronM04:
> I wonder why you don't have to set `LIBCLANG_PATH`.

It turns out that I added `export LIBCLANG_PATH=/usr/local/lib` to my profile and forgot about it...

blackgnezdo:
I proposed some code shrinking changes a [minor refactoring](https://github.com/RustAudio/cpal/compare/master...blackgnezdo:minor-refactoring?expand=1). If this works, I can massage it a bit more.

AaronM04:
@blackgnezdo thanks! It is quite an improvement. I merged your commits into this branch.

klemensn:
Hi everyone!

Over at https://github.com/jpochyla/psst/pull/191 psst switched to cpal and I'm working on OpenBSD support.
I'm using your feature branch to be able to play music at all, but unfortunately playback speed is noticably slow.
(before cpal, miniaudio was used and psst played music just fine on OpenBSD.)

My audio/rust/cpal/etc. foo is weak, hence I've only been testing cpal with psst.
Can you recommend another (possibly simpler) way of testing cpal's sndio support in this branch?
I'd like to help bring this forward in whatever way I can.

I've also rebased your branch onto latest cpal master and retested psst, but with no avail:  playback speed is still too slow.

AaronM04:
Hey @klemensn ! :wave: 

Can you define "slow"? Which of these is it?

* Stretched out/longer notes but the same pitch. A 4 minute song takes >4 minutes to play.
* Lower pitch. A 4 minute song takes >4 minutes to play. If so, it could be that psst is expecting one sample rate but cpal is using another one.
* Stuttering/choppy playback.

Lastly, is this an OpenBSD-specific problem with psst+cpal? I suspect not, but I just wanted to be sure.

klemensn:
`psst` reports 31 seconds of elapsed playback time while 60 seconds have passed
on the wall clock.

No other issues like stuttering or so, just playback speed.

Here are the outputs of `cpal`'s examples built with `cargo build --examples`.
`beep` produces a since, but I don't know which frequency (the source code did
not immediately give away this information).

```
$ ./target/release/examples/feedback
Using input device: "sndio default device"
Using output device: "sndio default device"
Attempting to build both streams with f32 samples and `StreamConfig { channels: 1, sample_rate: SampleRate(48000), buffer_size: Default }`.
Successfully built streams.
Starting the input and output streams with `150` milliseconds of latency.
Playing for 3 seconds... 
input stream fell behind: try increasing latency
input stream fell behind: try increasing latency
input stream fell behind: try increasing latency
input stream fell behind: try increasing latency
input stream fell behind: try increasing latency
input stream fell behind: try increasing latency
input stream fell behind: try increasing latency
input stream fell behind: try increasing latency
input stream fell behind: try increasing latency
input stream fell behind: try increasing latency
input stream fell behind: try increasing latency
input stream fell behind: try increasing latency
input stream fell behind: try increasing latency
input stream fell behind: try increasing latency
Done!
```

```
$ ./target/release/examples/enumerate
Supported hosts:
  [Sndio]
Available hosts:
  [Sndio]
sndio
  Default Input Device:
    Some("sndio default device")
  Default Output Device:
    Some("sndio default device")
  Devices: 
  1. "sndio default device"
    Default input stream config:
      SupportedStreamConfig { channels: 1, sample_rate: SampleRate(48000), buffer_size: Range { min: 480, max: 4800 }, sample_format: I16 }
    All supported input stream configs:
      1.1. SupportedStreamConfigRange { channels: 1, min_sample_rate: SampleRate(48000), max_sample_rate: SampleRate(48000), buffer_size: Range { min: 480, max: 4800 }, sample_format: I16 }
      1.2. SupportedStreamConfigRange { channels: 1, min_sample_rate: SampleRate(8000), max_sample_rate: SampleRate(8000), buffer_size: Range { min: 80, max: 800 }, sample_format: I16 }
      1.3. SupportedStreamConfigRange { channels: 1, min_sample_rate: SampleRate(44100), max_sample_rate: SampleRate(44100), buffer_size: Range { min: 441, max: 4410 }, sample_format: I16 }
    Default output stream config:
      SupportedStreamConfig { channels: 1, sample_rate: SampleRate(48000), buffer_size: Range { min: 480, max: 4800 }, sample_format: I16 }
    All supported output stream configs:
      1.1. SupportedStreamConfigRange { channels: 1, min_sample_rate: SampleRate(48000), max_sample_rate: SampleRate(48000), buffer_size: Range { min: 480, max: 4800 }, sample_format: I16 }
      1.2. SupportedStreamConfigRange { channels: 1, min_sample_rate: SampleRate(8000), max_sample_rate: SampleRate(8000), buffer_size: Range { min: 80, max: 800 }, sample_format: I16 }
      1.3. SupportedStreamConfigRange { channels: 1, min_sample_rate: SampleRate(44100), max_sample_rate: SampleRate(44100), buffer_size: Range { min: 441, max: 4410 }, sample_format: I16 }
```

```
$ ./target/release/examples/beep
Output device: sndio default device
Default output config: SupportedStreamConfig { channels: 1, sample_rate: SampleRate(48000), buffer_size: Range { min: 480, max: 4800 }, sample_format: I16 }
```

klemensn:
FWIW, recording through `cpal` works correctly on OpenBSD.
[aucat(1)](https://man.openbsd.org/aucat.1) plays the file in the correct speed, without problems.

```
$ ./target/release/examples/record_wav
Input device: sndio default device
Default input config: SupportedStreamConfig { channels: 1, sample_rate: SampleRate(48000), buffer_size: Range { min: 480, max: 4800 }, sample_format: I16 }
Begin recording...
Recording /home/kn/src/cpal/recorded.wav complete!
```
```
$ aucat -i /home/kn/src/cpal/recorded.wav
```

AaronM04:
@klemensn the `beep` example is a 440 Hz sine wave.

I see the same output as you for the `feedback` and `enumerate` examples (including the `input stream fell behind: try increasing latency` lines).

Try talking during the `feedback` example. Does your echoed voice sound normal?

I'll try it out with `psst` if I get some more time today.

AaronM04:
@klemensn I tried compiling `psst` after making it use the `cpal` from this branch, and I got this error from the `souvlaki` crate:

```
error[E0308]: mismatched types
  --> /home/aaron/.cargo/registry/src/github.com-1ecc6299db9ec823/souvlaki-0.4.0/src/platform/empty/mod.rs:13:9
   |
12 |     pub fn new(_config: PlatformConfig) -> Result<Self, Error> {
   |                                            ------------------- expected `Result<platform::platform::MediaControls, platform::platform::Error>` because of return type
13 |         Self
   |         ^^^^
   |         |
   |         expected enum `Result`, found struct `platform::platform::MediaControls`
   |         help: try using a variant of the expected enum: `Ok(Self)`
   |
   = note: expected enum `Result<platform::platform::MediaControls, platform::platform::Error>`
            found struct `platform::platform::MediaControls`

For more information about this error, try `rustc --explain E0308`.
error: could not compile `souvlaki` due to previous error
warning: build failed, waiting for other jobs to finish...
error: build failed
```
Did you encounter this error? If so, how did you fix it? Thanks.

klemensn:
Use my `openbsd` branch from https://github.com/jpochyla/psst/pull/191 to build
on OpenBSD, it contains all quirks, incl. an update to souvlaki where I fixed
the error you ran into.

Speaking during the `feedback` example plays back as expected, i.e. without any
slow down, pitch or anything else.

Thanks for looking into this!

AaronM04:
@klemensn I was able to reproduce the issue with your psst branch (I did make one change to a Cargo.toml to use my local cpal repository though). I played a song I am familiar with and the voices sounded very low pitch (stretched out). I think it is either a sample-rate confusion or a stereo-mono confusion.

I also noticed an error when hitting pause:

```
[2021-11-29T21:14:24Z ERROR psst_core::audio::output] failed to stop stream: A backend-specific error has occurred: pausing is not implemented
```

This suggests two things need to be fixed in this branch before psst will be working with cpal on OpenBSD:

* Fix the sample-rate or stereo-mono confusion. An example of the latter would be psst requesting stereo playback but cpal using mono -- this would cause everything to be one octave lower (half the frequency).

* Implement pause/resume. Essentially the sndio system needs to be closed and re-created.

Unfortunately, my time is over-committed so I can't take the lead in fixing these. I can definitely help another Rust developer understand the code that I wrote. I can also suggest where to do debugging or add print statements to better understand this.

klemensn:
@AaronM04 Thanks a bunch!

@jpochyla just committed further changes, specifically
https://github.com/jpochyla/psst/commit/a77465313e9d23dd76d71811d2721c83cca805db
https://github.com/jpochyla/psst/commit/df89890dd89283a50113cd8acaa319e64d8d2fac
which I rebased onto.

Something's still off as you can see in
https://github.com/jpochyla/psst/pull/191/commits/a4faacb6b111e3b9211ee4248b4210d29d359088
but playback speed is normal now and I can properly listen to songs!

So it seems as if @jpochyla followed your suggestion to fix "stereo-mono confusion".

The pause issue you mentioned is known (to me);  I have not done anything yet to fix it.
Pausing does work, though, as does resuming playback.

midnadimple:
just wanting to ping this request. i've tested it on my own repo. it fully merges into master and the examples run well. this should be noticed

AaronM04:
Hi @midnadimple . Thanks for the ping.

I think what this needs is another developer familiar with Rust, because I don't have much time to spend on this unfortunately.

There's only a bit of work left:

1. Resolve the conflict (yes, there is one).
2. Get another code review, and address any review feedback.
3. Make sure tests pass
4. Merge it

roderickvd:
I'm sad to have to close this for now, not because I don't want to add it as a host to cpal, but because almost four years have passed and the PR has become stale. If someone wants to give this another go, feel absolutely free to reopen or create a new PR.

Thanks for everyone's efforts so far.

Issue #473, opened 2020-09-05
cpal 0.12.1 does not seem to compile for ARM (armhf)


fhennig:
I get the following error on a blank project with just `cpal` 0.12.1 as the only dependency:

```
   Compiling cpal v0.12.1
error[E0308]: mismatched types
   --> /home/felix/.cargo/registry/src/github.com-1ecc6299db9ec823/cpal-0.12.1/src/host/alsa/mod.rs:906:59
    |
906 |         BufferSize::Fixed(v) => hw_params.set_buffer_size(v as i64)?,
    |                                                           ^^^^^^^^ expected `i32`, found `i64`
    |
help: you can convert an `i64` to `i32` and panic if the converted value wouldn't fit
    |
906 |         BufferSize::Fixed(v) => hw_params.set_buffer_size((v as i64).try_into().unwrap())?,
    |                                                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: aborting due to previous error

For more information about this error, try `rustc --explain E0308`.
error: could not compile `cpal`.
```

To reproduce:

Create a new directory, go inside and run `cargo init`.  Then add `cpal = "0.12.1"` to the dependencies. Then run `cargo build --target armv7-unknown-linux-gnueabihf`, the target is configured like this:

```
[target.armv7-unknown-linux-gnueabihf]
linker = "arm-linux-gnueabihf-gcc"
rustflags = ["-L", "/usr/lib/arm-linux-gnueabihf", "-L", "/lib/arm-linux-gnueabihf"]
```
in `~/.cargo/config`.

I did install the alsa development files like this: `sudo apt install libasound2-dev:armhf`.

My regular `cargo build` runs fine.

If I install 0.11 instead it compiles too! So it must be a change in 0.12.1.

Really weird error, I'm sure I'm doing something wrong but I don't know what it could be

Ralith:
Looks like a cpal bug to me--that `as i64` probably needs to be `as c_long` or something like that.

fhennig:
I checked the code, and in the snippet that rust provides, a `alsa::pcm::Frames` has been replaced with `i64`.  This is the original line:

    BufferSize::Fixed(v) => hw_params.set_buffer_size(v as alsa::pcm::Frames)?

[src/alsa/mod.rs:L906](https://github.com/RustAudio/cpal/blob/master/src/host/alsa/mod.rs#L906)

[...]

PR #503, opened 2020-11-09
Add CI for linux on armv7 (32bit)

HEnquist:
This PR is inspired by https://github.com/RustAudio/cpal/issues/473. 
It adds an action to run cargo check and test for the 32-bit armv7 arch. 
First it builds a docker container, by starting from `rustembedded/cross:armv7-unknown-linux-gnueabihf` and adding all dependencies. Then it uses `cross` (https://crates.io/crates/cross) to run check and test.


est31 (approved):
Thanks!

PR #504, opened 2020-11-10
change from COINIT_MULTITHREADED to COINIT_APARTMENTTHREADED in wasapi

JoshuaBatty:
There is currently a problem in windows when trying to use cpal with winit 0.20. 

Winit now uses `COINIT_APARTMENTTHREADED` because it is required by the OLE API they are using to properly handle file dropping. [See this link](https://github.com/rust-windowing/winit/issues/1185#issuecomment-535322283) for more information

@bolcom ran into this problem and submitted PR #330 to address this but it was never merged, and instead decided to spin up a whole separate thread for using winit. I also noted that @ishitatsuyuki was concerned that moving to apartment threading would require marshaling. I don't really understand what this means, would be nice to hear further clarification as to what this would actually involve.

It would be nice if cpal and winit could share the same threading model so users on windows didn't have to spin up separate threads. Keen to hear poeples thoughts.

### DIFF
--- a/src/host/wasapi/com.rs
+++ b/src/host/wasapi/com.rs
@@ -4,13 +4,13 @@ use super::check_result;
 use std::ptr;
 
 use super::winapi::um::combaseapi::{CoInitializeEx, CoUninitialize};
-use super::winapi::um::objbase::COINIT_MULTITHREADED;
+use super::winapi::um::objbase::COINIT_APARTMENTTHREADED;
 
 thread_local!(static COM_INITIALIZED: ComInitialized = {
     unsafe {
         // this call can fail if another library initialized COM in single-threaded mode
         // handling this situation properly would make the API more annoying, so we just don't care
-        check_result(CoInitializeEx(ptr::null_mut(), COINIT_MULTITHREADED)).unwrap();
+        check_result(CoInitializeEx(ptr::null_mut(), COINIT_APARTMENTTHREADED)).unwrap();
         ComInitialized(ptr::null_mut())
     }
 });


ishitatsuyuki:
See #348; I still have the same stance and I'm skeptical to the changes, as the Windows documentation on threading is very confusing and no one had the expertise to prove that changing the thread model would be safe under our `Send`/`Sync` implementations.

JoshuaBatty:
Just read through #348 again. Can you suggest any small actionable steps that can be taken in order to progress this then?

ishitatsuyuki:
I don't know if you use drag and drop, but if you don't, then winit has already merged a PR allowing cpal to winit to co-exist if you initialize winit first: https://github.com/rust-windowing/winit/pull/1524

Otherwise, I'm sorry but I don't think the situation would change unless someone can "prove" apartment threading is safe. Well, I don't know, if there's constant complaints that this need to change, then I might try the less safe way and see if it breaks in public use cases, but since winit has the compatibility code for now, I doubt that it's worth introducing some potential unsafety.

dheijl:
> 
> 
> I don't know if you use drag and drop, but if you don't, then winit has already merged a PR allowing cpal to winit to co-exist if you initialize winit first: [rust-windowing/winit#1524](https://github.com/rust-windowing/winit/pull/1524)
> 
> Otherwise, I'm sorry but I don't think the situation would change unless someone can "prove" apartment threading is safe. Well, I don't know, if there's constant complaints that this need to change, then I might try the less safe way and see if it breaks in public use cases, but since winit has the compatibility code for now, I doubt that it's worth introducing some potential unsafety.

According to [this link](https://docs.microsoft.com/en-us/windows/win32/learnwin32/initializing-the-com-library) CoInitialzeEx has to be called once **per thread**, so I think it is up to the application to make sure that it doesn't use incompatible COM libraries in the same thread. Winit obviously has to live in the main (GUI) thread, but cpal (WasApi) should be able to initialized in another thread than the main thread and use COINIT_MULTITHREADED so that after that cpal calls can be called from whatever thread you like. 
Perhaps documenting this is enough as a solution?

Issue #495, opened 2020-11-02
Fixed BufferSize not functioning as expected


xasopheno:
I'd like to switch to cpal from rust-portaudio, but I'm unable to reproduce the same behavior. 

When using the `beep.rs` example. When I set the buffer size via the config like this:

```
  let config = StreamConfig {
    channels: 2,
    buffer_size: BufferSize::Fixed(2048),
    sample_rate: cpal::SampleRate(44_100),
  };
```

and then inspect the size of the output buffer length, these are the lengths I see. 

```
[bin/pad.rs:57] &output.len() = 8190
[bin/pad.rs:57] &output.len() = 2544
[bin/pad.rs:57] &output.len() = 2030
[bin/pad.rs:57] &output.len() = 2544
[bin/pad.rs:57] &output.len() = 1598
[bin/pad.rs:57] &output.len() = 908
[bin/pad.rs:57] &output.len() = 840
[bin/pad.rs:57] &output.len() = 988
[bin/pad.rs:57] &output.len() = 484
[bin/pad.rs:57] &output.len() = 1470
[bin/pad.rs:57] &output.len() = 1474
[bin/pad.rs:57] &output.len() = 1472
[bin/pad.rs:57] &output.len() = 1462
[bin/pad.rs:57] &output.len() = 1482
[bin/pad.rs:57] &output.len() = 1472
[bin/pad.rs:57] &output.len() = 1338
[bin/pad.rs:57] &output.len() = 1606
[bin/pad.rs:57] &output.len() = 1476
[bin/pad.rs:57] &output.len() = 1470
[bin/pad.rs:57] &output.len() = 1472
[bin/pad.rs:57] &output.len() = 1470
[bin/pad.rs:57] &output.len() = 1470
[bin/pad.rs:57] &output.len() = 1474
[bin/pad.rs:57] &output.len() = 1472
[bin/pad.rs:57] &output.len() = 1472
[bin/pad.rs:57] &output.len() = 810
[bin/pad.rs:57] &output.len() = 662
[bin/pad.rs:57] &output.len() = 1472
[bin/pad.rs:57] &output.len() = 1470
[bin/pad.rs:57] &output.len() = 1474
[bin/pad.rs:57] &output.len() = 1472
[bin/pad.rs:57] &output.len() = 1472
[bin/pad.rs:57] &output.len() = 1472
[bin/pad.rs:57] &output.len() = 1470
[bin/pad.rs:57] &output.len() = 1472
[...]

PR #520, opened 2020-12-11
[ALSA] Improve stream setup parameters.

### LINKED ISSUE #369: Problem with playback under arch linux.

### LINKED ISSUE #391: Problem when setting buffer time (alsa backend)

### LINKED ISSUE #432: [Alsa Host] Start threshold of 0

alexmoon:
* Always set period before setting buffer size, even if `BufferSize::Fixed`.
* Set `start_threshold` based on `alsa::Direction` to avoid underruns on playback streams.

Also refactor the return values of the set params functions to reflect their actual use and resolve some clippy warnings (including a logic error in `stream_timestamp()`).

This pull request incorporates the changes from @sniperrifle2004 in #433 and #431. Based on that it should fix #369, fix #391 and fix #432.

### DIFF
--- a/src/host/alsa/mod.rs
+++ b/src/host/alsa/mod.rs
@@ -251,11 +251,8 @@ impl Device {
             Err((e, _)) => return Err(e.into()),
             Ok(handle) => handle,
         };
-        let can_pause = {
-            let hw_params = set_hw_params_from_format(&handle, conf, sample_format)?;
-            hw_params.can_pause()
-        };
-        let (_buffer_len, period_len) = set_sw_params_from_format(&handle, conf)?;
+        let can_pause = set_hw_params_from_format(&handle, conf, sample_format)?;
+        let period_len = set_sw_params_from_format(&handle, conf, stream_type)?;
 
         handle.prepare()?;
 
@@ -277,7 +274,9 @@ impl Device {
             _ => None,
         };
 
-        handle.start()?;
+        if let alsa::Direction::Capture = stream_type {
+            handle.start()?;
+        }
 
         let stream_inner = StreamInner {
             channel: handle,
@@ -387,7 +386,7 @@ impl Device {
                 }
             }
 
-            if rates.len() == 0 {
+            if rates.is_empty() {
                 vec![(min_rate, max_rate)]
             } else {
                 rates
@@ -420,14 +419,14 @@ impl Device {
             supported_formats.len() * supported_channels.len() * sample_rates.len(),
         );
         for &sample_format in supported_formats.iter() {
-            for channels in supported_channels.iter() {
+            for &channels in supported_channels.iter() {
                 for &(min_rate, max_rate) in sample_rates.iter() {
                     output.push(SupportedStreamConfigRange {
-                        channels: channels.clone(),
+                        channels,
                         min_sample_rate: SampleRate(min_rate as u32),
                         max_sample_rate: SampleRate(max_rate as u32),
                         buffer_size: buffer_size_range.clone(),
-                        sample_format: sample_format,
+                        sample_format,
                     });
                 }
             }
@@ -859,14 +858,13 @@ fn stream_timestamp(
         None => {
             let trigger_ts = status.get_trigger_htstamp();
             let ts = status.get_htstamp();
-            let nanos = timespec_diff_nanos(ts, trigger_ts)
-                .try_into()
-                .unwrap_or_else(|_| {
-                    panic!(
-                        "get_htstamp `{:?}` was earlier than get_trigger_htstamp `{:?}`",
-                        ts, trigger_ts
-                    );
-                });
+            let nanos = timespec_diff_nanos(ts, trigger_ts);
+            if nanos < 0 {
+                panic!(
+                    "get_htstamp `{:?}` was earlier than get_trigger_htstamp `{:?}`",
+                    ts, trigger_ts
+                );
+            }
             Ok(crate::StreamInstant::from_nanos(nanos))
         }
         Some(creation) => {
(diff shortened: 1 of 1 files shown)


alexmoon:
@mitchmindtree I think this PR is ready to merge. Let me know if you see anything blocking it.

strohel:
Hi, what is the status of this? It seems that merging this would unblock progress outlined in https://github.com/RustAudio/cpal/issues/495#issuecomment-750918795, which would help our adoption of `cpal`.

est31:
This PR has a conflict. I am ready to merge it but they need to be resolved.

strohel:
> This PR has a conflict. I am ready to merge it but they need to be resolved.

The conflict was easy to resolve, but I cannot push to @alexmoon's branch. Let's give Alex a bit more time to react.

In any case the rebased branch is at https://github.com/RustAudio/cpal/compare/master...strohel:alsa-setup

est31:
@strohel I've turned your branch into a PR over at #582 . Closing in favour of that PR.

PR #523, opened 2020-12-14
Examples improvements

alexmoon:
Since the Android target was added, the examples have been building as cdylibs instead of binaries. This obviously makes testing more difficult on platforms that are not Android. This PR creates a copy of the beep example specifically for building as an Android APK and sets the remaining examples back to a crate-type of bin.

This PR also adds argument parsing (using structopt) to several of the examples to allow selecting non-default devices and (for the feedback example) setting the latency. Finally, it improves error handling in the enumerate example in the case that supported_input_configs() or supported_output_configs() returns an error.

These changes make it easier to test cpal on systems with multiple audio devices.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,7 +18,7 @@ thiserror = "1.0.2"
 anyhow = "1.0.12"
 hound = "3.4"
 ringbuf = "0.2"
-structopt = "0.3.21"
+clap = { version = "2.33.3", default-features = false }
 
 [target.'cfg(target_os = "windows")'.dependencies]
 winapi = { version = "0.3", features = ["audiosessiontypes", "audioclient", "coml2api", "combaseapi", "debug", "devpkey", "handleapi", "ksmedia", "mmdeviceapi", "objbase", "profileapi", "std", "synchapi", "winbase", "winuser"] }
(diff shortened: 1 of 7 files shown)


Cargo.toml
```diff
@@ -52,22 +53,27 @@ ndk = "0.2"
...
+[[example]]
+name = "android"
+path = "examples/android.rs"
+crate-type = ["cdylib"]
+
 [[example]]
 name = "beep"
 path = "examples/beep.rs"
-crate-type = ["cdylib"]
+crate-type = ["bin"]
```
est31: Are these explicit lines needed? Can they just be removed?

est31 (changes requested):

est31:
Another issue I have with this PR: before applying it, running `cargo test --no-run` showed 40 compilation units for cargo to churn through. Now it shows 60 units. Not sure it's worth the saved code. Can you maybe disable default features? I've seen this saves 4 dependencies. Further suggestions to reduce the number are welcome!

Cargo.toml
alexmoon: I can remove the path and crate-type from the other examples if you want, but in the 2015 edition once one example is added to Cargo.toml all examples have to be explicitly named. See https://github.com/rust-lang/cargo/issues/5330 for details.

alexmoon:
> Another issue I have with this PR: before applying it, running `cargo test --no-run` showed 40 compilation units for cargo to churn through. Now it shows 60 units. Not sure it's worth the saved code. Can you maybe disable default features? I've seen this saves 4 dependencies. Further suggestions to reduce the number are welcome!

That's fair. If you want I can use the underlying `clap` crate instead of `structopt` which cuts the added compilation units down to 3 instead of 17 (on MacOS). It makes the code a bit more verbose (around 10-15 lines of argument parsing code at the start of `main` instead of 1 line) but otherwise works fine.

est31:
@alexmoon migrating to clap would be great!

Cargo.toml
est31: I was mainly wondering about the explicit crate-type. Thought bin was implied.

est31 (approved):

PR #537, opened 2021-01-31
Add Device::id()

zarik5:
This PR adds an `id` method to `Device`.

For WASAPI, the ID corresponds to the GUID of the device. On ALSA, the ID is just the device name. For now, all the other backends return an `"Unimplemented"` error.

The ID is useful when we want to store and load a device selection setting, and deal with devices with the same name (on Windows).

Tested on Windows 10 2004 and Ubuntu 18.04.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# Version 0.13.4 (2021-08-08)
+
+- wasapi: Allow both threading models and switch the default to STA
+- Update dependencies (core-foundation-sys, jni, rust-jack)
+- Alsa: improve stream setup parameters
+
 # Version 0.13.3 (2021-03-29)
 
 - Give each thread a unique name
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "cpal"
-version = "0.13.3"
+version = "0.13.4"
 authors = ["The CPAL contributors", "Pierre Krieger <pierre.krieger1708@gmail.com>"]
 description = "Low-level cross-platform audio I/O library in pure Rust."
 repository = "https://github.com/rustaudio/cpal"


Ralith:
`String` feels like a poor choice for this, especially in a world where USB and PCI vendor/device IDs exist. I expect Windows is the only platform where this API makes sense at all, which makes it a poor choice for an ostensibly portable abstraction.

zarik5:
I have to admit i'm not really familiar with the platform specific APIs or audio programming in general, but here's my thoughts.

WASAPI doesn't seem to provide vendor/device IDs, only the GUID of the device. On ALSA, device and vendor IDs can be obtained with `PCM::info()`, but they don't seem to uniquely identify the software level devices. Software devices that correspond to different surround settings or hw/plughw have the same vendor/device/subdevice id. I also looked at the web API and both approach seem to work, since the vendor id could be mapped to `MediaDeviceInfo.groupId` and the device id or universal id can be mapped to `MediaDeviceInfo.deviceId`.

I understand that `String` is not optimal, but that seemed the best choice for being serializable and for portability across backends. We could instead use different custom structures for each backend but then they would be opaque and it doesn't make much sense if we also want to serialize them, unless we also provide a method for serializing them using a trait.

Ultimately this `Device::id()` API wants to be a best-effort solution for identifiability across sessions/enumeration invocations.

antoineradio:
Hello, I think that would be useful to get a Device::id or Device::uid storing maybe a GUID provided by the audio API (I don't know the specs of low level audio stuffs but it must exist..).

It would allow to make difference between two soundcards wearing the same name and also storing a soundcard in configuration and being able to stay ready to play among the names update and co..

Currently I'm using the name because I didn't found equivalent property in the doc.. :)

Regards,

Ralith:
I also think this would be useful, and I've come around a bit on using `String`. Different platforms will have different stuff (maybe windows has a UUID, Linux could have some topological udev path) and an opaque bag of UTF-8 might be a more pleasant compromise than trying to define a complex non-exhaustive enum.

G2-Games:
I would really appreciate this functionality as well.

w1kee:
i would also like this feature

zmerp:
The lack of this feature is still causing many bugs on ALVR, when audio devices have the same name.

PR #548, opened 2021-03-11
clippy fixes

xMAC94x:


xMAC94x:
@est31  those are some clippy improvements, that where quite easy to improve :). ping

xMAC94x:
can we give each thread a UNIQUE name. in our game veloren we have 80 threads, created by many different libs, and we want a way to easier debug what thread is doing what. a thread name would help alot here.

xMAC94x:
(some ci check seems to fail, but it worked in a run before)

est31 (approved):
Thanks!

PR #552, opened 2021-03-15
More ease of use for StreamInstant

Hasenn:
This pr attempts to make `StreamInstant` easier to work with by adding:
- `TryFrom` / `TryInto` conversion with `Duration` (with an error for underflows)
- a to_duration method as sugar for `TryInto::<Duration>::try_into`

Further commits make `new` public, along with the struct's fields `secs` and `nanos`

## motivation

When writing a simple POC synth to experiment with some DSP the privateness of everything on StreamInstant was preventing me from getting the time associated with samples

```rust
fn run<T: Sample>(data: &mut [T], info: &cpal::OutputCallbackInfo) {
    let buf_start = info.timestamp().playback;
    for (i, sample) in data.iter_mut().enumerate() {
        let t = buf_start.add(Duration::new(0, (i as u32) / SAMPLE_RATE * NANOSECONDS_IN_SECOND)).unwrap(); 
        // ...
    }
}
```

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -160,6 +160,7 @@ pub enum StreamError {
         err: BackendSpecificError,
     },
 }
+
 /// Underflow while converting a StreamInstant to a Duration.
 #[derive(Clone, Debug, Error)]
 #[error("Underflow while converting a StreamInstant to a Duration.")]
--- a/src/error.rs
+++ b/src/error.rs
@@ -162,5 +162,5 @@ pub enum StreamError {
 }
 /// Underflow while converting a StreamInstant to a Duration.
 #[derive(Clone, Debug, Error)]
-#[error("the requested host is unavailable")]
+#[error("Underflow while converting a StreamInstant to a Duration.")]
 pub struct StreamInstantUnderflow;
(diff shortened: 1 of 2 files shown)


Ralith:
I think not offering direct conversions to `Duration` is intended for the same reason `std`'s `Instant` doesn't offer them: to encourage you to pick a specific reference point and compute offsets relative to it if needed.

Hasenn:
The issue with that imo is that many things we can do in DSP do not depend on that reference point apart from a phase that doesn't change how the sound is perceived.
For quick prototyping, having to handle thread state to store that reference point is very cumbersome, but i can see how it's a good practice for `real` applications.

Ralith:
What do you mean by "handle thread state"?

Hasenn:
@Ralith to get a StreamInstant that i can use as a reference in the callback of build_output_stream i need to keep track of the first StreamInstant i get, which requires keeping state, which i couldn't find an easy way to do while having my callback be `FnMut(&mut [T], &OutputCallbackInfo) + Send + 'static`

Hasenn:
std's `Instant` has an `elapsed() -> Duration` method that would help, but i guess it would need some overhead

Ralith:
You can keep all the state you like inside a closure that satisfies `FnMut` by closing over it.

Issue #554, opened 2021-03-18
PipeWire support


Be-ing:
PipeWire will soon be replacing both JACK and PulseAudio on Linux starting with Fedora 34 which will be released in a few weeks. Arch currently has packages that make it easy to switch to PipeWire. I am unclear about Ubuntu's plans and just [asked them about that](https://bugs.launchpad.net/ubuntu/+source/pipewire/+bug/1802533/comments/36).

PipeWire reimplements the JACK and PulseAudio APIs. Currently the PipeWire developer is not recommending to use the new PipeWire API if applications already work well with PipeWire via JACK or PulseAudio. I am not familiar enough with cpal to know, but there might be some benefits to using the PipeWire API instead of using PipeWire via the JACK API.

I have attempted to test cpal's JACK backend with PipeWire but have not gotten it to work. The `pw-uninstalled.sh` script sets the `LD_LIBRARY_PATH` environment variable so PipeWire's reimplementation of libjack is loaded before JACK's original library. I know Rust links statically to other Rust libraries but I am unclear how Rust links to dynamic C libraries. Maybe the `LD_LIBRARY_PATH` trick does not work with Rust?

```
cpal on  master is 📦 v0.13.2 via 🦀 v1.49.0 
❯ cargo run --release --example beep --features jack
    Finished release [optimized] target(s) in 0.03s
     Running `target/release/examples/beep`
Output device: default
Default output config: SupportedStreamConfig { channels: 2, sample_rate: SampleRate(44100), buffer_size: Range { min: 170, max: 1466015503 }, sample_format: F32 }
memory allocation of 5404319552844632832 bytes failed
fish: “cargo run --release --example b…” terminated by signal SIGABRT (Abort)

cpal on  master is 📦 v0.13.2 via 🦀 v1.49.0 
❯ ~/sw/pipewire/pw-uninstalled.sh
Using default build directory: /home/be/sw/pipewire/build
Welcome to fish, the friendly interactive shell
Type `help` for instructions on how to use fish

cpal on  master is 📦 v0.13.2 via 🦀 v1.49.0 
❯ cargo run --release --example beep --features jack
    Finished release [optimized] target(s) in 0.03s
     Running `target/release/examples/beep`
Output device: default
Default output config: SupportedStreamConfig { channels: 2, sample_rate: SampleRate(44100), buffer_size: Range { min: 170, max: 1466015503 }, sample_format: F32 }
memory allocation of 5404319552844632832 bytes failed
[...]

PR #556, opened 2021-03-18
Fix default port selection parameters for PipeWire compatibility

ishitatsuyuki:
Ref: #554

### DIFF
--- a/src/host/jack/stream.rs
+++ b/src/host/jack/stream.rs
@@ -3,6 +3,8 @@ use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::{Arc, Mutex};
 use traits::StreamTrait;
 
+const FLOAT_MONO_AUDIO: &str = "32 bit float mono audio";
+
 use crate::{
     BackendSpecificError, Data, InputCallbackInfo, OutputCallbackInfo, PauseStreamError,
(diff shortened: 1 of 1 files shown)


src/host/jack/stream.rs
```diff
@@ -149,9 +151,9 @@ impl Stream {
     pub fn connect_to_system_outputs(&mut self) {
         // Get the system ports
         let system_ports = self.async_client.as_client().ports(
-            Some("system:playback_.*"),
             None,
-            jack::PortFlags::empty(),
+            Some(FLOAT_MONO_AUDIO),
+            jack::PortFlags::IS_INPUT | jack::PortFlags::IS_PHYSICAL,
```
Be-ing, line 156: IIUC this just guesses to use an arbitrary hardware device?

Be-ing:
If the intention for the cpal API is to access the sound server's default hardware interface, I think it might be best to leave it as is using `system`. According to this [PipeWire issue](https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/446#note_807327), PipeWire should provide a `system` JACK device but this seems to not work properly yet.

ishitatsuyuki:
Okay, closing this then.

PR #558, opened 2021-03-18
Don't call AudioObjectGetPropertyData again if CFStringGetCStringPtr fails

micolous:
There's no indication that `CFStringGetCStringPtr` failing is destructive for the `CFString`:

https://developer.apple.com/documentation/corefoundation/1542133-cfstringgetcstringptr

So calling `AudioObjectGetPropertyData` a second time to get the device name isn't necessary.

It looks like this was an unintentional error in #399, but I don't have a wide-character install of macOS to try this against.

cc @thavlik

### DIFF
--- a/src/host/coreaudio/macos/mod.rs
+++ b/src/host/coreaudio/macos/mod.rs
@@ -162,15 +162,6 @@ impl Device {
 
             let c_string: *const c_char = CFStringGetCStringPtr(device_name, kCFStringEncodingUTF8);
             if c_string.is_null() {
-                let status = AudioObjectGetPropertyData(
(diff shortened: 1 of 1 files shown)


nico-franco-gomez:
I tested this and seems to be right. Maybe it could get merged? It's a minor thing, but it would be nice!

wgibbs-rs:
No part of this code (that I could find) is still present in `/macos/mod.rs`, and does not merge without conflicts. I strongly recommend closing this PR without merging.

roderickvd:
Thanks for the earlier PR. Like @wgibbs-rs stated, this code is no longer being used.

PR #561, opened 2021-03-19
Add JACK support to README

### LINKED ISSUE #529: Update README to note that Jack is now supported on Linux.
Currently only ALSA support is mentioned. But it seems that JACK is also supported, but the readme doesn't mention this.

Frando:
The readme doesn't mention JACK support which was merged a while ago (#389).

Fixes #529


est31 (approved):
Thanks!

PR #562, opened 2021-03-19
Fix alsa in some configs outputting bad audio

JCapucho:
This was introduced in `6b91485` and it caused issues with some alsa
configurations like with `pipewire-alsa` for me which caused the audio to be
become weirdly distorted.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "cpal"
-version = "0.13.2"
+version = "0.13.3"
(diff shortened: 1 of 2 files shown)


JCapucho:
The change are only on the alsa backend so I'm not sure why CI is failing.
Also I've asked someone to test, they were using pulseaudio and said everything seemed okay sound and perf wise.

est31 (approved):
Seems like a copy paste fix? I'll approve it. Thanks

Issue #383, opened 2020-03-20
failed to select a version for `clang-sys` which could resolve this conflict


duhruh:
I'm running into issues trying to run the examples with the asio feature turned on on
I might be doing something wrong here, wondering if anybody can help.
```
C:\project\foo>cargo run
    Updating crates.io index
error: failed to select a version for `clang-sys`.
    ... required by package `bindgen v0.42.0`
    ... which is depended on by `asio-sys v0.1.0`
    ... which is depended on by `cpal v0.11.0`
    ... which is depended on by `audio v0.1.0 (C:\project\foo)`
versions that meet the requirements `^0.24` are: 0.24.0

the package `clang-sys` links to the native library `clang`, but it conflicts with a previous package which links to `clang` as well:
package `clang-sys v0.28.1`
    ... which is depended on by `bindgen v0.53.1`
    ... which is depended on by `coreaudio-sys v0.2.4`
    ... which is depended on by `coreaudio-rs v0.9.1`
    ... which is depended on by `cpal v0.11.0`
    ... which is depended on by `audio v0.1.0 (C:\project\foo)`
```

mitchmindtree:
Hi @duhruh thanks for posting!

This is a particularly nasty issues where both coreaudio and asio require different versions of clang. See RustAudio/coreaudio-rs#68 and nannou-org/nannou#502. 

The amusing thing is that it's not possible to build both of these crates on the same platform, and the feature gating ensures it, however cargo is unaware of this and so forces both crates to fall back to the last valid version that has a matching clang-sys dependency.

We can potentially work around this for now by updating `asio-sys` to use the same version of `bindgen` that `coreaudio-sys` does. This unfortunately won't solve the issue where an unknowing user might update either `asio-sys`' or `coreaudio-sys`' version of bindgen in the future.

Intuitively it seems to me that cargo should allow this case, but I don't recall what the reason is for why it does not. I do know that it's been a known issue with cargo for a long time.

mitchmindtree:
Thinking on this a little more, I think the better solution might be to remove the dependency on bindgen from `asio-sys` altogether and include the bindings in the repo for a specific version of asio. We can leave the bindings generation code in a separate `asio-generate-bindings` crate, so that future users who require a newer version of asio-sys can regenerate the bindings and update `asio-sys` when necessary.

9to1url:
I also encounter this issue in MacOS:

```
$ cargo run
[...]

PR #568, opened 2021-04-09
Bumped asio-sys bindgen version

MarkDDR:
Note: I don't have a Windows computer so I can not test this change, but I'm assuming the `bindgen` interface didn't dramatically change in the past few versions.

This change is particularly significant as it would solve issue #383 for a good while as now both `asio-sys` and `coreaudio-sys` will be using a version of `bindgen` that targets `clang-sys v1.0`.

Releasing `asio-sys v0.2.1` after this would be a good idea as it would fix `cpal` compilation for people on M1 Macs, since `asio-sys` would no longer drag `coreaudio-sys` to an older version that doesn't work for M1 Macs

### DIFF
--- a/asio-sys/Cargo.toml
+++ b/asio-sys/Cargo.toml
@@ -10,7 +10,7 @@ keywords = ["audio", "sound", "asio", "steinberg"]
 build = "build.rs"
 
 [target.'cfg(any(target_os = "windows"))'.build-dependencies]
-bindgen = "0.54.0"
+bindgen = "0.58.0"
 walkdir = "2"
 cc = "1.0.25"
 


MarkDDR:
Well it looks like the Windows build failed on the build bot. As I said above I can't debug this issue as I do not own a Windows computer, but to anyone else who can do that, I'll mention that any `bindgen` version `0.55` or above would work to solve the issue I mentioned, as that is the first version that starts to target `clang-sys v1.0`. So if you can't get it to work on `bindgen v0.58` for whatever reason, you could also try the version in between

est31:
> Unable to find libclang: "the `libclang` shared library at C:\\msys64\\mingw64\\bin\\libclang.dll could not be opened: LoadLibraryExW failed

huh that's a weird bug.

@MarkDDR to some limited extent you can debug against the CI.

est31:
Merged #620 instead, although that one too triggered the CI failure. It would be great to have that one resolved.

PR #577, opened 2021-05-22
Bump bindgen version for asio-sys

toastmod:
So that #383 doesn't happen again :)

### DIFF
--- a/asio-sys/Cargo.toml
+++ b/asio-sys/Cargo.toml
@@ -10,7 +10,7 @@ keywords = ["audio", "sound", "asio", "steinberg"]
 build = "build.rs"
 
 [target.'cfg(any(target_os = "windows"))'.build-dependencies]
-bindgen = "0.54.0"
+bindgen = "0.56.0"
 walkdir = "2"
 cc = "1.0.25"
 


est31:
Runs into the same CI failure as #568 . Do you have a windows computer to debug it? Also why not update to 0.58?

toastmod:
Yes I can debug it. I was able to compile it and run it. Must just be the tests to figure out.

ottworks:
This is preventing me from using cpal!

est31:
FTR it's ok if the emscripten tests fail, as the error seems unrelated. But the windows 32 bit CI failure is quite related I think. It should be fixed before this is merged.

toastmod:
Thanks, I honestly totally forgot to go back and fix this up sorry I've been doing lots of school. I'll try taking a look right now.

toastmod:
Actually it looks like libclang is missing in the environment somehow? Not sure how this is missing, idk how these github test environments work. But here is the error. I think that this should still get merged however as it seems to be affecting others.
```
  --- stderr
  thread 'main' panicked at 'Unable to find libclang: "the `libclang` shared library at C:\\msys64\\mingw64\\bin\\libclang.dll could not be opened: LoadLibraryExW failed"', C:\Users\runneradmin\.cargo\registry\src\github.com-1ecc6299db9ec823\bindgen-0.56.0\src/lib.rs:1922:31
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Error: Process completed with exit code 1.
```

est31:
@toastmod the installation step of libclang is here: https://github.com/RustAudio/cpal/blob/47ef62d41392f39b2d55f5a672b6425c273c9b23/.github/workflows/cpal.yml#L221

The CI does work for i386 windows without this PR applied, see #581.

est31:
Merged #620 instead, although that one too triggered the CI failure. It would be great to have that one resolved.

PR #579, opened 2021-05-23
Update feedback examples to use mpsc's sync_channel.

divark:
### In General
1. Run `cargo test` to ensure that everything still passes.
2. Navigate to the `examples` directory.
3. Run `cargo run --example feedback`
4. Observe feedback from the default input device to the default output device.
### On Apple Device
1. Navigate to `ios-feedback`.
2. Follow instructions in `README.md` to build and run the project.
3. Observe feedback from the default input device to the default output device.

## Changes
- Removed ringbuf dependency from dev-dependencies.
- Replaced buffer with transmitter/receiver pattern.
- Some documentation polish.

What are your thoughts on these changes?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,7 +17,6 @@ thiserror = "1.0.2"
 [dev-dependencies]
 anyhow = "1.0.12"
 hound = "3.4"
-ringbuf = "0.2"
 clap = { version = "2.33.3", default-features = false }
 
 [target.'cfg(target_os = "windows")'.dependencies]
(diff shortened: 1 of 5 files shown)


Ralith:
A ring buffer is used here for a reason; std::mpsc channels can block the audio thread and cause popping.

divark:
Ah, okay, so ring buffers are used to prevent popping, which mpsc channels are more prone to due to its potential to block. That makes sense. Thanks for the clarification! I'll go ahead and close this then.

PR #588, opened 2021-06-29
Update dependency

remilauzier:


est31 (changes requested):
Can you remove the a.b.c → a.b.d changes? Those serve no purpose and only add noise.

remilauzier:
I am not sure what you are talking about, but I already squash all my commit to one. Or do you talk about the number like that: ``dependency = "0.2"`` instead of ``dependency = "0.2.2"``

est31:
@remilauzier yeah those aren't neccessary.

est31 (approved):
Thanks!

PR #592, opened 2021-07-09
Relax cpal wasm bounds

yamadapc:
Fixes compiling cpal together with other packages such as the iced master branch.

Let me know if this does not make sense & in that case sorry for the noise.


est31:
This doesn't do anything, as `^` is already the default.

Quoting the [cargo docs](
https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#specifying-dependencies-from-cratesio):

> The string "0.1.12" is a semver version requirement. Since this string does not have any operators in it, it is interpreted the same way as if we had specified "^0.1.12", which is called a caret requirement.

From your github it seems you are coming from the npm world. Here the default is `~` from what I can tell, which works a bit different from what `^` means in npm. Even then IIRC there is no difference. Happy Rust learning!

PR #595, opened 2021-07-30
Adapt to rust 2018

remilauzier:


est31 (approved):
Good to go, if the conflicts are resolved.

Maybe also update to 2021 if you are at it.

PR #612, opened 2021-11-02
Add sample-based generation example: synth_tones

Wandalen:
Sample-based generation of sound.

It is refactored version of the example `beep`. I bubbled up a callback to generate a sample. Also, I replaced `unwrap` with proper handling of errors. That does not add complexity because `anyhow` and `?` are already in use. That's the first step I would like it was done for me to start playing with `cpal` and integrating it into my projects. 

Hopefully, it will be useful for not only me.

#613

### DIFF
--- a/examples/synth_tones.rs
+++ b/examples/synth_tones.rs
@@ -1,0 +1,120 @@
+/* This example expose parameter to pass generator of sample.
+Good starting point for integration of cpal into your application.
+*/
+
+extern crate anyhow;
+extern crate clap;
+extern crate cpal;
+
+use cpal::traits::{DeviceTrait, HostTrait, StreamTrait};
+
+//
+
+fn main() -> anyhow::Result<()> {
+    let stream = stream_setup_for(sample_next)?;
+    stream.play()?;
+    std::thread::sleep(std::time::Duration::from_millis(3000));
+    return Ok(());
+}
+
+//
+
+fn sample_next(o: &mut SampleRequestOptions) -> f32 {
+    o.tick();
+    o.tone(440.) * 0.1 + o.tone(880.) * 0.1
+    /*
+    combination of several tones
+    */
+}
+
+//
+
+pub struct SampleRequestOptions {
+    pub sample_rate: f32,
+    pub sample_clock: f32,
+    pub nchannels: usize,
+}
+
+//
+
+impl SampleRequestOptions {
+    fn tone(&self, freq: f32) -> f32 {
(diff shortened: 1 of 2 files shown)


Wandalen:
Please let me know if any adjustments are required.

examples/synth_tones.rs
```diff
@@ -0,0 +1,120 @@
...
+Good starting point for integration of cpal into your application.
+*/
+
+extern crate anyhow;
+extern crate clap;
+extern crate cpal;
+
+use cpal::traits::{DeviceTrait, HostTrait, StreamTrait};
+
+//
```
est31: Can you remove these // ?

examples/synth_tones.rs
```diff
@@ -0,0 +1,120 @@
...
+
+fn main() -> anyhow::Result<()> {
+    let stream = stream_setup_for(sample_next)?;
+    stream.play()?;
+    std::thread::sleep(std::time::Duration::from_millis(3000));
+    return Ok(());
```
est31: You don't need a return statement here

examples/synth_tones.rs
```diff
@@ -0,0 +1,120 @@
...
+}
+
+//
+
+fn sample_next(o: &mut SampleRequestOptions) -> f32 {
+    o.tick();
+    o.tone(440.) * 0.1 + o.tone(880.) * 0.1
+    /*
+    combination of several tones
```
est31: This can be a // comment

examples/synth_tones.rs
```diff
@@ -0,0 +1,120 @@
...
+        .ok_or_else(|| anyhow::Error::msg("Default output device is not available"))?;
+    println!("Output device : {}", device.name()?);
+
+    let config = device.default_output_config()?;
+    println!("Default output config : {:?}", config);
+
+    Ok((host, device, config))
+}
+
+pub fn _stream_make<T, F>(
```
est31: Can you remove the leading underscore in the function name?

est31 (changes requested):

Wandalen:
@est31 sure! done. anything else?

est31 (approved):
Thanks!

Wandalen:
;)

est31:
Sorry for the long delay, I think I've enabled CI on it and then lost track of the PR. It's a bit annoying that it doesn't ping you after CI is finished.

Wandalen:
Thanks for good open source)

Ralith:
If you set CI as required in branch prediction, you can enable automerge on CI green, which is pretty handy in these cases.

PR #618, opened 2021-11-23
Faster sample format conversion

RamType0:


src/samples_formats.rs
```diff
@@ -77,11 +73,12 @@ unsafe impl Sample for i16 {
 
     #[inline]
     fn to_f32(&self) -> f32 {
-        if *self < 0 {
-            *self as f32 / -(i16::MIN as f32)
-        } else {
-            *self as f32 / i16::MAX as f32
-        }
+        const POSITIVE_MULTIPLIER:f32 = 1.0 / i16::MAX as f32;
+        const NEGATIVE_MULTIPLIER:f32 = 1.0 / -(i16::MIN as f32);
```
ollpu: You removed the asymmetry in `f32::to_i16` (which I think makes sense), why not here? ([relevant post](http://blog.bjornroche.com/2009/12/int-float-int-its-jungle-out-there.html))

RamType0: Removing asymmetry at here makes test failing.

src/samples_formats.rs
```diff
@@ -51,11 +51,7 @@ unsafe impl Sample for u16 {
 
     #[inline]
     fn to_i16(&self) -> i16 {
-        if *self >= 32768 {
-            (*self - 32768) as i16
-        } else {
-            (*self as i16) - 32767 - 1
-        }
+        (*self as i16).wrapping_add(i16::MIN)
```
est31, line 54: According to godbolt, this compiles to the same assembly as the old code. Not sure this is a speedup? Or do you mean debug mode?

Screenshot_20220113_160342

src/samples_formats.rs
```diff
@@ -91,11 +88,7 @@ unsafe impl Sample for i16 {
 
     #[inline]
     fn to_u16(&self) -> u16 {
-        if *self < 0 {
-            (*self - i16::MIN) as u16
-        } else {
-            (*self as u16) + 32768
-        }
+        self.wrapping_add(i16::MIN) as u16
```
est31, line 90: Same here, this also compiles to the same assembly if opt-level is 1 or above:

Screenshot_20220113_161137

src/samples_formats.rs
```diff
@@ -117,16 +110,12 @@ unsafe impl Sample for f32 {
...
-        } else {
-            (-*self * i16::MIN as f32) as i16
-        }
+        (*self * F32_TO_16BIT_INT_MULTIPLIER).floor() as i16
     }
 
     #[inline]
     fn to_u16(&self) -> u16 {
-        (((*self + 1.0) * 0.5) * u16::MAX as f32).round() as u16
+        self.mul_add(F32_TO_16BIT_INT_MULTIPLIER,F32_TO_16BIT_INT_MULTIPLIER).round() as u16
```
est31: This creates different code even with optimizations enabled, not sure about the performance impact. Have you benchmarked it?

Old code:

Screenshot_20220113_161415

New code:

Screenshot_20220113_161527

src/samples_formats.rs
```diff
@@ -117,16 +110,12 @@ unsafe impl Sample for f32 {
 
     #[inline]
     fn to_i16(&self) -> i16 {
-        if *self >= 0.0 {
-            (*self * i16::MAX as f32) as i16
-        } else {
-            (-*self * i16::MIN as f32) as i16
-        }
+        (*self * F32_TO_16BIT_INT_MULTIPLIER).floor() as i16
```
est31: This one seems to actually improve the code, but I'm not sure if the assymetry removal is good, doesn't that destroy roundtrip-ability?

est31 (changes requested):

src/samples_formats.rs
ollpu: Yeah, they should be consistent with each other, but I think a symmetric version makes more sense in general.

src/samples_formats.rs
RamType0: This function is a really hot path and could be resulted in audio interrupted in debugging.
So I also mean debug mode.

est31: Oh I see. Alright then, it can go in.

RamType0:
[I made a benchmarks](https://github.com/RamType0/sample_format_conversion_benchmarks).
And I executed this with `RUSTFLAGS='-C target-feature=+avx2,+fma'`.
here is the result
>test tests::bench_current_f32_to_i16 ... bench:         629 ns/iter (+/- 387)
test tests::bench_current_f32_to_u16 ... bench:         820 ns/iter (+/- 188)
test tests::bench_current_i16_to_f32 ... bench:         360 ns/iter (+/- 72)
test tests::bench_my_f32_to_i16      ... bench:         464 ns/iter (+/- 179)
test tests::bench_my_f32_to_u16      ... bench:         619 ns/iter (+/- 140)
test tests::bench_my_i16_to_f32      ... bench:         560 ns/iter (+/- 164)

i16_to_f32 was mistake. I reverted it.

src/samples_formats.rs
RamType0: I reverted it because it was slower

src/samples_formats.rs
RamType0: It is almost 33% faster in my environment.
https://github.com/RustAudio/cpal/pull/618#issuecomment-1012377045

est31:
Results for me:
```
test tests::bench_current_f32_to_i16 ... bench:         846 ns/iter (+/- 26)
test tests::bench_current_f32_to_u16 ... bench:       1,962 ns/iter (+/- 77)
test tests::bench_current_i16_to_f32 ... bench:         598 ns/iter (+/- 40)
test tests::bench_my_f32_to_i16      ... bench:         850 ns/iter (+/- 17)
test tests::bench_my_f32_to_u16      ... bench:       1,734 ns/iter (+/- 201)
test tests::bench_my_i16_to_f32      ... bench:         953 ns/iter (+/- 67)
```

With `RUSTFLAGS='-C target-feature=+avx2,+fma'`:
```
test tests::bench_current_f32_to_i16 ... bench:         740 ns/iter (+/- 25)
test tests::bench_current_f32_to_u16 ... bench:         957 ns/iter (+/- 53)
test tests::bench_current_i16_to_f32 ... bench:         479 ns/iter (+/- 28)
test tests::bench_my_f32_to_i16      ... bench:         551 ns/iter (+/- 30)
test tests::bench_my_f32_to_u16      ... bench:         722 ns/iter (+/- 23)
test tests::bench_my_i16_to_f32      ... bench:         521 ns/iter (+/- 15)
```

So indeed the newer versions of `f32_to_i16` and `f32_to_u16` are faster. `f32_to_i16` is only faster with the CPU features turned on. 

So `f32_to_u16` is good to go, I'm not so sure about `f32_to_i16` due to the symmetry issue.

est31:
Also there are formatting issues, see the CI failure.

est31:
Can you remove the `f32_to_i16` change for now? Then this PR should be good. We can discuss about faster f32 <-> i16 conversions in the future.

PR #620, opened 2021-11-25
bump asio-sys bindgen to 0.56, change version to 0.2.1

l1npengtul:


est31 (approved):

ishitatsuyuki:
Sorry for a late bump but it seems this is breaking Windows CI? (Does it build on your computer?) It seems that some enum-related defaults is changed in bindgen, so you might need to configure that in build.rs.

est31:
Yeah it broke the CI, and I merged without realizing that :/. The PR had multiple dupes, and staying on an old version forever is not good either. Users don't seem to complain, so it is likely some CI related issue rather than a bug (entirely possible that it is a bug tho).

PR #622, opened 2021-12-02
Fix ASIO driver.prepare_output_stream should take streams.input

hyranno:
In ASIO, input and output buffers need to be created at the same time.
So `Driver::prepare_output_stream` takes infos of already exist input.
But `Device::get_or_create_output_stream` was passing infos of exist output.
This breaks already exist input stream.

`Device::get_or_create_intput_stream` was fine, passing infos of exist output.


est31 (approved):
Thanks!

PR #624, opened 2021-12-13
update jack to 0.8

Be-ing:
This fixes the build with Pipewire's implementation of JACK:
https://github.com/RustAudio/rust-jack/pull/154


est31 (approved):
Thanks!

Be-ing:
Thanks for the quick merge!

PR #629, opened 2022-01-12
Cleanup and document cpal's ALSA backend

nyanpasu64:
Questions:

- [ ] Are the changes in "Remove report_error(), replace monadic error handling with match" and "Alternatively..." improvements?
- [ ] Should I merge the two commits?
- [ ] "Set it anyway I guess" is arguably too informal.
- [ ] Should `poll_descriptors_and_prepare_buffer()` loop and never return `Continue`?
- [ ] ~~Should we skip calling `set_start_threshold()` for capture streams?~~
	- The start threshold actually *does* matter in capture streams: when we hit `PollDescriptorsFlow::XRun`, we prepare the device but don't start it explicitly. Should we do so? (it doesn't make much of a difference in practice, since we read immediately from input streams... except we don't, we poll them! Is this a CPAL bug where input streams never restart after xrun?!)
	- I tried testing xrun. But when I suspend the cpal process (feedback talking to pipewire-alsa) with Ctrl+Z, then fg a second later, it somehow doesn't hit xrun. This even happens if I add a 500ms sleep in the loop of `input_stream_worker()`!
	- And cpal feedback `cargo run --example feedback -- front:CARD=Generic,DEV=0 front:CARD=Generic,DEV=0` won't detect `front:CARD=Generic,DEV=0` as an output device, only input, so I can't run it talking to my physical hardware! (`aplay /dev/urandom -D front:CARD=Generic,DEV=0 --format=S16_LE --channels=2 --rate=48000` *does* work.)
	- `cargo run --example record_wav -- front:CARD=Generic,DEV=0` also doesn't hit xrun when I ctrl+z. So I can't actually reproduce this case to see if cpal bugs out or not.
		- It also produces an empty wav file whether I ctrl+z or not. It produces a proper wav if I don't sleep in `input_stream_worker()`.
	- If I add a stop threshold, it never actually kicks in. Weeeeiiiiiird.
	- `stream.channel.state()` remains in `Running`, and `poll_descriptors_and_prepare_buffer()` always returns `PollDescriptorsFlow::Ready`, even with the 500ms sleeps.

It would be nice to write multi-paragraph-level commentary on the code, but I'm not sure where I should put it. In the file itself? In a new DESIGN.md file in the alsa directory? I have a Google Doc, but I don't want to share it because there are some IRC chat logs I didn't ask permission to share publicly.

Notes to self:

- [ ] Verify that the code is functionally identical to before.

### DIFF
--- a/src/host/alsa/mod.rs
+++ b/src/host/alsa/mod.rs
@@ -556,18 +556,22 @@ fn input_stream_worker(
 ) {
     let mut ctxt = StreamWorkerContext::default();
     loop {
-        let flow = report_error(
-            poll_descriptors_and_prepare_buffer(&rx, stream, &mut ctxt),
-            error_callback,
-        )
-        .unwrap_or(PollDescriptorsFlow::Continue);
+        let flow = match poll_descriptors_and_prepare_buffer(&rx, stream, &mut ctxt) {
+            Ok(val) => val,
+            Err(err) => {
+                error_callback(err.into());
+                PollDescriptorsFlow::Continue
+            }
+        };
 
         match flow {
             PollDescriptorsFlow::Continue => {
                 continue;
             }
             PollDescriptorsFlow::XRun => {
-                report_error(stream.channel.prepare(), error_callback);
+                if let Err(err) = stream.channel.prepare() {
+                    error_callback(err.into());
+                }
                 continue;
             }
             PollDescriptorsFlow::Return => return,
@@ -582,14 +586,15 @@ fn input_stream_worker(
                     StreamType::Input,
                     "expected input stream, but polling descriptors indicated output",
                 );
-                let res = process_input(
+                if let Err(err) = process_input(
                     stream,
                     &mut ctxt.buffer,
                     status,
                     delay_frames,
                     data_callback,
-                );
-                report_error(res, error_callback);
+                ) {
+                    error_callback(err.into());
+                }
             }
         }
     }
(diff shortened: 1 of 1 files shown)


est31 (changes requested):
The first commits are fine, but as for the second two commits, I'd like a different approach. The places which the "Alternatively..." commit touches are in fact improved, but I'm not sure about the other places. Generally it would be better to change those other places to something like `err().map(Into::into).map(error_callback)`, if that works.

nyanpasu64:
Alternatives:

```rust
            PollDescriptorsFlow::XRun => {
                stream.channel.prepare().err().map(|err| {
                    error_callback(err.into());
                });
                continue;
            }
```

I can tolerate it, though I still prefer `if let`. To me, the closure and semicolon make it clear that this is a procedure called for its effects, rather than a function called for its value. Though (as I mentioned below) had this been an iterator, this would be misleading, since `map` would be lazy and `error_callback` would never run.

- When trying to write this code independently, I didn't know the method was called `.err()`, and had to look at your comment again to find out. Arguably `match` and `if let` are more minimal/versatile building blocks than the large zoo of combinator methods to be memorized/googled, though more verbose since some methods require multiple lines of code (and sometimes block expressions) to replicate.
- Will a reader know what the code means? Right now I understand. I don't know if I would've got it right away if I saw this code a week ago, before reading your comment, when I was mostly writing C++.

```rust
            PollDescriptorsFlow::XRun => {
                stream
                    .channel
                    .prepare()
                    .err()
                    .map(Into::into)
                    .map(&mut *error_callback);
                continue;
            }
```

Objections:

- `Into<T>` is unspecified but inferred from the input type of `error_callback`, in a separate `map` call. I dislike bidirectional type inference when it "feels" this nonlocal (across separate method calls in a chain) and doesn't have corresponding clarity benefits (in my judgment).
- I prefer a closure over chaining two `.map()` calls, which also makes rustfmt break this code into far too many lines.
- I dislike using `.map()` for control flow where you depend on side effects. `Option::map` is eager. Had this been an iterator instead, `map` would be lazy and `error_callback` would never run. Personally I prefer to avoid relying on side effects of mapping, so readers won't have to rely on these memorized details to understand the code's behavior.
	- The [`trait Iterator::map` docs](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.map) say "If you’re doing some sort of looping for a side effect, it’s considered more idiomatic to use `for` than `map()`". I'm unsure if this advice applies to `Option` or not; the [`Option::map` docs](https://doc.rust-lang.org/core/option/enum.Option.html#method.map) don't mention it.
- `.map(error_callback)` fails to build, since it moves rather than reborrowing. I don't know why this happens. This is another reason I dislike implicit reborrowing (I wish you could type `.map(mut error_callback)` to reborrow).

est31 (approved):
Who cares really it's good one way or another. I'll merge this.

PR #635, opened 2022-01-27
Bump cpal crate version to 0.13.5

enfipy:
Is there anything that blocks from bumping the `cpal` version? We are struggling with `ndk-glue` collision errors, and it would be great to release a new `cpal` version to get rid of them.


est31 (approved):
Thanks!

PR #638, opened 2022-02-11
fix: Android - Return Device(None) as a default i/o device

alula:
In some cases, device with index 0 is NOT the default output device - in my case it happens to be the ear speaker. Returning Device(None) there makes Oboe choose the default input/output device - everything works as expected.


alula:
any feedback?

est31:
Is there some way of finding out the default i/o device?

est31:
Sounds good enough to me. Merged.

PR #641, opened 2022-02-15
android: Migrate from `ndk-glue` to `ndk-context`

MarijnS95:
`ndk-glue` suffers one fatal flaw: it's "only" supposed to be used by the crate providing `fn main()` and only supposed to end up in the dependency graph once as it has `static` globals which get duplicated across versions.

In the current case with `winit 0.26` still on `ndk-glue 0.5` but `cpal` on `ndk-glue 0.6` it'll always panic in `fn native_activity()` as the `static` globals on this version is not initialized.

Introducing `ndk-context`: a crate that holds these `static`s, with the intention/premise to not see a breaking release /ever/ and make this a problem of the past.  The crate is currently initialized with the VM and Android Context on `ndk-glue` 0.5.1 and 0.6.1 (0.4.1 pending) making it compatible with whatever is current, and the possibility for backporting to older `ndk-glue` versions too.

See also:
https://github.com/rust-windowing/android-ndk-rs/issues/211
https://github.com/rust-windowing/android-ndk-rs/pull/223

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,6 +20,9 @@ hound = "3.4"
 ringbuf = "0.2"
 clap = { version = "3", default-features = false, features = ["std"] }
 
+[target.'cfg(target_os = "android")'.dev-dependencies]
+ndk-glue = "0.6"
+
 [target.'cfg(target_os = "windows")'.dependencies]
 winapi = { version = "0.3", features = ["audiosessiontypes", "audioclient", "coml2api", "combaseapi", "debug", "devpkey", "handleapi", "ksmedia", "mmdeviceapi", "objbase", "profileapi", "std", "synchapi", "winbase", "winuser"] }
 asio-sys = { version = "0.2", path = "asio-sys", optional = true }
@@ -55,7 +58,7 @@ web-sys = { version = "0.3.35", features = [ "AudioContext", "AudioContextOption
 [target.'cfg(target_os = "android")'.dependencies]
 oboe = { version = "0.4", features = [ "java-interface" ] }
 ndk = "0.6"
-ndk-glue = "0.6"
+ndk-context = "0.1"
 jni = "0.19"
 
 [[example]]
--- a/src/host/oboe/android_media.rs
+++ b/src/host/oboe/android_media.rs
@@ -1,10 +1,9 @@
 use std::sync::Arc;
 
 extern crate jni;
-extern crate ndk_glue;
 
 use self::jni::Executor;
-use self::jni::{errors::Result as JResult, objects::JObject, JNIEnv, JavaVM};
+use self::jni::{errors::Result as JResult, JNIEnv, JavaVM};
 
 // constants from android.media.AudioFormat
 pub const ENCODING_PCM_16BIT: i32 = 2;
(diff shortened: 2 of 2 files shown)


Cargo.toml
```diff
@@ -19,6 +19,7 @@ anyhow = "1.0.12"
 hound = "3.4"
 ringbuf = "0.2"
 clap = { version = "3", default-features = false, features = ["std"] }
+ndk-glue = "0.6"
```
est31, line 24: Why is this needed?

MarijnS95: Moving it from normal dependencies to `dev-dependencies` for `examples/android.rs`.

est31: Normal dependencies can be accessed from the examples as well.

MarijnS95: The main point is removing `ndk-glue` from the normal dependencies and replacing it with `ndk-context`.

est31 (approved):
Thanks!

est31:
Hmm wait it failed in CI:
```
error: custom attribute panicked
 --> examples/android.rs:8:35
  |
8 | #[cfg_attr(target_os = "android", ndk_glue::main(backtrace = "full"))]
  |                                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = help: message: `"ndk-glue"` is not a valid identifier

warning: unused import: `HostTrait`
 --> examples/android.rs:6:33
  |
6 | use cpal::traits::{DeviceTrait, HostTrait, StreamTrait};
  |                                 ^^^^^^^^^
  |
  = note: `#[warn(unused_imports)]` on by default

warning: `cpal` (example "android") generated 1 warning
error: could not compile `cpal` due to previous error; 1 warning emitted
```

MarijnS95:
It seems that `target.'cfg(target_os = "android")'` doesn't work for `dev-dependencies`? Removing it makes the build succeed again (but breaks other targets for obvious reasons). https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#development-dependencies contains such an example with `cfg()` for `dev-dependencies`, any idea what I'm missing?

MarijnS95:
This looks to be an odd issues inside our macros, investigating...

est31:
This might be because cpal is still on the 2015 edition? The macro might be emitting a wrong span that points to cpal's sources so rustc treats it as edition 2015 code.

MarijnS95:
https://github.com/bkchr/proc-macro-crate/pull/15

I'll think about removing the erroneous fallback in `ndk-macro` too.

MarijnS95:
@est31 the fix to `proc-macro-crate` has been merged and released, re-running the CI should now succeed!

(The workaround has also been removed from `ndk-macro`, but since it's not critical I haven't bothered to create a release yet)

MarijnS95:
Well, I'm not sure how this happened: I made sure to test the fix locally, yet it seems we're now hit by https://github.com/bkchr/proc-macro-crate/pull/16... Not sure how I got it working without that at some point.

est31:
@MarijnS95 it's alright, take your time. I'm impressed by how quickly you figured out the cause of the problem and came up with a solution!

MarijnS95:
@est31 I just don't like making PRs (to `proc-macro-crate`) that are "half finished" and didn't actually fix the issue, shows that my testing methodology was/is messed up somewhere :flushed: 

Anyway, the second fix is merged and released now too. I double checked it locally and the CI should not fail on it now :crossed_fingers:

MarijnS95:
Looks like we're all set now, the current issues also appear on `master`. Perhaps these CI steps should be made infallible so that it looks a bit nicer in the git log, instead of seeing everything failed?

est31:
Thanks again, please write if you want a release. As for the CI failures, yeah maybe the three failling jobs should be commented out or be allowed to fail. Ideally at least the windows job would be fixed tho. PRs welcome :).

MarijnS95:
No need for a release from my side yet, just propagating our `android-ndk-rs` changes through the most popular crates in the ecosystem. Perhaps @zarik5?

zarik5:
I don't have any hurry, I'm keeping my cpal and oboe forks for my project. But once both this PR and the oboe one hit a new release I will be able to remove some duplicate code, thanks to `ndk_context::initialize_android_context()` (my project can't use ndk-glue yet)

PR #642, opened 2022-02-18
Rust 2021 and clap 3.0

remilauzier:
The api between clap 3.0 and 3.1 as change and make the test uncompilable.


est31 (approved):
Thanks! Ideally upgrading to 3.1 would be better though. Also, why does clap break its api in a non semver breaking number release.

PR #648, opened 2022-02-27
fix: follow default output source on mac osx

kaihagseth:
When playing out on default output source, respect macOS output device selection.

See issue posted in rodio:
https://github.com/RustAudio/rodio/issues/327


src/host/coreaudio/macos/mod.rs
```diff
@@ -421,7 +421,11 @@ struct StreamInner {
 }
 
 fn audio_unit_from_device(device: &Device, input: bool) -> Result<AudioUnit, coreaudio::Error> {
-    let mut audio_unit = AudioUnit::new(coreaudio::audio_unit::IOType::HalOutput)?;
+    let output_type = match device.is_default && !input {
```
est31: Can this be an if/else instead?

est31 (changes requested):

src/host/coreaudio/macos/mod.rs
kaihagseth: Indeed! Still pretty unfamiliar with rust. Changed this and ran cargo fmt as well.

PR #651, opened 2022-03-09
PipeWire host

m00nwtchr:


Cargo.toml
```diff
@@ -33,6 +33,8 @@ nix = "0.23"
 libc = "0.2.65"
 parking_lot = "0.11"
 jack = { version = "0.8", optional = true }
+# pipewire = { version = "0.4", optional = true }
+pipewire = { git = "https://gitlab.freedesktop.org/pipewire/pipewire-rs", optional = true }
```
ishitatsuyuki, line 41: It probably doesn't matter since a git dependency can't be included in a published crate anyway, but you should suffix it the URL with `.git` so GitLab doesn't complain about that.

src/platform/mod.rs
```diff
@@ -460,10 +460,23 @@ mod platform_impl {
         SupportedOutputConfigs as JackSupportedOutputConfigs,
     };
 
-    #[cfg(feature = "jack")]
+    #[cfg(feature = "pipewire")]
```
ishitatsuyuki, line 500: See #653 for an upcoming change that allows you to simplify this.

ishitatsuyuki (commented):
Some preliminary comments.

After digging a bit into `pipewire-rs` it seems that it still lacks a lot of proper binding for essential interfaces, including device enumeration. So I guess the backend will also be blocked on that.

I also look forward to making some MRs to `pipewire-rs` as a part of a different personal project, so hopefully this will be unblocked soon.

PR #653, opened 2022-03-18
platform: Conditional impl_platform_host based on feature flags

ishitatsuyuki:
With the upcoming PipeWire backend the macros for Linux start to look a bit messy (O(2^n) scaling), so let's implement a solution that fix that clutter.


est31 (approved):
Thanks!

PR #655, opened 2022-03-24
Fix example typos

sabjorn:
both feedback.rs and record_wav.rs have "bleep" as their application name when constructing `clap::Command`. This commit changes the constructor argument for each example to match their filenames.


est31 (approved):
Thanks!

sabjorn:
@est31 no problem :D

Issue #49, opened 2015-07-15
Expose raw platform-specific data structures


cmr:
On Linux, for example, I can use the ALSA fd for async operation with mio, allowing me to have a single thread listening for commands to switch voices etc and also feed samples.

roderickvd:
We're talking about if and how to do this in #1334

PR #659, opened 2022-04-25
WebAudio: expose AudioContext in WebAudioStream

Rodrigodd:
On Chrome, If a `AudioContext` is created before a user interaction, it starts in `suspended` state. To resume it, I must call `AudioContext.resume()`. For more information: https://developer.chrome.com/blog/autoplay/

To allow that, this PR exposes the `AudioContext` of `WebAudioStream`.

Related to #49.

### DIFF
--- a/src/host/webaudio/mod.rs
+++ b/src/host/webaudio/mod.rs
@@ -350,6 +350,14 @@ impl DeviceTrait for Device {
     }
 }
 
+impl Stream {
+    /// Return the [`AudioContext`](https://developer.mozilla.org/docs/Web/API/AudioContext) used
+    /// by this stream.
+    pub fn audio_context(&self) -> &AudioContext {
+        &*self.ctx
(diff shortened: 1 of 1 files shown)


est31:
I wonder if a manual extension trait that does resuming for you would be better.

Rodrigodd:
@est31  I have first added a `resume` method, but that felt a little ad-hoc, and this implementation is more comprehensive. And calling `stream.audio_context().resume()` is not worse than `stream.resume()`. And with this implementation I don't need to decide if the method should return the `Result<JsPromise, ...>` that `resume` returns.

I also think in implementing a extension trait, like I have seen in winit, but I avoided it because I don't find this already being used in this codebase. And if I could implement this trait directly to `Stream`, instead of `WebAudioStream` as I have done here, my other PR (#660) would not be necessary for me anymore.

PR #660, opened 2022-04-25
Expose underlying platform specific implemenations

Rodrigodd:
The conversion between platform specific structs to platform-agnostic ones is already implemented, but the reverse path does not exist.

To allow that, this PR exposes `DeviceInner`, `DevicesInner`, `HostInner` and `StreamInner` (structs that already existed before), by the methods `as_inner`, `as_inner_mut` and `into_inner`.

Other possible implementation would be replacing each struct by its own inner (like replace `Device` by `DeviceInner`), but this was not possible because of the `Stream` struct that contains a seconds field.

Another one would be to make the inner field public (i.e. field `0` in the current implementation), but this may guarantee better documentation and forward compatibility. And, to be honest, I only think of that possibility while writing this comment.

If you prefer a different implementation, I can change it.

### DIFF
--- a/src/platform/mod.rs
+++ b/src/platform/mod.rs
@@ -85,28 +85,29 @@ macro_rules! impl_platform_host {
             )*
         }
 
-        enum DeviceInner {
+        pub enum DeviceInner {
             $(
                 $(#[cfg($feat)])?
                 $HostVariant(crate::host::$host_mod::Device),
             )*
         }
 
-        enum DevicesInner {
+        pub enum DevicesInner {
             $(
                 $(#[cfg($feat)])?
                 $HostVariant(crate::host::$host_mod::Devices),
             )*
         }
 
-        enum HostInner {
+
+        pub enum HostInner {
             $(
                 $(#[cfg($feat)])?
                 $HostVariant(crate::host::$host_mod::Host),
             )*
         }
 
-        enum StreamInner {
+        pub enum StreamInner {
             $(
                 $(#[cfg($feat)])?
                 $HostVariant(crate::host::$host_mod::Stream),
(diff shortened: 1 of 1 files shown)


src/platform/mod.rs
```diff
@@ -138,6 +142,44 @@ macro_rules! impl_platform_host {
...
+        impl Devices {
+            /// Returns a reference to the underlying platform specific implementation of this
+            /// `Devices`.
+            pub fn as_inner(&self) -> &DevicesInner {
+                &self.0
+            }
+
+            /// Returns a mutable reference to the underlying platform specific implementation of
+            /// this `Devices`.
+            pub fn as_inner_mut(&self) -> &DevicesInner {
```
est31: This doesn't expose a mut.

src/platform/mod.rs
```diff
@@ -138,6 +142,44 @@ macro_rules! impl_platform_host {
...
+        impl Device {
+            /// Returns a reference to the underlying platform specific implementation of this
+            /// `Device`.
+            pub fn as_inner(&self) -> &DeviceInner {
+                &self.0
+            }
+
+            /// Returns a mutable reference to the underlying platform specific implementation of
+            /// this `Device`.
+            pub fn as_inner_mut(&self) -> &DeviceInner {
```
est31: same here

src/platform/mod.rs
```diff
@@ -148,6 +190,42 @@ macro_rules! impl_platform_host {
...
+
+            /// Returns a reference to the underlying platform specific implementation of this
+            /// `Host`.
+            pub fn as_inner(&self) -> &HostInner {
+                &self.0
+            }
+
+            /// Returns a mutable reference to the underlying platform specific implementation of
+            /// this `Host`.
+            pub fn as_inner_mut(&self) -> &HostInner {
```
est31: and here

src/platform/mod.rs
```diff
@@ -148,6 +190,42 @@ macro_rules! impl_platform_host {
...
+        impl Stream {
+            /// Returns a reference to the underlying platform specific implementation of this
+            /// `Stream`.
+            pub fn as_inner(&self) -> &StreamInner {
+                &self.0
+            }
+
+            /// Returns a mutable reference to the underlying platform specific implementation of
+            /// this `Stream`.
+            pub fn as_inner_mut(&self) -> &StreamInner {
```
est31: and here

est31 (changes requested):
I have changed my mind, I think exposing the inner type is okay.

est31 (approved):
Thanks!

Issue #667, opened 2022-05-29
Expose max buffer size for WASAPI


sockmaster27:
When a stream is built with an unknown buffer size, it would be very useful to get at least an upper limit, so that any intermediate buffers can be constructed up front. The WASAPI host already appears to have access to such a value through the [`IAudioClient::GetBufferSize`](https://docs.microsoft.com/en-us/windows/win32/api/audioclient/nf-audioclient-iaudioclient-getbuffersize) method. However, this is only available after the client has been initialized, which requires some information from the config, which is only available during the call to `build_output_stream`.

As I see it, these are the possible ways that this could be achieved:
1. Putting this value in the `OutputCallbackInfo`. I don't think this would be optimal, as it would force some setup logic into the callback, likely behind an `if first_run` or something.

2. Either obtaining an `IAudioClient2` or higher that exposes the [`IAudioClient2::GetBufferSizeLimits`](https://docs.microsoft.com/en-us/windows/win32/api/audioclient/nf-audioclient-iaudioclient2-getbuffersizelimits) before it is initialized, or initializing the `IAudioClient` early, and using the `IAudioClient::GetBufferSize` method. 
Both of these require a [`WAVEFORMATEX`](https://docs.microsoft.com/en-us/windows/win32/api/mmeapi/ns-mmeapi-waveformatex), which needs some info from the config (channels, sample rate, etc.). One solution would be simply feeding it back its own list of supported configs and finding the biggest value, but this seems a bit inefficient and imprecise, since a more appropriate value could have been gotten by knowing the actual configuration.

3. Altering the API by breaking up the `build_output_stream` method. One part initializes the client and returns the max buffer size, allowing the intermediate buffers to be constructed, and another part takes the callback closure with ownership of the buffers. These could even coexist with the current method.

chipnertkj:
I'm bumping this, as there is still no API for getting the actual buffer size used by a stream.
Adding to the issue, WASAPI does not respect `BufferSize::Fixed` values set through a `StreamConfig`. AFAIK, WASAPI then creates a buffer that's *at least* the requested size. You're out of luck if you need to know what size that is before playback.
Right now the only way to know is to check the length of the buffer supplied by the data callback of a stream. This, as mentioned in the original post, is not ideal. Whether this is even true for all hosts is not clarified in the documentation.

I'm not sure if any other hosts behave like this, but the library would benefit from some kind of support for this behavior.

roderickvd:
[...]

Issue #616, opened 2021-11-15
Plans for windows-rs?


Josh015:
I just wanted to inquire if there is currently any long-term plans for porting `cpal` to the `windows-rs` crate, and if anyone has already undertaken the effort?

est31:
Not going to happen until windows-rs has a better cross compilation story.

Be-ing:
@HEnquist has been working on safe Rust bindings for WASAPI: https://github.com/HEnquist/wasapi-rs. This could be used to replace the winapi based WASAPI backend in cpal.

james7132:
Seems like this has been resolved on `windows-rs`'s end: https://github.com/microsoft/windows-rs/issues/1799

LastExceed:
can be closed, we're using windows-rs now

Issue #544, opened 2021-02-21
Unable to change BufferSize on Windows.


CaseyB:
I'm running into latency issues on Windows. When I iterate over the `SupportedStreamConfig`s I only get one option:
```
SupportedStreamConfigRange { channels: 4, min_sample_rate: SampleRate(48000), max_sample_rate: SampleRate(48000), buffer_size: Unknown, sample_format: F32 }
```
And when I try to change the buffer size to anything other than `BufferSize::Default` it tells me
```
The requested stream configuration is not supported by the device.
```
I'm kind of at a loss for where to go from here.

downfall85:
Hello CaseyB,

I assume that you try to use WASAPI on Windows ? Right now, you can't choose the buffer size with this host, it's not implemented. By default, cpal try to get the smalest possible buffer. There is already on issue for that (https://github.com/RustAudio/cpal/issues/534). But you have to know that with WASAPI, you can't really choose the buffer size. You can ask for a buffer size and WASPI will try to give it to you but you have no warranty. you can check here : https://docs.microsoft.com/en-us/windows/win32/api/audioclient/nf-audioclient-iaudioclient-initialize (**If the call succeeds, the method allocates a buffer that is a least this large**). I did some test and the buffer size may vary over time. If you need precise buffer size, you should try Asio host. By the way, if you succed, I am interested since I can't get it to work (https://github.com/RustAudio/cpal/issues/539).

YPares:
In such case, is there a way to know ahead of time the buffer size that we will receive?

roderickvd:
Fixed buffer size support, as well as querying the current buffer size, has since been implemented.

Issue #534, opened 2021-01-22
[WASAPI] Using fixed buffer size when the only available output config shows 'Unknown' buffer size ?


the-drunk-coder:
I was trying to set a fixed buffer size on Windows/WASAPI, but as a result I only get the message that the device doesn't support this configuration, no matter which buffer size I try.

When checking the available configurations, the buffer size is set to `Unknown`. Does that mean it's generally impossible to set a fixed buffer size in this situation? Or does cpal just check against the available config and, seeing that there is none that supports the requested size, returns an error ?

I'm not super familiar with the windows audio world, so I'm not sure what the WASAPI capabilities are ...

calvinkosmatka:
I'm running into issues with this as well. It doesn't look like CPAL currently supports Fixed buffer sizes on WASAPI. It looks like WASAPI itself supports fixed buffer sizes, so this _should_ be possible to implement. Setting aside not being able to choose a buffer size, I'm not even able to get a constant buffer size - on my machine, for an output stream the first callback expects 1056 samples per channel, and every callback after that wants 480 per channel.

roderickvd:
The `Unknown` buffer size with no guidance: see #795 for how that was fixed.

`BufferSize::Fixed` being rejected has since also been improved on `master` by validating the requested buffer size.

What remains is that the first output callback will `max_frames_in_buffer` instead of `period_frames`. We can consider pre-filling the WASAPI ring buffer with silence to prevent that.

PR #669, opened 2022-06-03
Implement BufferSize::Fixed in wasapi, and migrate wasapi to officially-supported Windows crate

Osspial:
I migrated to `windows-rs` in this PR since exposing the buffer size range for hardware-offloaded audio processing requires `IAudioClient2`, which isn't exposed in the `winapi` crate, and `winapi` seems to be more or less unmaintained at this point.

Regarding fixed buffer sizes - `wasapi` doesn't seem to support requesting *exact* audio buffer sizes, but will always get an audio buffer that is at least the size requested. Should that be mentioned in the documentation? Also, the buffer exposed to the `data_callback` doesn't always match the size of the buffer requested, since `wasapi` incrementally updates the buffer as audio gets played, instead of updating it all at once once the buffer is cleared. I'm not sure if that's standard behavior or not but it may also be worth documenting.

Related to #544, #534, #667, and #616. @est31, you'd mentioned at https://github.com/RustAudio/cpal/issues/616#issuecomment-969904220 that `windows-rs` had some issues with cross-compilation, but from what I can tell from skimming their issue tracker the cross-compilation issues have been addressed. Are there other issues there that are unresolved?

### DIFF
--- a/src/host/wasapi/device.rs
+++ b/src/host/wasapi/device.rs
@@ -858,8 +858,8 @@ impl PartialEq for Device {
             // 16-bit null-terminated comparison.
             let mut offset = 0;
             loop {
-                let w1: u16 = *id1.0 .0.offset(offset);
-                let w2: u16 = *id2.0 .0.offset(offset);
+                let w1: u16 = *(id1.0).0.offset(offset);
+                let w2: u16 = *(id2.0).0.offset(offset);
                 if w1 == 0 && w2 == 0 {
                     return true;
                 }
--- a/src/host/wasapi/device.rs
+++ b/src/host/wasapi/device.rs
@@ -197,38 +197,6 @@ pub unsafe fn is_format_supported(
     client: &Audio::IAudioClient,
     waveformatex_ptr: *const Audio::WAVEFORMATEX,
 ) -> Result<bool, SupportedStreamConfigsError> {
-    /*
-    // `IsFormatSupported` checks whether the format is supported and fills
-    // a `WAVEFORMATEX`
-    let mut dummy_fmt_ptr: *mut Audio::WAVEFORMATEX = mem::uninitialized();
-    let hresult =
-        audio_client
-            .IsFormatSupported(share_mode, &format_attempt.Format, &mut dummy_fmt_ptr);
-    // we free that `WAVEFORMATEX` immediately after because we don't need it
-    if !dummy_fmt_ptr.is_null() {
-        CoTaskMemFree(dummy_fmt_ptr as *mut _);
-    }
-
-    // `IsFormatSupported` can return `S_FALSE` (which means that a compatible format
-    // has been found), but we also treat this as an error
-    match (hresult, check_result(hresult)) {
-        (_, Err(ref e))
-            if e.raw_os_error() == Some(AUDCLNT_E_DEVICE_INVALIDATED) => {
-            audio_client.Release();
-            return Err(BuildStreamError::DeviceNotAvailable);
-        },
-        (_, Err(e)) => {
-            audio_client.Release();
-            panic!("{:?}", e);
-        },
-        (Foundation::S_FALSE, _) => {
-            audio_client.Release();
-            return Err(BuildStreamError::StreamConfigNotSupported);
-        },
-        (_, Ok(())) => (),
-    };
-    */
-
     // Check if the given format is supported.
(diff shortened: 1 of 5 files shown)


est31:
I haven't tried cross compiling to windows yet, but looking at windows-rs CI there seems to be no cross compilation job.

Their [contributing guide](https://github.com/microsoft/windows-rs/blob/master/.github/contributing.md) doesn't mention targetting the windows platform from Linux either.

I don't know whether they really support that feature with the love it deserves.

est31:
I've filed an issue, let's see how they react: https://github.com/microsoft/windows-rs/issues/1799

est31:
Okay, I guess we can switch to windows-rs. If there's an issue we can always revert the switch. There are some conflicts for the PR.

Osspial:
Merge conflicts have been resolved.

est31:
Can you resolve them via rebase instead? It would be a pity to squash this.

est31:
Wait no sorry disregard, it's a history that needs squashing anyways I think.

Osspial:
Does this need any more revisions, or is it good to merge?

est31:
Stuff like this makes me worried tbh: https://github.com/omnisharp/omnisharp-vscode/issues/5276

Osspial:
That's... fair! But I think this case is materially different - if this PR were pulling in a more substantial Microsoft product I'd definitely share your concern, but they've consistently provided open access to the Windows C/C++ headers for multiple decades, even during the peak of their anti-competitive practices in the 2000s. The `windows` crate is more or less an official extension of that into the Rust ecosystem, and even with an exceedingly uncharitable view of Microsoft it doesn't square with their past behavior regarding the C headers for them to lock that down.

src/host/wasapi/device.rs
```diff
@@ -295,23 +241,24 @@ pub unsafe fn is_format_supported(
 
 // Get a cpal Format from a WAVEFORMATEX.
 unsafe fn format_from_waveformatex_ptr(
-    waveformatex_ptr: *const mmreg::WAVEFORMATEX,
+    waveformatex_ptr: *const Audio::WAVEFORMATEX,
+    audio_client: &Audio::IAudioClient,
 ) -> Option<SupportedStreamConfig> {
     fn cmp_guid(a: &GUID, b: &GUID) -> bool {
-        a.Data1 == b.Data1 && a.Data2 == b.Data2 && a.Data3 == b.Data3 && a.Data4 == b.Data4
+        a.data1 == b.data1 && a.data2 == b.data2 && a.data3 == b.data3 && a.data4 == b.data4
```
est31: while you are changing this, this could use `(a.data1, a.data2, ...) == (...)`

src/host/wasapi/device.rs
```diff
@@ -968,32 +842,24 @@ impl PartialEq for Device {
...
-                panic!("cpal: GetId failure: {}", rc1)
-            }
+            let id2 = other.device.GetId().expect("cpal: GetId failure");
             let id2 = IdRAII(id2);
             // 16-bit null-terminated comparison.
             let mut offset = 0;
             loop {
-                let w1: WCHAR = *id1.0.offset(offset);
-                let w2: WCHAR = *id2.0.offset(offset);
+                let w1: u16 = *id1.0 .0.offset(offset);
```
est31: I'm not sure about this spacing here.

est31 (changes requested):
Looks mostly good, thank you!

est31:
@Osspial you are making good points. Let's give them a chance. If they annoy folks enough we can always switch back.

src/host/wasapi/device.rs
Osspial: That seems to be due to https://github.com/rust-lang/rustfmt/issues/4543. It's fixed if you put `version = "Two"` in `rustfmt.toml`, but that option doesn't actually work on stable `rustfmt`, so I've revised these to work around that (see https://github.com/RustAudio/cpal/pull/669/commits/e89f4d5697bafe5eeeafddd87f4f6b4786adf5b2).

Osspial:
Is this good to merge?

est31:
@Osspial can you resolve the merge conflicts? After that it should be good to merge.

Osspial:
I'm not seeing any merge conflicts on my end.

est31 (approved):

PR #673, opened 2022-06-05
Update dependencies, switch to once_cell

### LINKED ISSUE #670: Migrate from lazy_static to once_cell
`lazy_static` involves macros, takes a bit longer to compile, and `once_cell` provides more flexible ways for lazy initialization, and is on the way to [being added to `std`](https://github.com/rust-lang/rust/issues/74465). Replace all usages of `lazy_static` with ` static VAR: Lazy<T>`.

One potential drawback is that `lazy_static` supports spin-waiting on the result, while `Lazy` doesn't. I don't think any of crate's usages are perf-bound by this though.

james7132:
Updates all of the dependencies that are currently outdated and switches lazy_static to once_cell. Fixes #670.


est31:
All of windows, linux and mac OS fail in CI due to errors that look very like they originated from a botched update.

james7132:
The enscripten failures are hard to explain here. Not sure why it's failing so readily.

est31:
Don't worry about emscripten, as it is broken on master as well. Windows being broken is more of an issue, but it's not been broken by this PR and it doesn't seem to make the situation any worse. I'll merge this now, thanks for the PR!

est31:
Also, PRs to fix windows and emscripten are absolutely welcome.

Issue #662, opened 2022-05-14
way to get i16 input even when it is not supported?


droundy:
There is code to convert from `f32` to `i16`, but I don't see how to write code that needs to work with `i16` in a portable way, that uses straight `i16` input where available, and otherwise performs a conversion.  Perhaps this already exists?

LastExceed:
I would argue that this is beyond cpal's scope, as it is not something that needs to be abstracted across different backends, but rather something you would build on top of cpal

roderickvd:
I'm not sure which "code to convert from `f32` to `i16`" you mean?

I used to agree with @LastExceed from the perspective of "cpal abstracts over system APIs". From the perspective of "cpal" as cross-platform abstraction however, this was discussed not so long ago [over at Discord](https://discord.com/channels/590254806208217089/672897096826748948/1525504566169436260):

> I guess we could add a `build_output_stream_converting::<T, D, E>()` that takes a `T` of the caller's choosing, and have cpal use `default_*_config()` internally to determine the device's default sample format and convert to that.

`dasp_sample` makes it easy so it wouldn't be hard to add or maintain, just a question of responsibility.

LastExceed:
> I'm not sure which "code to convert from f32 to i16" you mean?

Probably the `FromSample` / `ToSample` trait impls

Issue #488, opened 2020-10-18
Is there any reason for not using dasp-sample ?


dheijl:
cpal provides its own sample conversions, with limitations like no I24 support.
dasp-sample does all this and more, so why duplicate something that already exists?

mitchmindtree:
I kicked off `dasp` (originally `sample`) quite a while after the `cpal` project began and before the `cpal` crate was moved under the `RustAudio` umbrella.

`dasp` used to be a single crate, with quite a few modules unrelated to `cpal`'s needs, however since `dasp` was recently split into sub-crates (e.g. `dasp-sample`, `dasp-slice`, etc) I think it makes more sense to consider `dasp-sample` for the `Sample` trait abstraction.

dheijl:
OK Thanks for the explanation! I was thinking of I24 sample format support.

Be-ing:
It would be great to refactor cpal to use dasp wherever possible. I think using dasp traits in this foundational library could go a long way into unifying the ecosystem around dasp so libraries are easily interoperable.

Be-ing:
@dheijl could you reopen this? Your question was answered, but the refactoring has not been done yet.

Issue #396, opened 2020-04-25
[Question] turn a `Sample` to &[u8]?


Geobert:
Hi, 

I'm using `master` branch on `624ba659da7ed97cd973d98fe1c0eae331a6f84c` so I can follow the `examples` folder :). I'm trying to send a `Sample` over TCP using `tokio` and I'm reaching a point where I want to have a `&[u8]` out of a `cpal::Sample`. 

Is there a simple way to achieve that?

Regards,

didoloan:
`let bytes:&[u8] = unsafe { core::mem::transmute::<&[f32], &[u8]>(sample) };`

Klohger:
> `let bytes:&[u8] = unsafe { core::mem::transmute::<&[f32], &[u8]>(sample) };`

If you have a slice of samples, I would not recommend this, as directly transmuting between slices is UB if I recall, plus the size will most likely be wrong. Do something like this instead:
```
pub fn slice_to_u8_slice<T>(slice: &[T]) -> &[u8] {
    unsafe { core::slice::from_raw_parts(slice.as_ptr().cast(), core::mem::size_of_val(slice)) }
}
    let bytes = slice_to_u8_slice(&[
        0.0f32, 0.5f32, 1.0f32, 0.5f32, 0.0f32, -0.5f32, -1.0f32, -0.5f32,
    ]);
```

If you only want to convert one sample then I would probably create a trait akin to this:
```
pub trait Buf: Sized {
    fn buf(&self) -> &[u8] {
        unsafe {
            core::slice::from_raw_parts(
                core::ptr::from_ref(self).cast(),
                core::mem::size_of::<Self>(),
            )
        }
    }
}
impl Buf for u8 {
    fn buf(&self) -> &[u8] {
        core::slice::from_ref(self)
    }
}
impl Buf for i8 {
    fn buf(&self) -> &[u8] {
        core::slice::from_ref(unsafe { &*core::ptr::from_ref(self).cast::<u8>() })
    }
}
impl Buf for u16 {}
impl Buf for i16 {}
impl Buf for u32 {}
impl Buf for i32 {}
impl Buf for f32 {}
impl Buf for f64 {}
```
While this is something that could be implemented in cpal, I really see no reason for it.

Ralith:
Yes, this is completely out of scope for cpal, Use something like the above, or the `bytemuck` crate.

PR #690, opened 2022-09-13
Support more sample formats and endianness

kawogi:
This is a proposal for adding more sample formats and endianness-awareness to cpal. The current version has been prepared to support all byte-aligned sample types but so far I only implemented `i/u(8, 16, 32, 64)` and `f(32, 64)`. I will add more of them (`i/u24`) shortly if there is a consensus that this approach will be accepted.

Likely fixes https://github.com/RustAudio/cpal/issues/414, fixes https://github.com/RustAudio/cpal/issues/662, fixes https://github.com/RustAudio/cpal/issues/488 fixes https://github.com/RustAudio/cpal/issues/396

This PR would deprecate: https://github.com/RustAudio/cpal/pull/420, https://github.com/RustAudio/cpal/pull/531

The PR https://github.com/RustAudio/cpal/pull/689 is included.


est31:
This is nice, but first we need to get #689 merged.

kawogi:
Agreed and I'm working on it. But I wanted to collect some feedback in parallel.

kawogi:
I found a cleaner solution to solve this. I'll re-start with another PR.

PR #696, opened 2022-10-03
add timeout paramter to DeviceTrait stream builders

mbodmer:
This makes it possible for the library client to handle situations where the host takes too long to respond or even blocks forever, eg. a retry can be applied.

if the timeout passed is "None" the current behaviour is applied, eg. blocking forever. if the timeout passed is a Duration then it will be applied to the Device Streams. This is only implemented for ALSA, the other hosts will ignore the parameter for now.

Signed-off-by: Marc Bodmer <marc.bodmer@securiton.ch>

### DIFF
--- a/src/host/alsa/mod.rs
+++ b/src/host/alsa/mod.rs
@@ -563,7 +563,7 @@ struct StreamWorkerContext {
 
 impl StreamWorkerContext {
     fn new(poll_timeout: &Option<Duration>) -> Self {
-        let poll_timeout_i32: i32 = if let Some(d) = poll_timeout {
+        let poll_timeout: i32 = if let Some(d) = poll_timeout {
             d.as_millis().try_into().unwrap()
         } else {
             -1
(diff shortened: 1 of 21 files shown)


mbodmer:
Could please someone run it again, I would like to see whether all is ok, so that I can put this aside for my part, thanks.

est31:
I've re-ran CI, but the PR seems outdated. Can you rebase the PR?

mbodmer:
Ok and thanks, I've now rebased it, hope the merging went well

src/host/alsa/mod.rs
```diff
@@ -542,19 +555,36 @@ pub struct Stream {
...
-#[derive(Default)]
 struct StreamWorkerContext {
     descriptors: Vec<libc::pollfd>,
     buffer: Vec<u8>,
+    poll_timeout: i32,
+}
+
+impl StreamWorkerContext {
+    fn new(poll_timeout: &Option<Duration>) -> Self {
+        let poll_timeout_i32: i32 = if let Some(d) = poll_timeout {
```
est31: Can you remove the _i32 here? It's unidiomatic.

est31 (changes requested):

src/host/alsa/mod.rs
mbodmer: Thanks for the review, I've changed this now

est31 (approved):
Thanks!

PR #705, opened 2022-10-07
Fix CI failure

0323pin:


est31:
Still fails, I think you need to either run `cargo fmt` or maybe remove the trailing spaces (there are still ones remaining).

0323pin:
@est31 Fixed now.
`asmjs-wasm32-test (asmjs-unknown-emscripten)` still fails but, this has nothing to do with these changes.

est31 (approved):

est31:
> `asmjs-wasm32-test (asmjs-unknown-emscripten)` still fails but, this has nothing to do with these changes.

Yeah eventually it would be good to have that addressed too, but it didn't regress recently, so is less of an issue right now.

0323pin:
Awesome, thanks.
Now, I just need a `rodio` release using current `cpal` and I'll be able to fix the use of `rodio`-backend.

Issue #704, opened 2022-10-07
Input stream silently changes to system default input device when disconnected (macOS)


generalelectrix:
Hi there, first of all thank you SO MUCH for maintaining this library!

I'm running into what might be a bug on macOS.  I have an external USB audio device connected, and I start a stream reading from this input.  This works perfectly.  Now, I want to implement some logic to inform me if the device has disconnected/provide support for automatic re-connection.  Unfortunately, when I pull the USB cable and the device disappears, my `error_callback` is never invoked - the stream continues running, but appears to fall back to receiving input from whatever the system default input device happens to be at that moment.

Any suggestions?  Is this just an idiosyncracy of CoreAudio?

est31:
Thanks for using cpal and for reaching out. Sadly, I don't have macOS expertise and the project is also lacking a macOS expert in the maintainers, so it's likely that you won't get an answer here.

generalelectrix:
Thanks for the guidance - I might be able to take a stab at implementing something based on how other projects are handling it.  The main example I've found so far is this section of code in `rtaudio`
https://github.com/thestk/rtaudio/blob/master/RtAudio.cpp#L1907-L1916

Looks like the approach taken there is to register a callback that is triggered when the device is no longer "alive", which calls a stream close method.  If you could provide a bit of guidance as to how cpal is handling this problem in the general case (for other backends), I could at least dedicate a bit of time to experimentation and see if I can implement something.

Issue #373, opened 2020-01-31
Device Notification


msiglreith:
WASAPI or PulseAudio provide an event subscription mechanism for detecting for example arrival or removal of devices.

Be-ing:
If CPAL implements support for this, that alone would justify the cost of figuring out how to use CPAL from Mixxx which is written in C++. As far as I can tell, crossplatform audio hotplug is an unsolved problem. There were efforts to do this in PortAudio years ago but they were not finished or merged. RtAudio does not support this either.

This is crucial for live performance software. Having to restart the application when someone trips over a USB cable is very bad! (It's also a shame that we are stuck with physical connectors that are intentionally designed to be easy to unplug, which is the opposite of what we need on stage...)

Be-ing:
There are three pieces to this puzzle:
1. Detecting device disconnection. This is already implemented for most backends with the DeviceNotAvailable error.
2. Detection device connection.
3. Associating a newly connected device with a device that was unplugged before. This may be the hardest part.

hrydgard:
The most basic functionality here, but still very desirable, would be an option to automatically "follow" the OS default device when it changes. This could maybe be hidden internally so that a more generic notification mechanism doesn't need to be created, initially at least - although I guess when the new device have different sample rate or format restrictions it could get dicey... Hm.

Ralith:
+1, I'd love to be able to respond to changes in the default device. It would be especially helpful if cpal provided a way to move the callbacks between devices, so I don't have to rebuild all my inter-thread communications channels from scratch each time.

generalelectrix:
Overall this feature seems at least reasonable to implement.  Device reconnection is not very hard to solve - you're already dedicating an entire thread to audio processing, and if the device disconnects, that thread is no longer busy.  Even without a "device added" hook on the backend, once the device disconnects you can just poll the create action with the same name periodically until the device reappears.  The main trick here is getting ownership of the processing and error closures back from the processing thread.  This implementation would need to be backend-specific, since the handling of the closure is different in every case.

Support for this definitely exists in Core Audio, as the AudioUnit implementation already has methods that free the processing callback and return ownership.

[...]

PR #707, opened 2022-10-08
Implement disconnection detection for CoreAudio on MacOS.

generalelectrix:
Relates to #704 and #373.  Implements a very basic mechanism for polling to determine if the device underlying a CoreAudio-backed stream has been disconnected.  This definitely works, but is obviously a very platform-dependent solution as you have to unpack the underlying stream type to access it.

Feedback extremely welcome.  The main thing I'm dissatisfied with here is that we must poll to determine if the device has dropped to then trigger some action.

### DIFF
--- a/src/host/coreaudio/macos/mod.rs
+++ b/src/host/coreaudio/macos/mod.rs
@@ -28,7 +28,7 @@ use crate::{
     SupportedBufferSize, SupportedStreamConfig, SupportedStreamConfigRange,
     SupportedStreamConfigsError,
 };
-use std::cell::RefCell;
+use parking_lot::Mutex;
 use std::ffi::CStr;
 use std::fmt;
 use std::mem;
@@ -36,6 +36,7 @@ use std::os::raw::c_char;
 use std::ptr::null;
 use std::slice;
 use std::sync::mpsc::{channel, RecvTimeoutError};
+use std::sync::Arc;
 use std::time::{Duration, Instant};
 
 pub use self::enumerate::{
@@ -416,8 +417,8 @@ impl fmt::Debug for Device {
 struct StreamInner {
     playing: bool,
     audio_unit: AudioUnit,
-    /// Track whether the device has been disconnected.
-    disconnect_listener: Option<AudioObjectPropertyListener>,
+    /// Manage the lifetime of the closure that handles device diconnection.
+    _disconnect_listener: Option<AudioObjectPropertyListener>,
     // Track the device with which the audio unit was spawned.
     //
     // We must do this so that we can avoid changing the device sample rate if there is already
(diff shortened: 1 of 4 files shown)


generalelectrix:
Perhaps a better design might involve a customizable "on device disconnect" callback, though I would need to read up a bit on the details and pitfalls of passing closures across the FFI boundary.  It looks like there is some existing code that also handles this in the implementation of `AudioUnit` in the `coreaudio-rs` crate - basically boxing the closure, converting to a raw pointer, passing that pointer into C, ensuring the closure in C does not outlive the original injection context, then re-creating the box from the raw pointer in `AudioUnit`'s `Drop` impl.

est31:
Yeah a callback would be a dream, polling isn't nice. On the other hand, light polling, e.g. once every n seconds isn't much of an issue either.

generalelectrix:
Callback seems feasible and I think I have enough examples to make it happen!  I'll tap here for a review once I have something implemented, possibly later this evening.

generalelectrix:
Just leaving some hardware notes here for future reference.  There seemed to be a delay on the order of ~1s between unplugging the device and observing the callback fire via the console.  I did not measure this systematically.  Empirically, it looked like the input stream processor would suddenly stop calling the input handling callback as soon as the device was physically disconnected, then after a delay on the order of one second, the device disconnection message would fire and the input stream callback started getting frames again (from the current "default" device).  Testing with an old Rane TTM57SL mixer.

generalelectrix:
@est31 I implemented a convenience wrapper for registering audio object property listeners, then used that wrapper to clean up the existing (rather unsafe) mechanism for setting the sample rate.  I manually tested this and it seems to work fine.  I used this same wrapper for adding a method hook to register a callback that is called when the device is disconnected.

src/host/coreaudio/macos/mod.rs
```diff
@@ -783,12 +677,166 @@ impl Device {
...
     }
 }
 
+/// Attempt to set the device sample rate to the provided rate.
+/// Return an error if the requested sample rate is not supported by the device.
+fn set_sample_rate(
+    audio_device_id: AudioObjectID,
+    target_sample_rate: SampleRate,
+) -> Result<(), BuildStreamError> {
+    // Get the current sample rate.
```
generalelectrix, line 719: Most of this code is unchanged besides replacing the property listener calls with the helper type I created, and using a channel instead of a raw pointer/polling.

src/host/coreaudio/macos/mod.rs
```diff
@@ -799,6 +847,15 @@ impl Stream {
             inner: RefCell::new(inner),
         }
     }
+
+    /// Add or replace an existing callback that will be called if the device is disconnected.
+    pub fn set_on_disconnect<F: FnMut() -> () + 'static>(
```
morajabi: beautiful API

src/host/coreaudio/macos/property_listener.rs
```diff
@@ -0,0 +1,85 @@
...
+        }
+        self.removed = true;
+        Ok(())
+    }
+}
+
+impl Drop for AudioObjectPropertyListener {
+    fn drop(&mut self) {
+        if !self.removed {
+            self._remove().ok();
```
est31: I think `_ = ...` (or `let _ = ...`) is better than calling ok here, to ignore the error.

src/host/coreaudio/macos/property_listener.rs
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    /// Explicitly remove the property listener.
+    /// Use this method if you need to explicitly handle failure to remove
+    /// the property listener.
+    pub fn remove(mut self) -> Result<(), BuildStreamError> {
+        self._remove()
+    }
+
+    fn _remove(&mut self) -> Result<(), BuildStreamError> {
```
est31: Can you rename this to not have a leading _? That turns off the dead code lints. Maybe `remove_mut`?

est31 (changes requested):
Looks really good! Thanks for the PR :).

src/host/coreaudio/macos/property_listener.rs
generalelectrix: Heh, I'm so accustomed to "ensure every error is handled or returned" that I don't have well-formed opinions on how to ignore them 😄

Ralith:
Shouldn't this be surfaced through the existing [`StreamError::DeviceNotAvailable`](https://docs.rs/cpal/latest/cpal/enum.StreamError.html#variant.DeviceNotAvailable) in the existing error callback?

est31:
@Ralith good point. I guess that's how the other backends expose it. @generalelectrix does that fulfill your use case? If so, would it be possible to adjust the PR? I think it's better to have portable solutions. I wasn't aware of that, sorry.

generalelectrix:
@Ralith @est31 happy to refactor but might get into the weeds.  For the other backends - when a device disconnects, does the stream terminate, calling the `error_callback` a single time with `DeviceNotAvailable`?  Or does it continue to be called repeatedly?  It seems like there is some complexity in the CoreAudio case because of the notion of a "default" device and having the output follow the "default", such that there are some cases where you may not actually want to stop the stream/send `DeviceNotAvailable` if you are opting-in to that behavior.  This seems to be implemented for the output case by this PR, though not the input case:
https://github.com/RustAudio/cpal/pull/648

Ralith:
> when a device disconnects, does the stream terminate, calling the `error_callback` a single time with `DeviceNotAvailable`?

I'm not intimately familiar with how other backends handle this, but that's what my personal expectation as a user would be. If the user wants to handle that by reverting to another device, I think it's appropriate for them to do so manually, which I imagine is more portable.

generalelectrix:
@Ralith I refactored to implement the behavior discussed above.  It required introducing shared ownership via Arc/Mutex for the error callback (so it can be called from the disconnect action) and to the overall stream entity (so it can be stopped asynchronously from the disconnect callback).  For the case where the default device was explicitly opened, no disconnect action is attached to ensure we do not break the existing contract.  This may not actually be necessary since the default device type may never issue a disconnected event, but I'm currently away from home and need to scrounge up an audio input device to test the new behavior.

generalelectrix:
I do not currently have access to an external audio input device, but I was able to test the new disconnect behavior on the output side with an external display with audio.  The feature seems to work as desired: when the default output is selected, the existing behavior of tracking the output device is preserved.  However, when a specific named device is opened, disconnecting that device causes the stream to stop and the error callback to be called with `StreamError::DeviceNotAvailable`.

est31 (approved):
Looks good!

PR #712, opened 2022-10-20
Bump linux deps, simplify examples, remove nix dep

mbodmer:
Bumps most linux dependencies. Moves examples to clap 4.0 and changes them to use clap_derive. This simplifies the examples and let them focus on the real content they offer.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,13 +13,13 @@ edition = "2021"
 asio = ["asio-sys", "num-traits"] # Only available on Windows. See README for setup instructions.
 
 [dependencies]
-thiserror = "1.0.37"
-dasp_sample = "0.11.0"
+thiserror = "1.0"
+dasp_sample = "0.11"
 
 [dev-dependencies]
-anyhow = "1.0.65"
-hound = "3.5.0"
-ringbuf = "0.3.1"
+anyhow = "1.0"
+hound = "3.5"
+ringbuf = "0.3"
 clap = { version = "4.0", features = ["derive"] }
 
 [target.'cfg(target_os = "android")'.dev-dependencies]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,7 +34,7 @@ once_cell = "1.12"
 
 [target.'cfg(any(target_os = "linux", target_os = "dragonfly", target_os = "freebsd", target_os = "netbsd"))'.dependencies]
 alsa = "0.6"
-nix = "0.25"
+nix = "^0.23"
 libc = "0.2"
 parking_lot = "0.12"
 jack = { version = "0.10", optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,9 +17,9 @@ thiserror = "1.0.37"
 dasp_sample = "0.11.0"
 
 [dev-dependencies]
-ringbuf = "0.2"
 anyhow = "1.0.65"
 hound = "3.5.0"
+ringbuf = "0.3.1"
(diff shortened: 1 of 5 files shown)


mbodmer:
I don't get any clippy warnings locally. I assume it's the cargo issue here: https://github.com/rust-lang/cargo/issues/10307 workaround is to remove Cargo.lock and rebuild. Could someone with access to the build server please do this, or start a clean build.

mbodmer:
libjack was missing in the container and clippy has --all-features set, so it will certainly try to build with jack and not find it if not installed. How this could have worked before however is not clear to me.

Cargo.toml
```diff
@@ -34,7 +34,7 @@ once_cell = "1.12"
 
 [target.'cfg(any(target_os = "linux", target_os = "dragonfly", target_os = "freebsd", target_os = "netbsd"))'.dependencies]
 alsa = "0.6"
-nix = "0.25"
+nix = "^0.23"
```
est31: Which crate is blocking the nix update? I think we should either depend on an older version of nix, or a newer version.

est31 (changes requested):

Cargo.toml
mbodmer: alsa-rs has a dependency on 0.24, likely changing at some point in time. So right now we have a dependency on 0.23.1 (current CPAL) and 0.24 (current alsa-rs). To make this depend on the same version we can also change to 0.24 until alsa-rs changes this, then we will have a double dependency again. Or we give cargo a range and let it choose a common version for all if it can. The broader the range the better cargo can find a single common version. As nix is not very extensively used I think it's reasonable to support a broad range.

est31: The issue with ranges is that stuff might break on an update, which isn't really great. I'd prefer either staying on an old version of nix, or duplicating it.

mbodmer: Yeah, that's true. But it seemed the interface is very stable and thus shouldn't break that easy. But still it's a risk, fair enough. Should I set it to 0.24 as alsa-rs has? Or well, we know it's fine from 0.23.1 to 0.25, how about a range of these versions. Like: ">= 0.23.1, < 0.26" ?

est31: I think we should wait for https://github.com/diwic/alsa-rs/pull/89 for now, wait for a release, and then set it to 0.25. Alternatively, set it to whatever version the currently released alsa crate uses. If there is no release in a reasonable time frame, we can use a range instead like you suggested.

mbodmer: I've just saw, it's not using 0.24, but 0.23 so I will revert the version to 0.23.1 and wait for alsa, you are right

mbodmer:
Ok, the nix version is back to where it was. The rest has a mergable state. From my side this PR is ready to merge. Thanks for reviewing.

est31:
Currently we are blocked by: https://github.com/diwic/alsa-rs/issues/90

mbodmer:
Oh, didn't realize the mess this would start. CPAL now uses the alsa re-export of the nix::error since here: https://github.com/RustAudio/cpal/commit/16617ceadef0914ab14fc124354315afe63915fa
I did not understand the implications at that time however, just tried to fix the error I've got from the version change. See also: https://github.com/diwic/alsa-rs/issues/81#issuecomment-953689161

mbodmer:
I've now removed nix as a dependency, as It's not used anywhere anymore in CPAL.

est31 (approved):
Thanks!

mbodmer:
Thanks for merging :-) and sorry for the mess this caused, but for the good finally, imo

PR #519, opened 2020-12-09
Replace stdweb with wasm-bindgen, js-sys, and web-sys

brightly-salty:
Replaces stdweb (which is unmaintained) with a combination of wasm-bindgen, and js-sys web-sys (which are). Fixes #518.

PR #713, opened 2022-10-22
Remove stdweb in favor of js-sys, web-sys, and wasm-bindgen

### LINKED ISSUE #518: RUSTSEC-2020-0056: stdweb: stdweb is unmaintained
The stdweb crate is unmaintained, as recorded by [this RustSec advisory](https://rustsec.org/advisories/RUSTSEC-2020-0056.html). Here is the [reference issue](https://github.com/koute/stdweb/issues/403). 

Maintained alternatives include wasm-bindgen, js-sys, and web-sys.

I'm willing to look into this and see how difficult it is. I'm also willing to attempt it and open a pull request if it would be accepted.

brightly-salty:
This is the replacement PR for #519. Fixes #518 

Replaces stdweb dependency, which is unmaintained, with web-sys, js-sys, and wasm-bindgen for the emscripten OS.

This should be close to working, but I'll have to see if the test suite passes after this retry of the old PR.

### DIFF
--- a/src/host/emscripten/mod.rs
+++ b/src/host/emscripten/mod.rs
@@ -1,12 +1,8 @@
-use std::mem;
-use std::os::raw::c_void;
-use std::slice::from_raw_parts;
+use js_sys::Float32Array;
 use std::time::Duration;
-use stdweb;
-use stdweb::unstable::TryInto;
-use stdweb::web::set_timeout;
-use stdweb::web::TypedArray;
-use stdweb::Reference;
+use wasm_bindgen::prelude::*;
+use wasm_bindgen::JsCast;
+use web_sys::AudioContext;
 
 use crate::traits::{DeviceTrait, HostTrait, StreamTrait};
 use crate::{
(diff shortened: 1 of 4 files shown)


src/host/emscripten/mod.rs
```diff
@@ -223,125 +216,152 @@ impl DeviceTrait for Device {
...
+            config,
+            sample_format,
+            buffer_size_frames as u32,
         );
 
         Ok(stream)
     }
 }
 
+// TODO: Do something useful with the values, maybe through wasm_bindgen_futures
```
est31: What do you mean by values here?

brightly-salty: Sorry, I forgot to deal with that before committing. I was unsure about how to use a JS Promise and actually make sure it runs; I'll commit in a second a way using wasm-bindgen-futures to turn the JS Promise into a Rust future that we can then await and spawn. The only downside is that if the promise is rejected, we have to panic instead of actually returning an `Err`, unless we want to write a custom executor ourselves for this one usecase.

est31 (approved):
Thanks!

PR #714, opened 2022-10-23
remove nix as a dependency

mockersf:
It seems `nix` is not needed as a direct dependency, as `alsa` is re-exporting everything `cpal` needs from it


est31:
Thanks for the PR! Closed in favour of #712.

mockersf:
oh sorry for the noise, I missed that it was done in that PR

PR #732, opened 2022-12-01
Cargo.toml: drop dependency on nix

pacak:
It is not used directly, only reexported version from alsa

Direct use of `nix` crate was removed back in 214403f

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -33,7 +33,6 @@ once_cell = "1.12"
 
 [target.'cfg(any(target_os = "linux", target_os = "dragonfly", target_os = "freebsd", target_os = "netbsd"))'.dependencies]
 alsa = "0.6"
-nix = "0.25"
 libc = "0.2.65"
 parking_lot = "0.12"
 jack = { version = "0.9", optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "cpal"
-version = "0.14.1"
+version = "0.14.2"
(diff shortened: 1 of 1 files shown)


pacak:
I guess this was already done here https://github.com/RustAudio/cpal/pull/712 but as far as I can see this branch and not `master` gets published on crates.io, at least according to crate version.

est31:
Thanks but I think it's not needed. The fact that we use the nix dependency already transitively means that directly depending on it has no build time cost. It's just a beauty thing. I don't think there will be a 0.14.x release any more, new development is on 0.15.x.

pacak:
> The fact that we use the nix dependency already transitively means that directly depending on it has no build time cost. It's just a beauty thing

Well, no, not quite. Currently released version of cpal depends on nix 0.25 and alsa 0.6.0. alsa 0.6.0 depends on nix 0.23, my own code depends on nix 0.26, as a result when I compile something - it will pull and compile 3 different versions of nix: 0.23 because of alsa, 0.25 because of cpal and 0.26.

- https://crates.io/crates/cpal/0.14.1/dependencies
- https://crates.io/crates/alsa/0.6.0/dependencies

pacak:
Like so:

est31:
I see, that makes sense. I can merge your PR. Can you also update the version number of cpal in Cargo.toml to 0.14.2?

pacak:
Sure, done.

est31 (approved):
Thanks

pacak:
Thank you for quick response! :heart:

est31:
@pacak you are welcome!

PR #734, opened 2022-12-04
Update synth_tones.rs example program

novafacing:
The example for `synth_tones.rs` was slightly outdated and didn't compile for me on 0.14.2. This fixes the example so it works on the current version.

### DIFF
--- a/examples/synth_tones.rs
+++ b/examples/synth_tones.rs
@@ -6,11 +6,8 @@ extern crate anyhow;
 extern crate clap;
 extern crate cpal;
 
-use cpal::{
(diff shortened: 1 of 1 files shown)


est31:
If it fails on 0.14.2, you should maybe try the [stable-0.14](https://github.com/RustAudio/cpal/tree/stable-0.14) branch. The example is tested in CI at least whether it compiles, and it does compile, but it is adjusted for the release that will become 0.15.

novafacing:
Ah I see, I didn't realize it wasn't compiling because *I* had outdated code with 0.14.2, usually it's the opposite problem!

PR #626, opened 2021-12-24
fix: ASIO Segfault on Windows

robot-rover:
# Commit Description

This commit fixes a bug that occurs in the ASIO host on windows. It
commonly manifests in a message about Data Execution Prevention.
From what I can tell, the ASIO API call ASIOCreateBuffers() expects the callbacks
struct to exist for the lifetime of the buffers (i.e. until
ASIOFreeBuffers() is called). This isn't really mentioned in the ASIO
Docs unfortunately, and I only found it by hunting for this bug.

This commit fixes the issue by making the AsioCallbacks struct a static
object rather than a local that is constructed by a function. It has to
be mutable, unfortunately, because the C api for ASIOCreateBuffers()
asks for a non-const pointer. I can't think of any reason that it would
be mutated by the driver, so I still think this is the best solution.

# Further Elaboration

I initially fixed this bug by changing the code which calls ASIOCreateBuffers to look like this:

```rust
fn create_buffers(
	&self,
	buffer_infos: &mut [AsioBufferInfo],
	buffer_size: Option<i32>,
) -> Result<c_long, AsioError> {
	let num_channels = buffer_infos.len();

	// To pass as ai::ASIOCallbacks
	let callbacks = create_asio_callbacks();
	let callbacks = Box::new(callbacks); // CHANGE

	let mut state = self.inner.lock_state();
// ... snip ...
unsafe {
		asio_result!(ai::ASIOCreateBuffers(
			buffer_infos.as_mut_ptr() as *mut _,
			num_channels as i32,
			buffer_size,
			Box::leak(callbacks) as *mut _ as *mut _, // CHANGE
		))?;
	}
	*state = DriverState::Prepared;

	Ok(buffer_size)
}
```

And it stopped giving me the access violation. I'm fairly sure this PR would address
#596 
#539 
#407

PR #737, opened 2022-12-27
Support i32 for asio

RuairidhWilliamson:
## Add `SampleFormat::I32` support for ASIO

Hi, I was testing out asio and to get it working with ASIO I had to make these changes to allow ASIO to use i32.
I think the comment explaining why `ASIOSTInt32` is mapped to `SampleFormat::I16` is not relevant any more seeing as `SampleFormat::I32` has been added.

Note:
I also had to apply https://github.com/RustAudio/cpal/pull/626 to get cpal to work without segfault.


est31 (approved):
Thanks!

est31:
FTR #626 is blocked on the response to my review comment. I've been a bit late with it but I would still love a response. `static mut` is not the greatest way to resolve this.

PR #742, opened 2023-01-22
Fix CoTaskMemFree always getting called with null

AmionSky:
It was changed incorrectly in https://github.com/RustAudio/cpal/pull/726


est31 (approved):
Thanks!

PR #750, opened 2023-02-13
make oboe's shared-stdcxx feature accessible

LeonKowarschickKenbun:
This PR exposes oboes shared-stdcxx feature from cpal. Access to this was required for us to fix builds on a couple android targets


est31:
FTR you can also depend on `oboe` directly yourself and then enable the feature for that. Given cargo's unification of dependencies, this also enables the feature for oboe used by cpal (but you have to always keep the deps in sync).

est31 (approved):

PR #760, opened 2023-03-14
switch from mach to mach2

mockersf:
crate `mach` is unmaintained and hasn't been updated in 4 years https://github.com/fitzgen/mach/issues/63

crate `mach2` is now the updated version https://github.com/JohnTitor/mach2

This works the same on my Mac 👍


est31 (approved):
Thanks!

mockersf:
Yup that's how I came here, from the rustsec warning on Bevy 🙂 

Here is the update PR with the changelog: https://github.com/RustAudio/cpal/pull/761

Issue #767, opened 2023-03-29
[WebAudioAPI] AudioDestinationNode is always stereo whatever the given config.channels is


VincentJousse:
When building an output stream with the webaudio host, cpal creates a new `AudioContext` whose `destination`'s `channel_count` is not modified accordingly with the `StreamConfig` given as parameter.

PR #768, opened 2023-03-29
webaudio: set the channel_count in the AudioDestinationNode

VincentJousse:
Fix for #767 

[WebAudio host]
When building a new output stream, it uses the `channels` field of the given `config` to set the channel count of the AudioDestinationNode.


src/host/webaudio/mod.rs
```diff
@@ -220,15 +220,21 @@ impl DeviceTrait for Device {
...
+            |err| -> BuildStreamError {
+                let description = format!("{:?}", err);
+                let err = BackendSpecificError { description };
+                err.into()
+            },
+        )?;
+
+        let destination = ctx.destination();
+
+        if config.channels as u32 <= destination.max_channel_count() {
```
est31, line 236: should we error if the channel count is larger than the supported one? I'm not sure

VincentJousse: I think the `AudioDestinationNode` will apply a downmix in this case.
I mean the `AudioBufferSourceNode`s will have the number of channels we want, and `AudioDestinationNode` will have less, so it will apply a downmix.
See https://www.w3.org/TR/webaudio/#channel-up-mixing-and-down-mixing

Either we accept this fallback (which is the current behavior), or we return an error.

It would be logical to return an error but if we do, each user who builds a multichannel stream on a stereo system will get this error, I can't tell how many they are !!

est31: @VincentJousse fair points. Could you add a comment to the source code that explains why we don't error here?

VincentJousse: Done

VincentJousse: @est31 Is everything OK ?

Can we hope a patch release for this feature ?

est31 (approved):

est31:
@VincentJousse regarding a patch release, can you make a PR like #761 to bump the version? And adding the changes that have happened since then to the changelog?

Issue #589, opened 2021-07-06
must_use on stream


samoylovfp:
I've been a happy user of cpal for playing around with sound but got bitten in my foot two times by the same behavior:
I forget to persist the stream that come from `build_input_stream` and my program just silently does nothing.

I feel like this experience can be significantly improved if Streams had "must_use" attribute

est31:
Fixed in https://github.com/RustAudio/cpal/pull/772

PR #772, opened 2023-04-04
Added must use to stream. Fixes issue #589

Easyoakland:
This fixes issue #589.
This warning makes is such that if a function returns a Stream it gives a warning if the stream is not assigned.
It also works for `Result(Stream, Box<dyn Error>)` so using the `?` operator will still give a warning about an unused stream even if the result is unwrapped.
The is an abbreviated example:
```rust
fn run<T>(device: &Device, config: &StreamConfig) -> Result<Stream, Box<dyn Error>>{
            config,
            next_value,
            err_fn,
            None,
        )?;
    stream.play()?;
    Ok(stream)
}

let main() -> Result<(), Box<dyn Error>> {
    let _stream = run(&device, &config)?; // will work and will output sound until `_stream` is dropped

    // Previously didn't do anything on success. Now it will warn that it won't do anything.
    run(&device, &config)?;
}
```

### DIFF
--- a/src/platform/mod.rs
+++ b/src/platform/mod.rs
@@ -68,7 +68,7 @@ macro_rules! impl_platform_host {
         // functions within the callback.
         //
         // TODO: Confirm this and add more specific detail and references.
-        #[must_use = "If the stream is not used it will not play."]
+        #[must_use = "If the stream is not stored it will not play."]
         pub struct Stream(StreamInner, crate::platform::NotSendSyncAcrossAllPlatforms);
 
         /// The `SupportedInputConfigs` iterator associated with the platform's dynamically
--- a/src/platform/mod.rs
+++ b/src/platform/mod.rs
@@ -68,7 +68,7 @@ macro_rules! impl_platform_host {
         // functions within the callback.
         //
         // TODO: Confirm this and add more specific detail and references.
-        #[must_use = "If the stream is not assigned it will not play."]
+        #[must_use = "If the stream is not used it will not play."]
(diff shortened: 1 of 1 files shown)


src/platform/mod.rs
```diff
@@ -68,6 +68,7 @@ macro_rules! impl_platform_host {
         // functions within the callback.
         //
         // TODO: Confirm this and add more specific detail and references.
+        #[must_use = "If the stream is not assigned it will not play."]
```
est31:
```suggestion
        #[must_use = "If the stream is not stored it will not play."]
```
I think `stored` is a better word. In theory you could also use APIs like `HashMap::insert` or `vec.push` to store the stream, neither I would call assignment.

est31 (changes requested):

src/platform/mod.rs
Easyoakland: @est31 Changed in [085fa63](https://github.com/RustAudio/cpal/pull/772/commits/085fa6347fc1a8c125a230ffde2cab4f71078e2b) to more closely match https://doc.rust-lang.org/src/std/io/error.rs.html#799

est31: I still like stored more. Can you change it to that?

Easyoakland: @est31 changed in [e158915](https://github.com/RustAudio/cpal/pull/772/commits/e158915e3cb4b656afda928477592e804a51fa43)

est31 (approved):

PR #774, opened 2023-04-06
WASM: Fix issues with `play`, `pause`, and browser autoplay blocking

DouglasDwyer:
This PR improves the behavior of audio streams on WASM in the following ways:
- **Calling `play` multiple times on a `Stream` only calls `resume` on the underlying `AudioContext` a single time,** and only spawns background audio callbacks once. Previously, calling `play` from Rust multiple times in succession would spawn multiple closure instances and lead to stuttering audio.
- Browsers block audio playback until the user interacts with the webpage, and force the `AudioContext` to remain in the `Suspended` state. This is problematic for `cpal`'s WASM backend, because from the Rust side there is no way to determine whether audio playback has begun. **This PR causes the `Stream` to read and discard audio frames while the context is blocked, so it's as if the audio stream is always playing (but muted until the user interacts) once `play` is called.** Previously, the audio would simply hang if the user had not yet interacted with the page.
- **I've added support for running `cpal` with the `+atomics` target feature enabled.** This is accomplished by adding a secondary Javascript-side array into which frames are copied before they are played, since you cannot send frames to Javascript from a `SharedArrayBuffer` directly.

I'm happy to make any improvements necessary to get this PR merged! I would like to see all of these behaviors available in `cpal`, because they are the best fit for my use case.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -52,12 +52,12 @@ coreaudio-rs = { version = "0.11", default-features = false, features = ["audio_
 wasm-bindgen = { version = "0.2.58" }
 wasm-bindgen-futures = "0.4.33"
 js-sys = { version = "0.3.35" }
-web-sys = { version = "0.3.35", features = [ "AudioContext", "AudioContextOptions", "AudioBuffer", "AudioBufferSourceNode", "AudioNode",  "AudioDestinationNode", "Window", "AudioContextState"] }
+web-sys = { version = "0.3.35", features = [ "AudioContext", "AudioContextOptions", "AudioBuffer", "AudioBufferSourceNode", "AudioNode",  "AudioDestinationNode", "Performance", "Window", "AudioContextState"] }
 
 [target.'cfg(all(target_arch = "wasm32", target_os = "unknown"))'.dependencies]
 wasm-bindgen = { version = "0.2.58", optional = true }
 js-sys = { version = "0.3.35" }
-web-sys = { version = "0.3.35", features = [ "AudioContext", "AudioContextOptions", "AudioBuffer", "AudioBufferSourceNode", "AudioNode",  "AudioDestinationNode", "Window", "AudioContextState"] }
+web-sys = { version = "0.3.35", features = [ "AudioContext", "AudioContextOptions", "AudioBuffer", "AudioBufferSourceNode", "AudioNode",  "AudioDestinationNode", "Performance", "Window", "AudioContextState"] }
 
 [target.'cfg(target_os = "android")'.dependencies]
 oboe = { version = "0.5", features = [ "java-interface" ] }
--- a/src/host/webaudio/mod.rs
+++ b/src/host/webaudio/mod.rs
@@ -440,6 +440,14 @@ impl StreamCallbackState {
             let mut temporary_buffer = vec![0f32; self.buffer_size_samples];
             let mut temporary_channel_buffer = vec![0f32; self.buffer_size_frames];
 
+            #[allow(unused_variables)]
+            let temporary_channel_array_view: js_sys::Float32Array;
+            #[cfg(target_feature = "atomics")]
+            {
+                let temporary_channel_array = js_sys::ArrayBuffer::new((std::mem::size_of::<f32>() * self.buffer_size_frames) as u32);
+                temporary_channel_array_view = js_sys::Float32Array::new(&temporary_channel_array);
+            }
+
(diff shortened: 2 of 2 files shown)


est31:
Thanks for your PR! I think the `play` changes are good at least, and having `SharedArrayBuffer` support is also really neat.

> Previously, the audio would simply hang if the user had not yet interacted with the page.

Can you explain why you prefer the new behaviour?

DouglasDwyer:
@est31 sure! Let me emphasize here that this new behavior should be **non-breaking**: it is a strict superset of the old behavior, and thus backwards compatible.
To reiterate, the current behavior is that any audio stream started before the user interacts with the page simply hangs. This remains true even after the user interacts; the audio stream continues to hang, and it's necessary to call `play` again to make it start. This means that any current users on WASM need to wait until a user interaction to call `play` on the stream.
These changes just extend this by allowing `play` to be called, and work correctly, before a user interacts. Consumers of `cpal` can still wait until user interaction to call play (which was previously required), but this gives them an extra option: they can just call `play` immediately and everything works.
For my personal use case, I'd like to embed `cpal` into a cross-platform game. It's much easier for me to begin an audio stream on startup, and not worry about web-specific nonsense like automute :)

est31:
@DouglasDwyer my question was more why you think it's a good idea, not why it's allowed for us to do it. I agree that we are allowed to do it, but I want to hear arguments why the new behaviour is better than the old one. Why is it bad if it hangs? What does hanging mean? can you still interact? Why is it better that it is muted but plays in the background?

DouglasDwyer:
Okay. My argument above was meant to state that I think it's a good idea because it gives users an additional option: they can have the audio _not hang_ if that's what they want. Surely having that option is better than not having it?

Anyway, let me try and answer your questions.

> What does hanging mean? Can you still interact?

The audio stream is not started, and no audio frames are consumed from the output callback. The application otherwise functions normally.

> Why is it bad if it hangs?

There's no way to detect the hanging behavior from Rust; currently, the `play` call still returns `Ok(())`. This strikes me as confusing - Rust crates typically use `Result`s to indicate success, so I would expect that `Ok(())` means that the stream was started.

Further, having the stream hang could lead to audio buffer overruns. If audio frames are being streamed into the application, but the output callback is never invoked, then these frames could accumulate and cause an out-of-memory issue.

> Why is it better that it is muted and plays in the background?

Because as an end user, I expect that a successful call to `play` starts the audio stream, consuming audio frames. On other platforms, this is how `play` behaves. It's cumbersome that WASM functions differently. I think that playing (muted) more closely matches how `cpal` works on native platforms, and I think that cross-platform consistency is highly desirable.

Ralith:
As a user, I'd rather (web) applications whose audio I can't hear not be spending potentially large amounts of CPU time rendering said inaudible audio.

DouglasDwyer:
@Ralith, an alternative solution that I think would be acceptable would be for `play` to return `Err` if the user hasn't yet interacted with the page. That way, at least, the Rust application could respond by dropping audio frames. Unfortunately, it does not appear to be possible to synchronously detect a stream failing to start - the Javascript AudioContext's `resume` function returns a promise.

Nek:
+1 too as I was struggling with the same problems in WASM world.

volodalexey:
Currently bevy `0.11.0` uses this library for sound.
Mobile browsers can not handle current implementation in WASM see my issue https://github.com/bevyengine/bevy/issues/9211.

+1
I would like to see this merged!

cybersoulK:
i just got this issue as well. i require +atomics for something else, but right now it's crashing on both chrome and firefox

DouglasDwyer:
Maintainers - please let me know if there is anything more that I can do to push this PR forward!

cybersoulK:
i can confirm this PR solved it

est31:
I'm still not really happy about the hang removal changes. But outside of that, I think the general idea of the PR looks good. I need to give it a closer look though.

kettle11:
Maintainers (@est31) can we find a way to get this merged? `cpal` crashing when `atomics` are enabled is blocking getting Bevy multithreaded on web.

Speak2Erase:
Any updates on this PR?

Davidster:
@Speak2Erase looks like we might be good to go since the recent changes from @kettle11! I will test on my project tonight. Thank you!!

est31:
Given #837 I'm closing this, but open for any of the non controversial changes to be submitted in a PR. Thanks.

Issue #764, opened 2023-03-25
asio-sys build failed by "cannot find ai::*"


Dicklessgreat:
I've followed the instructions in the "ASIO on Windows" section and managed to build ASIO.
But asio-sys fails to compile at src\bindings\mod.rs.
What is the current state of ASIO development? Has anyone tried to build cpal with features = ["asio"]?

"cannot find" objects are:
ai::kAsioSelectorSupported
ai::kAsioResetRequest
ai::kAsioEngineVersion
ai::kAsioResyncRequest
ai::kAsioLatenciesChanged
ai::kAsioSupportsTimeInfo
ai::kAsioSupportsTimeCode
ai::kAsioSupportsInputMonitor
ai::kAsioResetRequest
ai::kAsioResyncRequest
ai::kAsioLatenciesChanged
ai::kAsioEngineVersion
ai::kAsioSupportsTimeInfo
ai::kAsioSupportsTimeCode

mizuki0629:
sorry. I'm not good at English, so I use a translator.
It seems that bindgen cannot generate constants using enum.
bindgen used in asio-sys is old, so I think to upgrade bindgen first to fix the issue.
I'll try and see if I can fix the issue.

est31:
#775 has been merged

Issue #596, opened 2021-08-02
WIN10-ASIO 0xc0000005 Access Violation (DEP) When Both Streams Built


keeripes:
I see there are some other issues similar to this, and I have been banging my head against the wall for a couple of days now with this issue which I will outline as succinctly as possible.  

### The Error

I am getting the same behavior on two machines with different audio interfaces: 

 - when *only* an input stream is initialized I successfully get inbound audio from the audio interface.
 - when **both** input and output streams are initialized I get the access violation shown below
 - when *only* an output stream is initialized I get output audio on one machine/interface pair and the exception on the other.

The error is the same - after the stream starts I get the error:

`error: process didn't exit successfully: '...' (exit code: 0xc0000005, STATUS_ACCESS_VIOLATION)`

If I step either through with a debugger I get a bit more information:

`Exception: Exception 0xc0000005 encountered at address 0x000060: User-mode data execution prevention (DEP) violation at location 0x00000060`

*(Not sure if this is a debugger artifact or not...  I just wanted to put it here)*

This access violation exception happens randomly while stepping through in debug, leading me to think it might be caused by the ASIO bufferSwitch callback - maybe the first time it gets called on the output buffers.

I am getting this same behavior on two different machines with two different interfaces (in both `debug` and `release` builds):

 - **Machine 1** - Windows 10 (64-bit) VM on QEMU w/ USB passthrough with a MOTU UltraLiteMk3.

 - **Machine 2** - Windows 10 (64-bit) Bare metal with a Digigram VX882 PCIe card.

I should note here that both interfaces work flawlessly with REAPER on their parent machines.  I do not suspect a driver issue.  Likely a **me** issue.

### The test case - `feedback.rs`

In its current form, the `feedback.rs` example doesn't work out of the box with ASIO (on my machines), so in an effort to troubleshoot I cloned the cpal repo and started poking around.  After looking at the ASIO implementation in cpal, I saw two things that jumped out given the exception is potentially due to Data Execution Prevention.  In my feeble mind I start thinking that something has moved on the rust side, invalidating the pointers held either by the driver (audio buffers) or the ASIO callbacks on the rust side.  

The two things I see as possible pinch points (though they appear 100% correct) are:

 - the audio buffers are split apart into different Options after being initialized via ASIOCreateBuffers in (`Driver::create_streams()`)
 - when the second opposing direction stream is created, new buffers and ASIO callback pointers are created (`Driver::create_buffers()`).

[...]

Issue #539, opened 2021-02-04
Segfault on Execution, Exit Code: STATUS_ACCESS_VIOLATION


ifacodes:
When running the beep example using an ASIO Host, on windows 10, the program segfaults in unsafe code space with the exit code of `0xc0000005, STATUS_ACCESS_VIOLATION`.

I am using a Focusrite 2i2 Gen 2 as my asio host. 

The same program executes perfectly fine with Wasapi, in ASIO this segfaults occurs around 70% of the time the program is run.
I have also tested it with a rudimentary pulse wave, and triangle wave, and both of those work in Wasapi, but segfault 100% of the time in ASIO. 

I'm not too sure how to debug this myself, although I am going to try.

downfall85:
Hello @poww10s,

Have you tried to release with caro build --release and run the executable?

I have the same problem as you if I just try to do a cargo run

poww10s:
If I release with `cargo build --release` and run the executable, it run's correctly, but no sound is outputted, and it doesn't sleep, suggesting to me that something is broken with how it handles asio. I've made a ton of progress on my program, but for now I'm limited to wasapi with cpal, which is a bit of a pain

poww10s:
I feel like the issue might have something to do with my focusrite using I16, but the example using F32 as it's output. I tried changing it around manually, but I'm a bit confused about how the Sample trait handles the conversion between types.

downfall85:
I'm stuck in the step as you. Worse, I have an error when I try to restart the program a second time and I have to unplug and plug in my USB sound card. I think the execution crashes without error and the resources are not freed.

What makes you think this is related to the I16 format? I think that's how ASIO works. My sound card is also in I16 with ASIO while it works in F32 with WASPI.

I don't have too much idea how to debug this behavior. If you have any ideas, don't hesitate to share them.

poww10s:
I know someone else used a debugger and found that it's segfaulting in unsafe code, but not in any rust unsafe code, but one of the Focusrite DLL's a driver I think. (#517)

I know just from using other applications with the drivers I'm pretty sure Focusrite default is I32 format, but because cpal currently only supports I16 it converts all formats it doesn't support to the ones it does.

est31:
#775 should have fixed this, if not, please write!

Issue #407, opened 2020-05-22
feedback example with ASIO backend produces STATUS_ACCESS_VIOLATION seg fault


JoshuaBatty:
Just trying to get the feedback example working with the asio backend. Compiles and runs and then instantly seg faults. See console out put below. 

console log is:

```
Using default input device: "ASIO4ALL v2"
Using default output device: "ASIO4ALL v2"
Attempting to build both streams with f32 samples and `StreamConfig { channels: 2, sample_rate: SampleRate(44100) }`.
Successfully built streams.
Starting the input and output streams with `150` milliseconds of latency.
Playing for 3 seconds...
error: process didn't exit successfully: `target\debug\examples\feedback.exe` (exit code: 0xc0000005, STATUS_ACCESS_VIOLATION)
Segmentation fault
```

Karlinde:
I am also experiencing this when modifying the code from the example to use an explicit ASIO host and device.

I am using a Focusrite Scarlett 18i20 2nd gen with the native drivers (not ASIO4ALL) and according to the debugger this crash occurs in the thread "FocusriteUSBAsio64.dll", with the debugger message: 

```
Exception thrown at 0x0000000000000000 in asiotest.exe: 0xC0000005: Access violation executing location 0x0000000000000000
```

Oddly enough, this crash doesn't occur if I launch the program using the debugger. In that case the code runs to completion (although with some serious distortion and dropped samples in the output but that's another bug I suppose). It only seems to happen when running the program outside of the debugger, or if I start the program outside but with a sleep instruction right before the "build_input_stream..." lines so that I have time to attach a debugger, in order to get the above exception message.

est31:
#775 should have fixed this, if not, please write!

AlexanderTreml:
I experience a similar issue: Same exit code when trying to setup the ASIO feedback example on Windows 10. It might be related to the old audio interface I use (M-Audio Fast Track Pro). See the stack trace below:

```
void core::slice::specialize::impl$1::spec_fill<u8>(struct ref_mut$<slice2$<u8> >, unsigned char) (@<[T] as core::slice::specialize::SpecFill<T>>::spec_fill:40)
void core::slice::impl$0::fill<u8>(struct ref_mut$<slice2$<u8> >, unsigned char) (@core::slice::<impl [T]>::fill:8)
[...]

PR #775, opened 2023-04-08
Issue asio sys

mizuki0629:
### Build faild #764 
bindgen cant generate enum value `kAsio*.`  (` .whitelist_var("kAsio*")` )
Because they are not "var" and "type".
```c
// asio/common/asio.h
// asioMessage selectors
enum
{
	kAsioSelectorSupported = 1,	// selector in <value>, returns 1L if supported,
								// 0 otherwise
    kAsioEngineVersion,			// returns engine (host) asio implementation version,
...
	kAsioOverload,              // driver detected an overload

	kAsioNumMessageSelectors
};
``` 
So, I created enum manually on rust. Like AsioSampleType.
https://github.com/RustAudio/cpal/blob/848abf256fcedf05a05515b3570adc507859cf25/asio-sys/src/bindings/mod.rs#L118-L156
And bindgen 0.58 deprecated methods `whiltelist_*`.
update bindgen 0.56->0.64 and change `whiltelist_*` to `allowlist_*`
### Segmentation fault #596, #539, #407
#626 used `static mut` .
I change to `Lazy<Mutex`


Dicklessgreat:
Thanks! It works fine for me now.
BTW, it emits warnings at `asio-sys\src\bindings\errors.rs:42:30` and `asio-sys\src\bindings\errors.rs:79:67`, saying 
```
use of deprecated associated function `std::error::Error::description`: use the Display impl or to_string()
```
I think it's probably because it has not been builded for a long time.
Should it be fixed in another pull request? Any other things to change? who is the maintainer of `asio-sys`? @est31 @mitchmindtree

mbwilding (approved):

mbwilding:
`[dependencies]
cpal = { git = "https://github.com/mizuki0629/cpal", branch = "issue-asio-sys", features = ["asio"] }`

Thanks, compiles well!

est31 (approved):

asio-sys/src/bindings/mod.rs
```diff
@@ -481,12 +534,16 @@ impl Driver {
         if let DriverState::Prepared = *state {
             state.dispose_buffers()?;
         }
+
+        // To pass as ai::ASIOCallbacks
+        let callbacks = &mut (*ASIO_CALLBACKS.lock().unwrap());
```
est31, line 539: The issue with this is that this creates a pointer, and then the pointer lives on even though the guard has already been dropped.

Maybe we'll have to leak the memory? `Box::leak` maybe?

mizuki0629: I am not a native English speaker. I also use a translator, so the sentences may be strange.

It can happen, but the cause is closed inside the unsafe code.
Also, it is easy to identify the driver as the cause of the problem, because the problem does not occur with a properly implemented driver.
If I don't manage memory in unsafe code, I don't think I need to leak my memory.

est31: I have thought about the problem again. If the driver does not modify the callbacks, then it is not needed to have a Lazy and a Mutex. I think a non-mut `static` would be enough. @mizuki0629 can you try changing `ASIO_CALLBACKS` to just be `static ASIO_CALLBACKS: AsioCallbacks = ...;` and then creating the pointer like `&ASIO_CALLBACKS as *const _ as *mut _` ?

mizuki0629: It works. And better.

est31: @mizuki0629 that great! Can you make a PR? That would be nice.

mizuki0629: Sure.
I'm new to working on github, so it's slow.
Please wait a little.

est31: @mizuki0629 take your time! It is a free time project :). I am very thankful for your help.

PR #777, opened 2023-04-14
Remove unnecessary Lazy Mutex

mizuki0629:
## Commit Description
Remove unnecessary Lazy Mutex.
Change to non-mut static.
https://github.com/RustAudio/cpal/pull/775#discussion_r1166606279


est31 (approved):
Thanks!

Issue #727, opened 2022-11-13
Switching from windows-rs to windows-sys


est31:
There is a huge amount of churn with the windows crate, see https://github.com/microsoft/windows-rs/issues/1720, even though it is a crate that is supposed to be foundational. We can't follow each windows-rs release. They have said however that windows-sys will have breaking changes less often, and that seems to be the case indeed. Therefore, we should switch to windows-sys.

cc @Osspial who has done the winapi -> windows-rs switch.

Jasper-Bekkers:
Notice that unfortunately windows-sys doesn't really expose COM objects; instead they define the names for them as `*mut void`, see for example https://docs.rs/windows-sys/latest/windows_sys/Win32/Media/Audio/type.IAudioCaptureClient.html

MarijnS95:
`windows-sys` was mainly introduced to combat compile times, but in my experience `windows` hasn't been nearly as bad since introducing proper feature flags to _only compile what you need_.

I ended up [making the same choice in our `gpu-allocator` crate](https://github.com/Traverse-Research/gpu-allocator/pull/107) for needing access to COM and the desire to rid our code-base of error-prone pointer casts and lifetime/refcount issues.

Now that [`windows` is also promised to slow down on releases](https://github.com/microsoft/windows-rs/issues/2156#issuecomment-1318725376) this doesn't sound like an actionable issue (at all) anymore.

est31:
@Jasper-Bekkers I see that is a valid point. It is also great news that the maintainers of the windows crate have promised to slow down the release cadence, thanks @Jasper-Bekkers for filing the issue. I'm not sure what their new target is however. Anything faster than "1 semver breaking release per year" will likely be too much for the ecosystem, and I'm not sure if one semver breaking release per year would not be...

It also seems that they expect this churn to stop in the future, so I hope it is indeed a temporary problem.

I'm closing this issue as it seems to be unlikely we want to use `windows-sys`.

SkiFire13:
I found this thread before asking whether `cpal` would/had considered switching from `windows` to `windows-sys`.

> `windows-sys` was mainly introduced to combat compile times, but in my experience `windows` hasn't been nearly as bad since introducing proper feature flags to _only compile what you need_.

[...]

PR #778, opened 2023-04-18
Replace `windows` with embedded bindings

Jake-Shadle:
This PR changes from using the `windows` crate to the `windows-sys` crate, embedding the various COM interfaces needed by this crate and then using `windows-sys` for the other functions/types used in the crate. This drastically reduces compile time since `windows` is both extremely generic as well as each feature including a ton of code of which only a small fraction is actually used by this crate. (eg. the `Win32_Media_Audio` feature is ~14k LOC). Note that #616 mentions https://github.com/HEnquist/wasapi-rs, however that crate is outdated and still has a dependency on `windows`, but I can split out the COM code in here to a separate crate if that is wanted.

* Crate download: ~12MiB -> ~2.5MiB
* Untarred source: ~171MiB -> ~30MiB
* Build (debug): ~9.8s (1.2s codegen) -> ~1.5s
* Build (release): ~11.04s (2.8s codegen) -> ~1.17s

See: #727
Closes: #616


src/host/wasapi/com.rs
```diff
@@ -56,3 +56,260 @@ impl Drop for ComInitialized {
...
+    }
+}
+
+impl<T> Clone for Object<T> {
+    fn clone(&self) -> Self {
+        unsafe { ((*self.0.cast::<Interface<IUnknownV>>()).vtbl().add_ref)(self.ptr()) };
+        Self(self.0)
+    }
+}
+
```
Jake-Shadle: This is just minimal basic types for wrapping the COM pointers.

src/host/wasapi/com.rs
```diff
@@ -56,3 +56,260 @@ impl Drop for ComInitialized {
...
+    __get_previous: usize,
+    capture_propagation_context:
+        unsafe extern "system" fn(this: *mut c_void, languageexception: *mut c_void) -> HRESULT,
+    __get_propagation_context_head: usize,
+}
+
+#[repr(transparent)]
+struct ILanguageExceptionErrorInfo2(Object<ILanguageExceptionErrorInfo2V>);
+
+pub(crate) fn get_error_message(e: HRESULT) -> String {
```
Jake-Shadle: This block of code + associated types above are just replicating the message retrieval that `windows` uses in its result type. Ideally this would be in its own crate but I didn't think it made sense to use both windows and windows-sys since there are already version headaches for them.

src/host/wasapi/com/audio.rs
```diff
@@ -0,0 +1,680 @@
+#![allow(non_snake_case)]
```
Jake-Shadle, line 1: This file just has all of the audio COM objects/interfaces used by cpal, and just keeps the same interface that windows was using. All methods not used by cpal are just replaced with `usize` to keep the layout correct but not take the compile hit of typechecking a function pointer that is never actually used.

src/host/wasapi/com/threading.rs
```diff
@@ -0,0 +1,26 @@
...
+pub unsafe fn CreateEventA(
+    attrs: Option<*const std::ffi::c_void>,
+    manual_reset: bool,
+    initial_state: bool,
+    name: Option<*const u8>,
+) -> Result<isize, std::io::Error> {
+    // We manually create this extern so that we don't have to include the "Security" feature,
+    // which adds 70 structs, 394 constants, 133 functions, 31 type aliases, and
+    // 5 unions....none of which this crate uses
+    windows_targets::link!("kernel32.dll" "system" fn CreateEventA(lpeventattributes: *const std::ffi::c_void, manualreset: i32, initialstate: i32, name: windows_sys::core::PCSTR) -> isize);
```
Jake-Shadle: As noted in the comment, we just manually bind this as otherwise the Security feature needs to be enabled which is entirely pointless for this crate.

src/host/wasapi/device.rs
```diff
@@ -11,22 +11,21 @@ use std::fmt;
...
 use std::os::windows::ffi::OsStringExt;
-use std::ptr;
 use std::slice;
 use std::sync::{Arc, Mutex, MutexGuard};
 use std::time::Duration;
 
 use super::com;
+use super::com::audio;
+use super::com::threading;
+use super::com::ComResult;
```
Jake-Shadle: We add these imports with different casing, so that the case just needs to be changed on the bindings embedded in this crate, but then the data types (eg WAVEFORMATEX) just come from windows-sys instead of windows.

src/host/wasapi/device.rs
```diff
@@ -422,13 +420,8 @@ impl Device {
...
         let lock = match self.ensure_future_audio_client() {
             Ok(lock) => lock,
-            Err(ref e) if e.code() == Audio::AUDCLNT_E_DEVICE_INVALIDATED => {
-                return Err(SupportedStreamConfigsError::DeviceNotAvailable)
-            }
             Err(e) => {
-                let description = format!("{}", e);
-                let err = BackendSpecificError { description };
-                return Err(err.into());
+                return Err(windows_err_to_cpal_err(e));
```
Jake-Shadle, line 421: This pattern came up several times, so was an opportunity to get rid of redundant code.

src/host/wasapi/stream.rs
```diff
@@ -234,23 +230,29 @@ fn process_commands(run_context: &mut RunContext) -> Result<bool, StreamError> {
 // next event might be some command submitted by the user (the first handle) or
 // might indicate that one of the streams is ready to deliver or receive audio.
 fn wait_for_handle_signal(handles: &[Foundation::HANDLE]) -> Result<usize, BackendSpecificError> {
-    debug_assert!(handles.len() <= SystemServices::MAXIMUM_WAIT_OBJECTS as usize);
+    debug_assert!(handles.len() <= 64 /*SystemServices::MAXIMUM_WAIT_OBJECTS as usize*/);
     let result = unsafe {
```
Jake-Shadle: The `SystemServices` feature includes 14.6k LOC in windows and 10k LOC in windows-sys which is...kind of ridiculous when we only need 1 constant.

Ralith:
Awesome improvements! Would it make sense to take this a step further and use the windows-bindgen crate, to avoid even downloading the unnecessary -sys bindings?

Jake-Shadle:
I'm totally up for that, I started off with windows-sys because previous PRs I've made that get rid of dependencies altogether have all gotten push back which has been annoying, but it's up to the maintainers IMO so if you're up for it then that's even better.

Ralith:
I can't speak for maintainership, but it's at least an interesting possibility.

est31:
> but I can split out the COM code in here to a separate crate if that is wanted.

What about filing a PR to the `windows` crate directly to further split features or suggesting it to them? In [this discussion](https://github.com/microsoft/windows-rs/issues/1720#issuecomment-1484672558) the windows maintainers seem to be kinda open to suggestions.

Generally I'm not the greatest fan of the `windows` not having stopped with the rapid release schedule. This is not good for cpal in specific and the crates ecosystem in general.

est31:
I haven't made up my mind about this PR yet, and haven't given it a very close look yet, but thanks for filing it because it's an important thing to discuss -- I've [suggested switching to the -sys crate earlier](https://github.com/RustAudio/cpal/pull/726), and it's really amazing to see such build time improvements. But on the other hand, the diff is larger than I'd like. Accepting any PR means for me as a maintainer that I either have to maintain that merged code in some fashion, or have to do work to revert to the old state.

So I don't know how that code should be maintained. Where is it  from? Was it auto-generated? What if windows_sys changes their API, is there a way to auto-generate it again? If so, maybe that should be made as easy as possible, i.e. not putting it into files with manually written code, isolating it into one directory, and documenting how to do it. Putting it into a separate crate isn't that helpful, it means an overhead when publishing things if the crate is ours, and also additional complexity for doing changes whether it is ours or not.

Jake-Shadle:
> What about filing a PR to the windows crate directly to further split features or suggesting it to them?

I don't know if something like that would ever happen. The feature granularity is essentially just inherited from the windows metadata namespaces, so either those namespaces would need to be more granular, or the windows crate would need to add additional sub features on top of them specifically when creating Rust bindings, which seems tedious for both the maintainers and end users. IMO the "bind what your crate needs" approach used in this PR makes the most sense.

> Generally I'm not the greatest fan of the windows not having stopped with the rapid release schedule. This is not good for cpal in specific and the crates ecosystem in general. (Edit: switching to windows-sys doesn't directly put us into a better situation here, just that the cost that two versions of windows-sys in a crate graph incur is lower).

I'm very much aware and agree since I was the one who opened https://github.com/microsoft/windows-rs/issues/1720, I would have removed windows-sys completely but I've gotten pushback in [several](https://github.com/steffengy/schannel-rs/pull/90) [previous](https://github.com/notify-rs/notify/pull/473) [PRs](https://github.com/console-rs/console/pull/161) about this, so keeping windows-sys was my attempt at increasing the chances this PR would be merged, removing it is entirely is easily doable though.

> I haven't made up my mind about this PR yet, and haven't given it a very close look yet, but thanks for filing it because it's an important thing to discuss -- I've https://github.com/RustAudio/cpal/pull/726, and it's really amazing to see such build time improvements. But on the other hand, the diff is larger than I'd like. Accepting any PR means for me as a maintainer that I either have to maintain that merged code in some fashion, or have to do work to revert to the old state.

Understandable, I'm opening this in the hopes of merging, but even if it isn't merged in this current state at least starting a discussion of changes how windows is interacted with, in whatever form that takes. This crate is one of only 3 in our dependency graph (of many hundreds) that depend on windows so we're motivated to find a solution.

(the other two)

- https://github.com/PolyMeilex/rfd/pull/118
- https://github.com/Boddlnagg/midir/pull/131

 > So I don't know how that code should be maintained. Where is it from? Was it auto-generated? What if windows_sys changes their API, is there a way to auto-generate it again? If so, maybe that should be made as easy as possible, i.e. not putting it into files with manually written code, isolating it into one directory, and documenting how to do it. Putting it into a separate crate isn't that helpful, it means an overhead when publishing things if the crate is ours, and also additional complexity for doing changes whether it is ours or not.

The current code is a mix of copy paste of the vtables from the windows crate and just manual implementation of the methods to preserve the existing API. I already have a tool to do bindgen of the windows-sys level of the API, but it doesn't have COM support, though that would be trivial to add, that way one could generate the vtables (excluding the methods that the crate doesn't want) as well as all of the data types needed (without windows_sys), so if that's wanted I can add that and open source the tool so the code can be regenerated/changed over time.

riverar:
@Jake-Shadle Is there something about the version churn claims that we're missing in https://github.com/microsoft/windows-rs/issues/1720? We haven't seen any data to support your ecosystem impact claims. But do please throw something in there if we're missing something.

Any particular reason `windows-bindgen` isn't used here despite [indicating you would](https://github.com/microsoft/windows-rs/issues/1720#issuecomment-1486795961)?

Would be great to also get more context/data around why you are aggressively stripping windows crate dependencies at the expense of [increasing crate maintanence](https://github.com/RustAudio/cpal/pull/778#issuecomment-1516921403), [churn](https://github.com/amodm/webbrowser-rs/pull/62), and [failure risk](https://github.com/Amanieu/parking_lot/pull/378). Are these perhaps dependencies your projects are using and you're trying to trim the fat?

Jake-Shadle:
Version churn is still an issue, but removing `windows` is more about reducing compile times/download size/diskspace.

> Would be great to also get more context/data around why you are aggressively stripping windows crate dependencies at the expense of https://github.com/RustAudio/cpal/pull/778#issuecomment-1516921403, https://github.com/amodm/webbrowser-rs/pull/62, and https://github.com/Amanieu/parking_lot/pull/378. Are these perhaps dependencies your projects are using and you're trying to trim the fat?

Yes, I don't go around submitting PRs to crates we don't use.

* Increasing crate maintenance - This is IMO not a big deal, yes it is more code in the crate, but the maintenance of that code is not even close to the same level as normal code because the API doesn't change. The only time the code would really meaningfully need to change would be adding support for some new Windows audio feature or interface, but that would presumably require regular work in the code calling into windows regardless.
* Churn - The PR you linked had 1 small mistake that was easily fixed, the whole point of putting the bindings inside the crate is to _reduce_ the churn of updating `windows/-sys` versions (though to be fair `windows` cadence is less bad than `windows-sys`)
* Failure risk - I'm fine with using `windows-targets`, it doesn't impact compile times, but I also think the "failure risk" is exaggerated

Removing the dependency on `windows` for this crate and the others means we get rid of a 30+s of build time
image

The reason that time is so much longer than the times I stated above is because in a large workspace with hundreds of crates cargo will by default saturate the CPU, and thus crates with longer compile times tend to become even longer since they are competing with each other for CPU time as opposed to the isolated case of eg. just building `cpal` which can't saturate the CPU due to how few crates are built.

The situation is of course much better for `windows-sys` as unlike `windows` it doesn't actually have any implementation to compile/massive amount generics, but in a large enough project you are _guaranteed_ to have multiple versions of windows-sys due to how often it bumps minor versions (we currently have 3). But still not great...

image

The compile time here is not the worst, but it has the negative impact that it blocks 23 crates from beginning compilation, and the compilation time can be _far_ better since every feature enabled on `windows-sys` will _by default_ include dozens to _thousands_ of lines of type/struct/constant/function bindings of which a vast majority _will not be used_ but are still paid for.

Jake-Shadle:
@est31 This PR now replaces `windows` with embedded bindings for everything instead of mixing manual COM interfaces and `windows-sys` bindings for types/functions.

The code is generated with a tool that I haven't published yet ([source code here](https://github.com/EmbarkStudios/minwin/tree/impl/src)) as cpal was one of the crates I was using as a test case, but is otherwise just using `[windows-metadata`](https://docs.rs/windows-metadata/0.49.0/windows_metadata/) the same as the windows/windows-sys crates to get and read the metadata used to generate the bindings.

While the embedded bindings may _seem_ large at ~1120 lines, they are a minute fraction of code that would otherwise be compiled using either `windows` or `windows-sys`.

```
Summary for 11 namespaces
  records: 52 / 2765 => 1.88%
  functions: 9 / 2375 => 0.38%
  constants: 25 / 23085 => 0.11%
  aliases: 12 / 856 => 1.40%
  func pointers: 0 / 604 => 0.00%
  interfaces: 10 / 10 => 100.00%
```

PS: The windows CI failures happen on master, I'll rebase once/if they are fixed.

MarijnS95:
> The code is generated with a tool that I haven't published yet ([source code here](https://github.com/EmbarkStudios/minwin/tree/impl/src))

@Jake-Shadle was this repo removed now that Microsoft feature-completed and published their official tooling that achieves the same?

And are you perhaps able to respin this PR using the official tooling? Here's what that looks like: https://github.com/Traverse-Research/amd-ext-d3d-rs/commit/fa55fdb98772c07f4ba0825a4a5f770131a09c05. Most important is `bindings.txt` and `api_gen/src/main.rs`.

Jake-Shadle:
Probably not, the official tooling configurability is too limited and the output has a lot of undesirable pieces that users can't control. I'm more likely going to end up going a different route and just generating a patch crate for windows/-sys since the patch process means I don't have to make PRs that people don't like.

MarijnS95:
Fair enough, the tooling had a bunch of push-back on useful features so it seems unlikely that any requested or contributed improvement makes it in.

In particular the generator emits _all_ functions for a type, even those that depend on external modules which you will now also have to generate. There's the `cfg` option to emit `cfg(feature)` bounds but that is **exactly what you don't want** for generating embedded bindings: those internal guards shouldn't leak out to the public features interface in `Cargo.toml`.
Is that what you meant by "the output has a lot of undesirable pieces that users can't control"? Perhaps I should make an upstream issue about this.

Looking forward to seeing your patches/fork of those crates, keep me in the loop!

PR #779, opened 2023-04-18
Fixed warnings in asio_sys

Dicklessgreat:
From #775 [(comment)](https://github.com/RustAudio/cpal/pull/775#issuecomment-1500953921)

Fixed `asio-sys/src/bindings/errors.rs`
`std::Error::description()` has been deprecated.
It should be replaced with `std::fmt::Display`
I found some code duplication in the `std::fmt::Display` and `std::Error` trait, so I have combined the code in `std::fmt::Display`.
This will be a breaking change, is that OK?
The `std::Error` implementation of `LoadDriverError` and `AsioError` is empty, but I think this is necessary for error handling, so I left it.

Fixed `src/host/asio/stream.rs`
I wondered what to do with the `asio_stream` field, but decided to ignore the warning as it might be there to extend the lifetime.I am not 100% sure that this is the right way to do it.

### DIFF
--- a/asio-sys/src/bindings/errors.rs
+++ b/asio-sys/src/bindings/errors.rs
@@ -39,7 +39,21 @@ pub enum AsioErrorWrapper {
 
 impl fmt::Display for LoadDriverError {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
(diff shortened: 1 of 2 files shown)


est31 (approved):
Thanks!

PR #784, opened 2023-05-29
Improve synth tones example

jamierpond:
Now also plays through a series of waveforms :)

Also here making some aesthetic judgements on naming/convention which am very happy to have a conversation about!

### DIFF
--- a/examples/synth_tones.rs
+++ b/examples/synth_tones.rs
@@ -13,54 +13,100 @@ use cpal::{
 use cpal::{FromSample, Sample};
 
 fn main() -> anyhow::Result<()> {
-    let stream = stream_setup_for(sample_next)?;
+    let stream = stream_setup_for()?;
     stream.play()?;
-    std::thread::sleep(std::time::Duration::from_millis(3000));
+    std::thread::sleep(std::time::Duration::from_millis(4000));
     Ok(())
 }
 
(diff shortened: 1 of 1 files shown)


est31 (approved):
Looks good, although could you put a trailing 0 instead of trailing dot for the floats?

jamierpond:
@est31 do you know why that windows test might be failing? I imagine that's something that's regressed outside the scope of this PR?

examples/synth_tones.rs
```diff
@@ -82,29 +128,44 @@ pub fn host_device_setup(
...
+
     let stream = device.build_output_stream(
         config,
         move |output: &mut [T], _: &cpal::OutputCallbackInfo| {
-            on_window(output, &mut request, on_sample)
+            // for 0-1s play sine, 1-2s play square, 2-3s play saw, 3-4s play triangle_wave
+            let time_since_start = std::time::Instant::now()
+                .duration_since(time_at_start)
+                .as_secs_f32();
+            if time_since_start < 1. {
```
est31: here too

jamierpond: oh thanks!

est31 (approved):

hugovdm:
I have questions about this PR... and at the very least, a fix that makes it play 440Hz instead of the 880Hz it is currently playing:
https://github.com/RustAudio/cpal/pull/792

PR #792, opened 2023-07-30
Fix synth_tones frequency

hugovdm:
Each *_wave() function calls advance_sample(). Having tick() also call it results in skipping every second sample.

Given the name "tick()", one could argue whether advance_sample() rather belongs in this function than in *_wave().

More interesting to me though: I wondered about generative_waveform() trying to build what are fairly simple waveforms out of a finite number of sine waves. It's an interesting demonstration, but it's also (a) an inaccurate approximation (e.g. why stop at the nyquist frequency, vs continuing to 2x or 3x the nyquist frequency? - that does in fact give cleaner results in the time domain), and (b) error prone, as per the triangle wave being wrong.

Here's a graph showing each of the three waveforms currently produced by synth_tones. I've plotted all of them in one plot, 2.5 cycles of a waveform at a time - and due to the hacky way I grabbed samples, I've plotted the sine wave twice. Note that what was supposed to be a triangular wave, somehow turned out to be closer to "circular"? Maybe this can be fixed by changing the parameters to generative_waveform(), but... is this indirect demonstration of the DFT valuable enough to keep, or should we not make this example simpler and just generate the waves directly and accurately in the time domain?
NyquistOriginal
NyquistOriginalTriangle
NyquistOriginalSquareWave
NyquistOriginalSaw

And for curiosity's sake, a plot with generative_waveform() continuing to 4x the nyquist frequency:
NyquistTimesFour

I was indeed curious under which conditions the sampled nature of the signal could avoid the ringing caused by using a finite number of waves. It's not going to help that the usual sampling rates (both 44100 Hz and 48000 Hz) aren't multiples of 440 Hz: so each cycle, there's a little phase shift in where the theoretical/analytical waveform is sampled.

### DIFF
--- a/examples/synth_tones.rs
+++ b/examples/synth_tones.rs
@@ -35,7 +35,7 @@ pub struct Oscillator {
 
 impl Oscillator {
     fn advance_sample(&mut self) {
-        self.current_sample_index = (self.current_sample_index + 1.0) % self.sample_rate;
+        self.current_sample_index = self.current_sample_index + 1.0;
     }
 
     fn set_waveform(&mut self, waveform: Waveform) {
--- a/examples/synth_tones.rs
+++ b/examples/synth_tones.rs
@@ -35,7 +35,7 @@ pub struct Oscillator {
 
 impl Oscillator {
     fn advance_sample(&mut self) {
-        self.current_sample_index = self.current_sample_index + 1.0;
+        self.current_sample_index = (self.current_sample_index + 1.0) % self.sample_rate;
     }
 
     fn set_waveform(&mut self, waveform: Waveform) {
--- a/examples/synth_tones.rs
+++ b/examples/synth_tones.rs
@@ -81,7 +81,6 @@ impl Oscillator {
     }
 
     fn tick(&mut self) -> f32 {
-        self.advance_sample();
         match self.waveform {
             Waveform::Sine => self.sine_wave(),
             Waveform::Square => self.square_wave(),
--- a/examples/synth_tones.rs
+++ b/examples/synth_tones.rs
@@ -76,6 +76,8 @@ impl Oscillator {
         self.generative_waveform(1, 1.0)
     }
 
+    // DO NOT SUBMIT: PR discussion to be had on generative_waveform(), and
+    // triangle_wave() is not producing anything triangular.
     fn triangle_wave(&mut self) -> f32 {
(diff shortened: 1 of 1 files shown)


examples/synth_tones.rs
```diff
@@ -35,7 +35,7 @@ pub struct Oscillator {
 
 impl Oscillator {
     fn advance_sample(&mut self) {
-        self.current_sample_index = (self.current_sample_index + 1.0) % self.sample_rate;
+        self.current_sample_index = self.current_sample_index + 1.0;
```
hugovdm: Uhm - I might be wrong here. It *is* supposed to cycle back every second. :thinking: - I wonder why I saw a discontinuity in the time-domain then, haven't figured this out yet.

hugovdm: Precision in the significand of f32: 24 bits. Without the `% sample_rate`, at 44.1kHz, this increment breaks at around 6m20s. An i32 instead of an f32 would handle 13.5 hours, and a u32 lasts a bit more than a day.

(So in my own code, I'm currently counting samples with an Int64. If supporting e.g. 16 channels at 192 kHz, 32 bits only lets you count samples for 23.3 minutes.)

hugovdm:
What's the PR review culture in RustAudio/cpal? I would instinctively add the author and reviewer of #784 to this PR... Instead, for now, I'll wait and watch to see how reviewers get assigned. :-)

est31:
Good point about the removal of the `advance_sample` call. It makes sense to do that. Can you remove the `DO NOT SUBMIT` line?

est31:
> And for curiosity's sake, a plot with generative_waveform() continuing to 4x the nyquist frequency

that looks really beautiful. Maybe we should do that instead, together with a comment? Ideally that would be a followup PR though, I'd like to merge the first fix for 440Hz more quickly.

hugovdm:
I've removed the "do not submit", so that this PR is just a one-liner. I'll create a new PR for debating `generative_waveform()` - I don't like it, even at 4*nyquist_frequency: notice the nasty-looking ringing on both the square wave and the saw wave. If we want to demonstrate how the DFT works in one of the examples, then how about we make an example that more explicitly does that? E.g. recreating the "Additive Triangle wave sound sample" from https://en.wikipedia.org/wiki/Triangle_wave

(Pity there's no obvious choice for adding visual representations to examples? :slightly_smiling_face: )

Issue #277, opened 2019-06-06
Minimum Rust version policy


ishitatsuyuki:
Throughout the codebase there are extensive use of old syntax that were superseded by new features like match ergonomics, inclusive ranges, and struct field names sugar.

What's our policy on minimum Rust compiler version? I guess we can just opt for bumping it in next major (0.x) release.

mitchmindtree:
I'm not aware of any existing policy, but it would be nice to establish one!

Perhaps first it would be worth identifying the current minimum required version (not sure how to best go about this)? Then we could take a look at the plans for upcoming work (e.g. #268, #204) and identify any more modern-rust features we might need? I guess it's reasonable enough to raise the minimum to some more recent version if there's a motivating factor in the code-base and it happens between major version bumps?

Lokathor:
Hi, I've been going around trying to get folks to have such a policy and so I'm happy to find this issue.

It seems that currently [appveyor](https://github.com/tomaka/cpal/blob/master/appveyor.yml) is still using `stable` and so is [travis](https://github.com/tomaka/cpal/blob/master/.travis.yml). The first step to finding a current minimum version would be to just edit the word `stable` to be a particular full version number, such as `1.36.0` and let CI run again. You can "binary search" the version up and down until CI breaks on one version and works on the next, then you know.

est31:
Through trial and error I've determined that current master needs at least Rust 1.32.0. Rust 1.31.0 doesn't work because there, macro_literal_matcher isn't stable yet. This is incurred by code in the crate itself as well as dependencies.

Note that taking this version and fixing it forever is not doable without additional steps like commiting Cargo.lock into git like the failure crate does, because one of the recursive deps of cpal has a 2-releases-back MSRV policy: https://github.com/rust-lang/backtrace-rs/issues/211#issuecomment-504970353

We can also think about turning off optional features of the failure crate. Then we'd have less deps and would avoid the backtrace crate.

joseluis:
Acording to `cargo-msrv` the MSRV as of today is `1.60.0`.

roderickvd:
MSRV is 1.70 but the question is if we have a policy. We don't but I think it'd be good to have one. Given `cpal`s rather low-level and ubiquitous nature, I think it should be fairly conservative. Open to proposals.

Lokathor:
[...]

PR #800, opened 2023-09-20
Remove once_cell dependency

james7132:
Removes the `once_cell` dependency, using `std::sync::OnceLock` instead.

This bumps the MSRV of the crate from 1.60 to 1.70. Given that #277 is currently unresolved, is this acceptable?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,6 +7,7 @@ documentation = "https://docs.rs/cpal"
 license = "Apache-2.0"
 keywords = ["audio", "sound"]
 edition = "2021"
+rust-version = "1.70"
 
 [features]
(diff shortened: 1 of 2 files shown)


est31:
I think this is a little bit *too* soon. 1.70 was released June 1st, we should wait at least until mid/end of december.

adsick:
mid of december has come)

est31 (approved):

est31:
there seems to be CI failure. @james7132 could you have a look?

james7132:
The new CI failures seem to be unrelated.

est31:
Regarding the CI failures, see #810

Issue #801, opened 2023-10-12
[ASIO] Pre build script


alisomay:
Hey 👋 

I followed the readme for the asio setup which involves multiple steps. 
It worked fine for me.
Then I've also written a PowerShell script to automatize it a bit.
I'd like to share it here maybe it will be useful to someone.  
This is not a suggestion for adoption but it makes things easier in development.

Also it is a bit annoying that we can do pre-build scripts in rust build system, it would have been useful for such cases.

All the best.

```ps1
Write-Output "============================================================================="
Write-Output "Pre build script for cpal asio feature."
Write-Output "Make sure that you have sourced this script instead of directly executing it."
Write-Output "============================================================================="

function Invoke-VcVars {
    <#
    .SYNOPSIS
    This function sets the Visual Studio build environment for the current session.

    .DESCRIPTION
    The function first determines the system architecture. It then searches for the vcvarsall.bat file 
    specific to the detected architecture and executes it to set the Visual Studio build environment 
    for the current session.
    #>
    
    # Determine the system architecture
    Write-Output "Determining system architecture..."
    $arch = if ([Environment]::Is64BitOperatingSystem) {
        switch -Wildcard ((Get-CimInstance -ClassName Win32_Processor).Description) {
            "*ARM64*"{ "arm64" }
            "*ARM*"{ "arm" }
            default { "amd64" }
        }
    } else { 
        "x86" 
    }

    Write-Output "Architecture detected as $arch."

    # Define search paths based on architecture
    $paths = if ($arch -eq 'amd64') {
        @('C:\Program Files (x86)\Microsoft Visual Studio\', 'C:\Program Files\Microsoft Visual Studio\')
    } else {
        @('C:\Program Files\Microsoft Visual Studio\')
    }

    # Search for vcvarsall.bat and execute the first instance found with the appropriate architecture argument
    Write-Output "Searching for vcvarsall.bat..."
    foreach ($path in $paths) {
        $vcvarsPath = Get-ChildItem $path -Recurse -Filter vcvarsall.bat -ErrorAction SilentlyContinue | Select-Object -First 1
        if ($vcvarsPath) {
            Write-Output "Found vcvarsall.bat at $($vcvarsPath.FullName). Initializing environment..."
            $cmdOutput = cmd /c """$($vcvarsPath.FullName)"" $arch && set"
            foreach ($line in $cmdOutput) {
                if ($line -match "^(.*?)=(.*)$") {
                    $varName = $matches[1]
[...]

PR #804, opened 2023-10-16
Automate asio sys build

alisomay:
Hello all,

Having a rather manual pre build preparation process for the windows ASIO support could be cumbersome for users who'd like to install applications we publish to crates.io which depends on `cpal` by `cargo install` method because `cargo install` builds from source. 

This PR is a suggestion to automate the two steps which is laid out in the [readme.](https://github.com/RustAudio/cpal#asio-on-windows).

- Automate the finding and retrieval of ASIO SDK. 
- Set the build environment by invoking `vcvarsall.bat`

Trying to install LLVM would be out of scope for this script in my opinion because that is more general and essential to have a development environment. If the build fails due to that, normally the user can easily derive the reason from the error message. 

I didn't update the `README.md` yet but if you find this PR welcome, I'd gladly do so.

Looking forward to your review!

*I've tested all the code paths on a x86_64 machine running Windows 10. 
*The PR is an adaptation of a ps1 script I use for my project to automate the process in https://github.com/RustAudio/cpal/issues/801

### DIFF
--- a/asio-sys/Cargo.toml
+++ b/asio-sys/Cargo.toml
@@ -10,7 +10,7 @@ keywords = ["audio", "sound", "asio", "steinberg"]
 build = "build.rs"
 
 [target.'cfg(any(target_os = "windows"))'.build-dependencies]
-bindgen = "0.64"
+bindgen = "0.68"
 walkdir = "2"
 cc = "1.0.25"
 
--- a/asio-sys/Cargo.toml
+++ b/asio-sys/Cargo.toml
@@ -12,7 +12,9 @@ build = "build.rs"
 [target.'cfg(any(target_os = "windows"))'.build-dependencies]
 bindgen = "0.68"
 walkdir = "2"
-cc = "1.0.25"
+cc = "1.0.83"
+reqwest = { version = "0.11" , features = ["blocking"] }
+zip = "0.6"
 
 [dependencies]
 once_cell = "1.12"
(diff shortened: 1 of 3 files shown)


alisomay:
I can also develop it further if more robustness is required.

asio-sys/Cargo.toml
```diff
@@ -10,9 +10,11 @@ keywords = ["audio", "sound", "asio", "steinberg"]
 build = "build.rs"
 
 [target.'cfg(any(target_os = "windows"))'.build-dependencies]
-bindgen = "0.64"
+bindgen = "0.68"
 walkdir = "2"
-cc = "1.0.25"
+cc = "1.0.83"
+reqwest = { version = "0.11" , features = ["blocking"] }
```
est31: we should not include reqwest for a tiny task like a build script. it's better to invoke some available windows-specific command that downloads a file from an url. same goes for the zip crate, if possible

est31 (commented):
in general it's a good idea to have a fallback mode where we download the asio SDK, but it should not cause additional dependencies for users who already have the SDK installed.

and it should also be mentioned in README :).

alisomay:
> in general it's a good idea to have a fallback mode where we download the asio SDK, but it should not cause additional dependencies for users who already have the SDK installed.
> 
> and it should also be mentioned in README :).

Thanks for the quick look 🎉 

I agree, the PR is meant to be a vehicle to show my intention.
I'll re-work it in a way which it avoids dependencies and uses windows apis. 

Also update the README.

Please don't hesitate to add any other ideas.

alisomay:
- `README.md` updated.
- Build script uses powershell commands now.
- Build script detects if `vcvarsall.bat` is invoked in the session or not.

est31 (approved):

PR #808, opened 2023-10-22
Fix runtime errors on Android targets

PhilosophicalProgrammer:
When developing an Android app using `cpal` for sound effects I had to configure my `build.rs` to do this to avoid crashes at startup on certain phones. I figured I should help people avoid having to do the same debugging / work (it took many hours to figure this out) by adding the necessary code to `cpal`'s `build.rs`.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -1,9 +1,13 @@
 use std::env;
-
 const CPAL_ASIO_DIR: &str = "CPAL_ASIO_DIR";
 
 fn main() {
     println!("cargo:rerun-if-env-changed={}", CPAL_ASIO_DIR);
+
+    // This avoids runtime errors on Android targets.
+    if std::env::var("TARGET").unwrap().contains("android") {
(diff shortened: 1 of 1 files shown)


est31:
Is there any further reading on these errors? Error messages, stack overflow threads, etc.

PhilosophicalProgrammer:
@est31 Yeah, I think it was something about a dynamic library it couldn't find (?). Removing `cpal` from the dependencies fixed the error, so it was definitely related to this. So I googled a bit and found this fix. To be honest I don't completely remember the details, it's been a while. If necessary I could remove the fix from my project and re-create the error to put it here.

est31:
@PhilosophicalProgrammer

> If necessary I could remove the fix from my project and re-create the error to put it here.

If possible, that would be very helpful, because then we know why we added this piece of code (and can tell if the code will need it in the future). Also it might be helpful to be more concrete with the phone models. Are they recent? are they older? Is the fix still needed in 5 years?

PR #809, opened 2023-11-05
Improve asio sys build script

alisomay:
# Improvements

- `vcvarsall.bat` is executed exhaustively for any supported host and target architecture combinations now. 
-  Downloading the ASIO SDK is working on all platforms now which enables cross compilation. 
-  A section on the `README.md` is added which explains cross compilation scenarios for `cpal` with the `asio` feature.
-  `vcvarsall.bat` search is improved greatly to execute smarter and faster to cover for CI scenarios where the disk I/O might be a bottle neck. 
-  A type mismatch bug which was apparent on using the `mingw-w64` tool chain is fixed. 

With these changes I want to enable:
- Better building experience in CI
- Improve existing changes to be cross compilation ready.

What I have tried:

- macOS aarch64 (host) -> windows x86, x86_64 (target) ✅ 
- linux (ubuntu) aarch64 (host) -> windows x86, x86_64 (target) ✅ 
- windows aarch64 (host)-> windows x86, x86_64, aarch64 (target) ✅ 
- windows x86_64 (host)-> windows x86, x86_64, aarch64 (target) ✅ 

If I missed something, I'll correct it along the way.
Thank you for the last merge, and I look forward to your review 🙏

### DIFF
--- a/asio-sys/Cargo.toml
+++ b/asio-sys/Cargo.toml
@@ -9,12 +9,18 @@ license = "Apache-2.0"
 keywords = ["audio", "sound", "asio", "steinberg"]
 build = "build.rs"
 
-[target.'cfg(any(target_os = "windows"))'.build-dependencies]
-bindgen = "0.68"
+[build-dependencies]
+bindgen = "0.69"
 walkdir = "2"
 cc = "1.0.83"
+parse_cfg = "4.1.1"
 
 [dependencies]
 once_cell = "1.12"
-num-derive = "0.3"
+num-derive = "0.4"
 num-traits = "0.2"
+
+
+
+
+
--- a/asio-sys/Cargo.toml
+++ b/asio-sys/Cargo.toml
@@ -19,8 +19,3 @@ parse_cfg = "4.1.1"
 once_cell = "1.12"
 num-derive = "0.4"
 num-traits = "0.2"
-
-
-
-
-
--- a/asio-sys/build.rs
+++ b/asio-sys/build.rs
@@ -421,7 +421,7 @@ fn search_vcvars_all_bat() -> PathBuf {
     }
 
     // Define search paths for vcvarsall.bat based on architecture
-    let paths = vec![
+    let paths = &[
         // Visual Studio 2022+
         "C:\\Program Files\\Microsoft Visual Studio\\",
         // <= Visual Studio 2019
@@ -533,8 +533,9 @@ fn guess_vcvars_all_bat() -> Option<PathBuf> {
 /// Because of this there is no native or cross compilation is supported for 32 bit arm processors.
 fn determine_vcvarsall_bat_arch_arg() -> String {
     let host_architecture = std::env::consts::ARCH;
+    let target_architecture = std::env::var("CARGO_CFG_TARGET_ARCH").expect("Target not set.");
 
-    let arch_arg = if cfg!(target_arch = "x86_64") {
+    let arch_arg = if target_architecture == "x86_64" {
         if host_architecture == "x86" {
             // Arg for cross compilation from x86 to x64
             "x86_amd64"
@@ -547,7 +548,7 @@ fn determine_vcvarsall_bat_arch_arg() -> String {
         } else {
             panic!("Unsupported host architecture {}", host_architecture);
         }
-    } else if cfg!(target_arch = "x86") {
+    } else if target_architecture == "x86" {
         if host_architecture == "x86" {
             // Arg for native compilation from x86 to x86
             "x86"
@@ -560,7 +561,7 @@ fn determine_vcvarsall_bat_arch_arg() -> String {
         } else {
             panic!("Unsupported host architecture {}", host_architecture);
         }
-    } else if cfg!(target_arch = "arm") {
+    } else if target_architecture == "arm" {
         if host_architecture == "x86" {
             // Arg for cross compilation from x86 to arm
             "x86_arm"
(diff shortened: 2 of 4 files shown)


alisomay:
I've also checked a bit about the reason why `asmjs-wasm32-test` fails the CI.
This snippet
```rust
#[cfg(target_os = "emscripten")]
impl wasm_bindgen::convert::IntoWasmAbi for BufferSize {
    type Abi = wasm_bindgen::convert::WasmOption<u32>;
    fn into_abi(self) -> Self::Abi {
        match self {
            Self::Default => None,
            Self::Fixed(fc) => Some(fc),
        }
        .into_abi()
    }
}
```
Has this type `wasm_bindgen::convert::WasmOption<u32>`.

But that type does not exist anymore in the [module](https://docs.rs/wasm-bindgen/latest/wasm_bindgen/convert/index.html).

The changes made to that snippet was 12 months ago and the last time I have checked the CI was green a few weeks ago. 

I wonder if my changes made a side effect..

alisomay:
No significant rush but when approximately you may look at this?
Thanks 🙏

asio-sys/Cargo.toml
```diff
@@ -9,12 +9,18 @@ license = "Apache-2.0"
...
 cc = "1.0.83"
+parse_cfg = "4.1.1"
 
 [dependencies]
 once_cell = "1.12"
-num-derive = "0.3"
+num-derive = "0.4"
 num-traits = "0.2"
+
+
```
est31: why these blank lines?

asio-sys/build.rs
```diff
@@ -279,72 +279,316 @@ fn get_asio_dir() -> PathBuf {
...
     // Define search paths for vcvarsall.bat based on architecture
-    let paths = if arch == "amd64" {
-        vec![
-            "C:\\Program Files (x86)\\Microsoft Visual Studio\\",
-            "C:\\Program Files\\Microsoft Visual Studio\\",
-        ]
-    } else {
-        vec!["C:\\Program Files\\Microsoft Visual Studio\\"]
-    };
+    let paths = vec![
```
est31: this can just be a `&[...]` right?

asio-sys/build.rs
```diff
@@ -279,72 +279,316 @@ fn get_asio_dir() -> PathBuf {
...
-    env::var("VCINSTALLDIR").is_ok()
+
+/// Determines the right argument to pass to `vcvarsall.bat` based on the host and target architectures.
+///
+/// Windows on ARM is not supporting 32 bit arm processors.
+/// Because of this there is no native or cross compilation is supported for 32 bit arm processors.
+fn determine_vcvarsall_bat_arch_arg() -> String {
+    let host_architecture = std::env::consts::ARCH;
+
+    let arch_arg = if cfg!(target_arch = "x86_64") {
```
est31: `target_arch` in a build script also refers to the arch of the host, as the build scripts are always compiled for the host.

est31 (changes requested):

est31:
> I wonder if my changes made a side effect..

I don't think so.

> I don't know how it is used in the wild but this may provide a solution:

feel free to make a PR!

asio-sys/build.rs
alisomay: This is indeed confusing. I didn't catch that thanks!

Running this build script on macos `aarch64`
with the command `cargo build --target=x86_64-pc-windows-gnu` :

```rust
fn main() {
    let target_arch = std::env::var("CARGO_CFG_TARGET_ARCH").unwrap();
    let host_architecture = std::env::consts::ARCH;

    if cfg!(target_arch = "x86_64") {
        println!(r#"cfg!(target_arch = "x86_64")"#);
    }

    if cfg!(target_arch = "aarch64") {
        println!(r#"cfg!(target_arch = "aarch64")"#);
    }

    println!(
        "CARGO_CFG_TARGET_ARCH: {}\nstd::env::consts::ARCH: {}",
        target_arch, host_architecture
    );

    panic!();
}
```
yields

```
  --- stdout
  cfg!(target_arch = "aarch64")
  CARGO_CFG_TARGET_ARCH: x86_64
  std::env::consts::ARCH: aarch64
```

which indeed proves the `target_arch` in the `cfg!` macro is actually the host arch from my point of view.

I'll apply the changes accordingly replacing those with the target arch parsed from the triple. 
I'll also double check if it uses`CARGO_CFG_*` variables. 
https://github.com/rust-lang/cargo/issues/4302

asio-sys/build.rs
alisomay: Probably it can. 👍

asio-sys/Cargo.toml
alisomay: Probably forgotten, I'll just remove them.

asio-sys/build.rs
alisomay: [2d6b063](https://github.com/RustAudio/cpal/pull/809/commits/2d6b063f68431030db0bed654ba597332f418e84)
Fixes it.

asio-sys/build.rs
alisomay: [2d6b063](https://github.com/RustAudio/cpal/pull/809/commits/2d6b063f68431030db0bed654ba597332f418e84)

asio-sys/Cargo.toml
alisomay: [2d6b063](https://github.com/RustAudio/cpal/pull/809/commits/2d6b063f68431030db0bed654ba597332f418e84)

est31 (approved):

PR #825, opened 2024-01-25
Update actions

mockersf:
Even on a successful run, GitHub actions have a lot of warnings:

Screenshot 2024-01-25 at 06 42 38

This PR updates actions:
- switch all `actions/checkout` to the latest v4
- replace `actions-rs/toolchain` by `dtolnay/rust-toolchain`. also use it instead of manual target installs at some places
- replace `actions-rs/cargo` by running the cargo command
- replace `actions-rs/clippy-check` by running clippy
- update `mymindstorm/seup-emsdk` to the latest v14 (and also don't depend on an old version of emscripten as the bug linked as been fixed)
- update `katyo/publish-crates` to the latest v2 (I couldn't test this one)

`actions-rs` actions have not been updated for 3 years and were archived in October 2023

### DIFF
--- a/.github/workflows/cpal.yml
+++ b/.github/workflows/cpal.yml
@@ -111,9 +111,7 @@ jobs:
     steps:
     - uses: actions/checkout@v4
     - name: Setup Emscripten toolchain
-      uses: mymindstorm/setup-emsdk@v10
-      with:
-        version: 2.0.9 # https://github.com/rust-lang/rust/issues/85821
+      uses: mymindstorm/setup-emsdk@v14
     - name: Install stable
       uses: dtolnay/rust-toolchain@stable
       with:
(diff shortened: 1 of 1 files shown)


simlay (commented):
Not like a review from me was requested, but this looks pretty good.

I think my only feedback would be to suggest adding dependabot for actions. [`coreaudio-sys` and `coreaudio-rs` do this](https://github.com/RustAudio/coreaudio-sys/blob/db0f487e5c2953caf3908847c62273c54706c612/.github/dependabot.yml#L4-L8). It obviously wouldn't help with moving from `actions-rs/toolchain` to `dtolnay/rust-toolchain` but it would automate the PRs for `actions/checkout` (at the cost of a little bit of spam).

est31 (approved):
thank you!

PR #832, opened 2024-02-08
ASIO audio mixing fixes

jesnor:
- Fixed ASIO output stream mixing
- Fixed some clippy warnings
- Fixed silencing of buffer when multiple ASIO device streams

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,7 +15,6 @@ oboe-shared-stdcxx = ["oboe/shared-stdcxx"] # Only available on Android. See REA
 
 [dependencies]
 dasp_sample = "0.11"
-byteorder = "1.5"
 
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -15,6 +15,7 @@ oboe-shared-stdcxx = ["oboe/shared-stdcxx"] # Only available on Android. See REA
 
 [dependencies]
 dasp_sample = "0.11"
+byteorder = "1.5"
```
est31: I don't like adding new dependencies, especially not to replace existing code (unless the savings are significant or it is tricky code, neither seems to be the case here).

jesnor: Ok, I removed the dependency and used stdlib methods instead

Issue #818, opened 2024-01-17
Implement `Send` for `Stream`


timstr:
Currently, `Stream` does not implement `Send`, and consequently, I am not allowed to move it from one thread to another in safe Rust. When I want to associate a stream with another struct that is moved between threads, I find myself coming up with the akward workaround of creating an additional thread just to create and store the `Stream` local to the thread:
```rust
// NOTE: Stream is not Send, using a dedicated thread as a workaround
std::thread::spawn(move || {
    let stream = device
        .build_input_stream(&config, data_callback, |err| {
            panic!("CPAL Input stream encountered an error: {}", err);
        })
        .unwrap();
    stream.play().unwrap();
    // synchronization...
    // --- snip ---
    stream.pause().unwrap();
});
```
It's my understanding that CPAL and its audio backends will create dedicated threads already for audio processing, so this overhead seems wasteful and unecessary. It would be much simpler and more efficient if `Stream` was `Send` as well, and the following could simply work:
```rust
let stream = device
    .build_input_stream(&config, data_callback, |err| {
        panic!("CPAL Input stream encountered an error: {}", err);
    })
    .unwrap();
stream.play().unwrap();

std::thread::spawn(move || {
    // synchronization...
    // --- snip ---
    stream.pause().unwrap();
});
```
To be clear, I am not asking that `Stream` be `Sync`, since it seems clear that a `Stream` shouldn't be mutated from multiple threads at the same time without additional synchronization.

Browsing the CPAL source, I find this comment at [src/platform/mod.rs](https://github.com/RustAudio/cpal/blob/310160fbbf507bc27ef751a976550692540b6b9e/src/platform/mod.rs#L64-L72)
```rust
/// The `Stream` implementation associated with the platform's dynamically dispatched
/// [`Host`] type.
// Streams cannot be `Send` or `Sync` if we plan to support Android's AAudio API. This is
// because the stream API is not thread-safe, and the API prohibits calling certain
// functions within the callback.
//
// TODO: Confirm this and add more specific detail and references.
#[must_use = "If the stream is not stored it will not play."]
pub struct Stream(StreamInner, crate::platform::NotSendSyncAcrossAllPlatforms);
```

I'm not personally familair with Android development, nor do I have an Android device to test on, but the [Android AAudio docs on thread safety](https://developer.android.com/ndk/guides/audio/aaudio/aaudio#thread-safety) say (emphasis mine):
[...]

PR #840, opened 2024-02-19
Implement `Send` for `Stream` on certain platforms

bushrat011899:
# Objective

- Implement feature in request #818

## Solution

Moved `NotSendSyncAcrossAllPlatforms` into its own module, `maybe_send` (for clarity) and implemented `Send` for an instance where (I believe) it is safe to do so. I've deliberately left the `cfg` statement controlling this implementation verbose to make it clear how other platforms could be included where it is confirmed to be safe to do so:

```rust
#[cfg(any(
    // Windows with WASAPI allows for a Send Stream
    all(windows, not(feature = "asio"))
))]
unsafe impl Send for NotSendSyncAcrossAllPlatforms {}
```

To add this support to Linux but only on FreeBSD using Jack (as an example) then a single statement can be added without modifying any of the others:

```rust
#[cfg(any(
    // Windows with WASAPI allows for a Send Stream
    all(windows, not(feature = "asio"))
    // FreeBSD using Jack allows for a Send Stream
    all(target_os = "freebsd", feature = "jack"),
))]
unsafe impl Send for NotSendSyncAcrossAllPlatforms {}
```

I have deliberately only implemented `Send` on Windows under WASAPI in this pull request as I believe it is known to be `Send` safe and should be relatively straight-forward to test. I would encourage others to test on platforms they can also confirm the behaviour on and then make follow-up PRs to increase support.

## Motivation

I personally ran into this issue myself when working with Bevy and thought it was a little annoying. Not a deal breaker by any stretch of the imagination, the workarounds are well documented and perfectly reasonable. This is more of a quality of life improvement than anything else.

### DIFF
--- a/src/platform/maybe_send.rs
+++ b/src/platform/maybe_send.rs
@@ -1,0 +1,23 @@
+//! The following zero-sized type is for applying [`Send`]/[`Sync`]` restrictions to ensure
+//! consistent behaviour across different platforms. The verbosely named type is used
+//! (rather than using the markers directly) in the hope of making the compile errors
+//! slightly more helpful.
+
+// TODO: Remove this in favour of using negative trait bounds if they stabilise.
+
+/// A marker used to remove the `Send` and `Sync` traits.
+pub(crate) struct NotSendSyncAcrossAllPlatforms(std::marker::PhantomData<*mut ()>);
+
+impl Default for NotSendSyncAcrossAllPlatforms {
+    fn default() -> Self {
+        NotSendSyncAcrossAllPlatforms(std::marker::PhantomData)
+    }
+}
+
+// TODO: Implement Send on platforms which support it.
+
+#[cfg(any(
+    // Windows with WASAPI allows for a Send Stream
+    all(windows, not(feature = "asio")),
+))]
+unsafe impl Send for NotSendSyncAcrossAllPlatforms {}
--- a/src/platform/mod.rs
+++ b/src/platform/mod.rs
@@ -4,6 +4,10 @@
 //! type and its associated [`Device`], [`Stream`] and other associated types. These
 //! types are useful in the case that users require switching between audio host APIs at runtime.
 
+mod maybe_send;
+
+pub(crate) use maybe_send::*;
+
 #[doc(inline)]
 pub use self::platform_impl::*;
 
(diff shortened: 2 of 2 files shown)


bushrat011899:
As a small test, the `beep` example can be modified like below to demonstrate the send-nature of a `Stream` (on supported platforms):

```rust
pub fn run<T>(device: &cpal::Device, config: &cpal::StreamConfig) -> Result<(), anyhow::Error>
where
    T: SizedSample + FromSample<f32>,
{
    // snip

    // Create the stream on the main thread...
        config,
        move |data: &mut [T], _: &cpal::OutputCallbackInfo| {
            write_data(data, channels, &mut next_value)
        },
        err_fn,
        None,
    )?;

    // ...then send it to a new thread and play it...
    let stream = std::thread::spawn(move || {
        stream.play()?;
    
        std::thread::sleep(std::time::Duration::from_millis(1000));

        Ok::<_, anyhow::Error>(stream)
    });

    // ...get it back and then pause and play it again.
    let stream = stream.join().unwrap()?;

    stream.pause()?;

    std::thread::sleep(std::time::Duration::from_millis(1000));
    
    stream.play()?;

    std::thread::sleep(std::time::Duration::from_millis(1000));

    Ok(())
}
```

n1ght-hunter:
is there anyway we can get this merged?
seems like a really useful PR.

pbitty:
This change does seem very helpful - it would indeed help in my current use-case also.  I am trying to store the stream in a struct field so that it remains in scope and playing, but when I do this I cannot call this struct from any other threads.  I am finding work-arounds though.

However, would it be simpler to omit the global `NotSendSyncAcrossAllPlatforms` "constraint" that is currently added to all Streams, and instead add it as a zero-width field specifically to the Streams that do not support `Send`?

If the intent is to allow `Send` for all streams that support it, it feels like unnecessary complexity to add a blanket `NotSendSyncAcrossAllPlatforms` and then add extra code to opt out of it.

ColorBuffer:
android is the only platform that prevents **Stream** to be **Send**. (am I wrong?)

See #961

est31:
I don't think only changing things for Windows makes sense, it subverts the idea of that type: make us non-Send/Sync across all platforms because some don't implement it. For non-desktop platforms like Android or such I think we can make an exception though.

Issue #106, opened 2016-03-02
Support Exclusive Mode


fb39ca4:

mitchmindtree:
I'm guessing this is referring to the windows backend as I remember briefly seeing some documentation on this when working on the wasapi backend. Could you elaborate on how exclusive mode behaves differently from shared mode? Do you see a path towards supporting this for all backends?

tomaka:
When in exclusive mode you can write (or read, I guess) directly to (from) the hardware. It is therefore faster than shared mode, but no other application can access the device in the meanwhile, and we cannot access it if another application already uses it in shared mode.

mitchmindtree:
Ah, thanks for the explanation. I guess this might be useful for some pro-audio software like DAWs and live performance applications. That said, I don't remember seeing anything like this in the coreaudio or alsa backends.

I guess if this were to be implemented, it would be an addition of a whole new `wasapi_exclusive` backend, perhaps enabled by a feature?

Alternatively, we could introduce something similar to PortAudio's `Host` API? I've opened up #204 for discussion.

blueneogeo:
CoreAudio also has a hog mode property: [kAudioDevicePropertyHogMode](https://developer.apple.com/documentation/coreaudio/kaudiodevicepropertyhogmode)

Hog mode is important if you go for the best audio quality.  For example, streaming software clients such as Spotify and Tidal have hog mode support. Spotify turns this on by default.

blueneogeo:
Any update on this? Or is it already implemented in one of the other closed tickets?

PR #843, opened 2024-02-21
Added WASAPI exclusive host

jesnor:
In exclusive mode only one stream can be opened per device. This is different from the ASIO host which does its own mixing of multiple streams using the sample format of the device. I don't think this is a good idea because it can lead to audio artifacts like lack of precision and overflow. So it's better to let the application do the mixing, preferably using f32 or f64 samples which are then converted to the device sample format.

Also removed duplicated code in the WASAPI `Device` methods for creating input/output streams.

### DIFF
--- a/src/host/mod.rs
+++ b/src/host/mod.rs
@@ -26,5 +26,7 @@ pub(crate) mod null;
 pub(crate) mod oboe;
 #[cfg(windows)]
 pub(crate) mod wasapi;
+#[cfg(windows)]
+pub(crate) mod wasapi_exclusive;
 #[cfg(all(target_arch = "wasm32", feature = "wasm-bindgen"))]
 pub(crate) mod webaudio;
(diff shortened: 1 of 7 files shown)


JerwuQu:
This is great! I have use for this in an application that requires low-latency audio.

Is there anything currently blocking this? (aside from merge conflicts)

jesnor:
This PR needs to updated and bug fixed. I could create a new PR later if there interest in adding this feature to cpal.

JerwuQu:
Given that there are two open issues #156 and #106 and external projects also waiting for this feature (like @Drewol above, and my own), I'd say there is interest.

I tried the latest version of this branch after rebasing on master, and it seems to have worked with output/speaker but not input/microphone (something about invalid stream format despite using a supposedly supported one), but maybe that's related to the bugs you mentioned.

Just tag me if you want help or testing with anything!

PR #846, opened 2024-02-24
Remove parking_lot dependency

james7132:
The standard library is using a similar implementation to parking_lot now. Other than APIs that are unstable, there shouldn't be that much of a reason to rely on parking_lot anymore. Parts of the cpal codebase already are using std locks. This PR removes the remaining parking_lot uses and replaces them with the standard library's.


est31:
Thanks for the PR! There is some CI errors that seem relevant to your changes, could you check them out?

est31 (approved):
Thanks!

PR #886, opened 2024-05-12
ALSA: Don't panic when handling invalid stream timestamps.

jwagner:
For some reason the timestamps returned by ALSA when using the onboard ALC1220 are invalid (0) causing a panic in cpal, crashing my application.

 I don't understand ALSA enough to judge why this is happening. This fix only addresses the symptoms by returning an Err instead of panicing. I couldn't find any reason why a panic would be necessary in this case, but might have missed something.


est31 (approved):
yeah if it can come from a buggy driver we shouldn't panic but should treat it like an error instead.

PR #887, opened 2024-05-20
PipeWire implementation

valaphee:
This PR implements PW,

it works by having a main loop and use uni-directional channels to pass control messages to the main loop.

There are some open questions, and I use jack as foundation. For example there is not really any equivalent for devices, maybe it can be set that a device is the node the newly created node connects to.

Or its just like Jack, that there is a default device, and the user can create devices. But each individual stream should be a node.

As I can't test with ALSA, when a stream is dropped it gets released and its a permanent operation in contrary to play and pause which are temporary. Theoretically play and pause could also be mapped to connect, disconnect, but idk if this is correct.

Fixes https://github.com/RustAudio/cpal/issues/554

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,6 +46,7 @@ num-traits = { version = "0.2.6", optional = true }
 alsa = "0.9"
 libc = "0.2"
 jack = { version = "0.11", optional = true }
+pipewire = { version = "0.8" }
 
 [target.'cfg(any(target_os = "macos", target_os = "ios"))'.dependencies]
 core-foundation-sys = "0.8.2" # For linking to CoreFoundation.framework and handling device name `CFString`s.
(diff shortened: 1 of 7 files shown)


valaphee:
Its WIP because of the open question and it needs some polishing, I also temporary didn't mark it as separate feature as otherwise my code completion doesn't work.

jacksongoode:
Would this mean that cpal could play without pipewire-alsa installed?

Ralith:
>  I also temporary didn't mark it as separate feature as otherwise my code completion doesn't work.

You should set your code completion provider to enable all features, or at least the features you're developing.

cynthia2006:
And also, play and pause don't map no connect and disconnect; rather active and inactive. Sure, the stream could be configured in a way that the session manager terminates the connection when it is idle.

PR #893, opened 2024-06-26
Adopt windows version range >=0.54, <=0.57

cart:
`windows` is a "heavy" crate. Duplicate `windows` versions in a tree can add precious seconds to compile times (or even _minutes_ by some Bevy user reports). On the Bevy side, we are trying to ensure that `windows` is only compiled once. We also believe it is in the ecosystem's best interest (and in the interest of crate owners that consume `windows`) to align their versioning with the rest of ecosystem. Therefore, I would like to encourage crate owners to adopt the following approach to `windows` crate usage:

1. Adopt "range versioning" for windows instead of locking to a specific version.
2. Before adding a new version to the range, test that each previous version in the range still compiles.
3. Only remove old versions from the range when _absolutely necessary_. Try your hardest to ensure that your version support aligns with other ecosystem crates.

We've picked 0.54 as a baseline because breaking API changes were made to use Rust Results instead of HRESULT.

I have personally tested that cpal compiles with 0.54, 0.56 (0.55 does not exist), and 0.57.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,7 +26,7 @@ clap = { version = "4.0", features = ["derive"] }
 ndk-glue = "0.7"
 
 [target.'cfg(target_os = "windows")'.dependencies]
-windows = { version = "0.54.0", features = [
+windows = { version = ">=0.54, <=0.57.0", features = [
     "Win32_Media_Audio",
     "Win32_Foundation",
     "Win32_Devices_Properties",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -26,6 +26,8 @@ clap = { version = "4.0", features = ["derive"] }
 ndk-glue = "0.7"
 
 [target.'cfg(target_os = "windows")'.dependencies]
+# Support a range of versions in order to avoid duplication of this crate. Make sure to test all
+# versions when bumping to a new release, and only increase the minimum when absolutely necessary.
 windows = { version = ">=0.54, <=0.57.0", features = [
     "Win32_Media_Audio",
     "Win32_Foundation",


Cargo.toml
```diff
@@ -26,7 +26,7 @@ clap = { version = "4.0", features = ["derive"] }
 ndk-glue = "0.7"
 
 [target.'cfg(target_os = "windows")'.dependencies]
-windows = { version = "0.54.0", features = [
+windows = { version = ">=0.54, <=0.57.0", features = [
```
BD103, line 31:
```suggestion
# Support a range of versions in order to avoid duplication of this crate. Make sure to test all
# versions when bumping to a new release, and only increase the minimum when absolutely necessary.
windows = { version = ">=0.54, <=0.57.0", features = [
```

This is copied from my review of https://github.com/GuillaumeGomez/sysinfo/pull/1295. I don't want the reasoning behind this change to be lost.

Ralith:
I think this needs CI checking compatibility with the oldest listed version to not be a liability.

BD103:
> I think this needs CI checking compatibility with the oldest listed version to not be a liability.

The [`windows-test`](https://github.com/RustAudio/cpal/blob/0246442da2f401895d2c82bbf941e3ebf6e93a04/.github/workflows/cpal.yml#L153-L156) job could be modified with a matrix to test this. I know its possible on to resolve the minimum versions of all dependencies on nightly, though I don't know if you can resolve just a single dependency. (Perhaps you need to modify the lockfile or patch in the version?)

est31:
These ranges are a good idea to the too many windows versions problem, so good catch!

It's kinda outsourcing the problem of checking for compatibility to the community, but still better than having multiple versions of windows in your crate graph.

I agree with @Ralith, there should be CI checks for all of the supported windows versions, and we'd probably still only want to support a limited amount to keep the number of CI checked versions down (but ideally 3-4 different versions would sufficiently avoid duplication). You'd probably do something like `cargo update --precise` to get the needed windows version.

BD103:
> You'd probably do something like `cargo update --precise` to get the needed windows version.

I tried using `cargo update --precise`, but I had some issues with duplicate dependencies. The one solution that worked for me was using patch files, though I'm sure Regex would be more stable.

JMS55:
Are there any blockers for this? Bevy currently has 4(!) different versions of `windows` in it's dependency tree, with cpal contributing the oldest. It should be fairly straightforward to merge this and publish a new version of cpal + rodio, no?

est31:
as I've said, we should add some way in CI to test for all windows versions.

roderickvd:
Through #1020 I had updated `windows` to "latest & greatest" v0.62, not having read this. Indeed I had to update some parts of the API to accommodate.

Is it still current that we truly need a `windows` version below latest?

est31:
There have been 3 major releases this year so far... I don't think one can get yearly updates to users that quickly. Better to depend on a 12 month window if possible.

roderickvd:
> There have been 3 major releases this year so far... I don't think one can get yearly updates to users that quickly. Better to depend on a 12 month window if possible.

Superseded by #1023 that takes the approach of this PR and tests it in the CI too. Thanks for the PR @cart!

PR #899, opened 2024-07-12
Allow creation of OutputCallbackInfo and InputCallbackInfo objects (for test purposes)

danutsu:
This change makes the fields of `StreamInstant`, `OutputCallbackInfo` and `InputCallbackInfo` pub.

The intent is to allow `OutputCallbackInfo` instances to be created, which then allows unit testing the data callback function. Because it is a self-contained function that simply fills data in a buffer, it is, in principle, very unit-testable -- but it's very difficult to do so today because the test code cannot create an `OutputCallbackInfo` for it.

I don't believe there's any good reason for these fields to not be pub. In particular, I noticed the fields of `OutputStreamTimestamp` _are_ public, so the "middle" layer is already creatable, just not `StreamInstant` and `OutputCallbackInfo`. Also, the structures are pretty plain data objects with no special logic or anything like that, so there really is no harm in allowing them to be created.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -363,14 +363,14 @@ pub struct OutputStreamTimestamp {
 /// Information relevant to a single call to the user's input stream data callback.
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct InputCallbackInfo {
-    timestamp: InputStreamTimestamp,
+    pub timestamp: InputStreamTimestamp,
 }
 
 /// Information relevant to a single call to the user's output stream data callback.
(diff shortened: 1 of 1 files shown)


danutsu:
Fixed the emscripten issue, please take a look and let me know if I need to do anything else.

roderickvd:
It's been a while - sorry for that - but new maintainer here and I follow your rationale. Would you resolve the conflicts and update the PR?

roderickvd:
For consideration: instead of making everything `pub`, how about offering `new` constructors and some getters?

LastExceed:
superseded by #1139 i think?

PR #914, opened 2024-09-21
ALSA: fix buffer size / period size configuration, JACK: update to 0.13.0 and fix compilation errors

### LINKED ISSUE #913: Invalid period size specification for ALSA?
https://github.com/RustAudio/cpal/blob/6ecfec4ed0e4e31f816f90831014f3a701de2e7d/src/host/alsa/mod.rs#L1066

Hi,
The code above seems wrong to me.
As I understand it, a cpal buffer size correspond to an alsa period size.
So the alsa buffer size should be: `nperiods * cpal_buffer_size`.

While it is currently `cpal_buffer_size / 4` and I suppose that `nperiods == 4`.

I think that better defaults for cpal would be a `nperiods == 2`.

Cheers,
Alex

abique:
With ALSA, the buffer size is `nperiod * period_size`. We want 2 periods.
The default was far too large, 1024 samples is a much more reasonable default (still too large for real-time audio).

Fixes #913


abique:
I can confirm that this fixes intialization of Hercules DJControl Inpulse 500 on Linux with ALSA.

abique:
I've split this pull request into 4 ones, for easier review and accept/reject.

See:
- #915 
- #916 
- #917 
- #918

PR #916, opened 2024-09-24
ALSA: better card enumeration

abique:
- better device name
- list physical devices
- inject most common virtual ones (default, pipewire, pulse, ...)

Replaces #914


est31 (approved):

abique:
:heart:

PR #918, opened 2024-09-24
Intial support for 24 bits sample format (24bit stored on 4 bytes)

abique:
There are two common ways of storing 24 bits samples:
- on 4 bytes (called 24, most common)
- on 3 bytes (called 24_3)

With the current code, 3 bytes seems difficult to implement and would require a larger code refactoring.

That being said, having at least '24' helps to work with hardware that only offer it.

Replaces #914


est31:
Hi, I've since merged #927, which adds support for 24 bit asio samples. Are you interested in adding support for other backends?

abique:
Hi @est31 ,

Good that you've made progress on the 24 bit samples, it did prevent some hardware (that only worked with 24 bits samples) to function.

Though, I don't plan to do any further changes to cpal given that https://github.com/RustAudio/cpal/pull/917 didn't get any feedback so far.

Cheers,
Alex

abique:
I'd suggest to have float in/out and keep the complexity of sample encoding as an internal thing.
You won't be able to work with more complicated encoding this way:
- bit-stream
- 24 bits encoded on 3 or 4 bytes
- big/little endian

roderickvd:
Superseded by #1011 based on this work. Thanks!

Issue #802, opened 2023-10-13
Memory leak in WASAPI `input_devices`


nicolaiunrein:
The below code is causing a memory leak on my machine:

```
use cpal::traits::HostTrait;

fn main() {
    let host = cpal::ALL_HOSTS
        .into_iter()
        .find(|id| id.name() == "WASAPI")
        .and_then(|id| cpal::host_from_id(*id).ok())
        .expect("No WASAPI host found");

    loop {
        let _ = host.input_devices().unwrap().collect::<Vec<_>>();
        std::thread::sleep(std::time::Duration::from_millis(10));
    }
}
```
## System Info
- OS: Running Microsoft Windows 11 Pro
- Version: 10.0.22621 Build 22621
- cpal version: 0.15.2
- Toolchain: stable-x86_64-pc-windows-msvc

rustc -Vv output
```
rustc 1.72.0 (5680fa18f 2023-08-23)
binary: rustc
commit-hash: 5680fa18feaa87f3ff04063800aec256c3d4b4be
commit-date: 2023-08-23
host: x86_64-pc-windows-msvc
release: 1.72.0
LLVM version: 16.0.5
```

chanderlud:
I am experiencing the same issue
### System Info
- OS: Windows 11 Home
- Version: 23H2 Build 22631.3737 
- cpal version: 0.15.3
- Toolchain: stable-x86_64-pc-windows-msvc

rustc -Vv output
```
rustc 1.78.0 (9b00956e5 2024-04-29)
binary: rustc
commit-hash: 9b00956e56009bab2aa15d7bff10916599e3d6d6
commit-date: 2024-04-29
host: x86_64-pc-windows-msvc
release: 1.78.0
LLVM version: 18.1.2
```

PR #919, opened 2024-10-08
Fixed WASAPI memory leak

jesnor:
Fixed memory leak for parameter ppClosestMatch in call to IAudioClient::IsFormatSupported. Fixes issue #802.


nicolaiunrein:
This may fix [#802](https://github.com/RustAudio/cpal/issues/802)

est31 (approved):

PR #924, opened 2024-11-02
Make all error enums `Clone`

abey79:
Make all error enums `Clone` for convenience.

Context: I'm wrapping most of these in a `thiserror::Error` and not being able to make it `Clone` occasionally makes error management a bit annoying.


abique:
LGTM

abique (approved):

est31 (approved):

Issue #810, opened 2023-11-13
emscripten support is broken (since wasm-bindgen does not work there?)


benjamin-sieffert:
When I try to use cpal (via rodio) in an emscripten build, I run into the following error:
`"cannot call wasm-bindgen imported functions on non-wasm targets" @ /home/ax/.cargo/registry/src/index.crates.io-6f17d22bba15001f/web-sys-0.3.64/src/features/gen_AudioContext.rs:5`
Afaik, `wasm-bindgen` fundamentally does not work on emscripten, at least not without jumping through a lot of hoops.

Before #713 , "stdweb" was used, which advertises direct emscripten support; so it probably (somewhat) worked at some point.
Unfortunately the old version(s) can no longer be compiled, due to various other changes in rustc/libc/emscripten.

According to the emscripten docs, "OpenAL" should be the best way to get audio in that environment: https://emscripten.org/docs/porting/Audio.html

benjamin-sieffert:
Since it appears I am the only developer on emscripten, I suggest simply removing the advertised emscripten support from the docs, to not bait any more victims. ;)
As for myself, I will see if I can’t just use the SDL2_mixer extension.

alisomay:
https://github.com/RustAudio/cpal/pull/809#issuecomment-1793875672

est31:
Maybe if @alisomay  made a PR, @benjamin-sieffert could try it out? Then we can see if we can merge it or not.

benjamin-sieffert:
@alisomay @est31
I think this lone compilation failure is not the main issue with emscripten. The whole cpal-for-emscripten implementation is based on `wasm-bindgen` through and through, but that doesn’t work on emscripten at runtime!! It compiles because the compiler finds the right cpu arch ("wasm32"), but in practice the environment created by the emscripten wrapper inside the browser does look very different than the normal wasm environment.
https://rustwasm.github.io/wasm-bindgen/reference/rust-targets.html
>The wasm-bindgen target does not support the wasm32-unknown-emscripten nor the asmjs-unknown-emscripten targets. There are currently no plans to support these targets either. All annotations work like other platforms on the targets, retaining exported functions and causing all imports to panic.

I can try your patch instead of mine, but I doubt it helps with the runtime failure.
[...]

PR #934, opened 2025-01-09
Fix infinite loop on broken ALSA pipes

cdellacqua:
When an ALSA interface ends up in an unrecoverable error state (e.g. device disconnected), the thread that runs `input_stream_worker` (or `output_stream_worker` depending on the type of stream) goes into an infinite loop due to the polling always returning an error.

This infinite loop causes calls to `drop(stream)` to hang indefinitely as the Drop implementation tries to join the polling thread.

## Proposed solution

By adding a flag that's raised when a `Stream` is dropped, we can add an early exit condition to `poll_descriptors_and_prepare_buffer` that forces a return value of `PollDescriptorsFlow::Return`.

## Testing

I successfully tested the fix on my local machine by unplugging the microphone while listening to it.

### DIFF
--- a/src/host/alsa/mod.rs
+++ b/src/host/alsa/mod.rs
@@ -10,6 +10,7 @@ use crate::{
     SupportedBufferSize, SupportedStreamConfig, SupportedStreamConfigRange,
     SupportedStreamConfigsError,
 };
+use std::cell::Cell;
 use std::cmp;
 use std::convert::TryInto;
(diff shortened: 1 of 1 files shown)


cdellacqua:
Note: the failed pipeline is wasm/emscripten-specific. The error looks quite similar to the one described in issue #810 but it's out of scope for this PR

est31 (approved):

PR #943, opened 2025-01-26
Update `coreaudio-rs`

madsmtm:
Corresponding PR to https://github.com/RustAudio/coreaudio-rs/pull/128. Will finish this once that merges and is released.


madsmtm:
@est31: 0.13.0 has been released, and I've rebased this to use that dependency instead

Issue #939, opened 2025-01-18
proposal: allow setting thread priority for stream threads


Sorseg:
When my CPU is under high load, the sound thread does not get enough compute time, which leads to buffer underruns and clicking.
Setting high priority on sound producing threads allows producing uninterrupted sound even with a high load, but this is not straightforward with the current API.

## Do we want this to be configurable?
Maybe. There might be some environments where there should be limited number of realtime threads and stuttering sound is less of a priority.

## If we do want to configure the priority, what is the API?
I see two paths: 
1. add an argument to the `build_input_stream`
2. add another method `build_input_stream_with_priority`

Thank you for maintaining this library!

edwloef:
Maybe CPAL should always set the audio thread to realtime via something like https://github.com/mozilla/audio_thread_priority, or make this at least be an option.

Ralith:
This should certainly be the default when cpal spawns a thread itself. A config option should not be added unless there's a clear use case for one.

edwloef:
Looks like this is already done on WASAPI, but not on other platforms? I'm honestly not sure.

https://github.com/RustAudio/cpal/blob/4a8c9b8ea6566b1fcca1bf7b816464c3151e7230/src/host/wasapi/stream.rs#L325-L334

Sorseg:
I am using ALSA host, which seems to just spawn a thread
https://github.com/RustAudio/cpal/blob/4a8c9b8ea6566b1fcca1bf7b816464c3151e7230/src/host/alsa/mod.rs#L942-L960

Sorseg:
One issue I discovered is that at least in my Ubuntu installation, user processes do not have permission to set both `nice` and `rtprio` thread priority. Setting rtprio up to 95 is granted to the `audio` group, which the user does not belong by default.

```
unsafe { libc::nice(-5) } = -1
unsafe { *libc::__errno_location() } = 1
```
(1 is `EPERM`)

jangler:
> Looks like this is already done on WASAPI, but not on other platforms? I'm honestly not sure.

In a way it is, but I get much more reliable WASAPI performance using the `audio_thread_priority` crate than with `THREAD_PRIORITY_TIME_CRITICAL`.

est31:
Done in #955

PR #955, opened 2025-02-22
set realtime priority for stream threads in alsa and wasapi

edwloef:
implements #939. As someone who doesn't know a whole lot about cpal's inner workings it looks to me like these are the only threads that the library spawns itself, leading me to think these are the only threads that require manual priority changes. However, `audio_thread_priority` claims to also support MacOS, so maybe someone could look into whether that's applicable here?

I'm not too massive of a fan of `eprintln!` here, however I think the failure should be shown but not prevent the audio thread from running.

I've tested this on my Linux laptop, some verification that this also works for the folks on WASAPI and the BSD's would be nice.

### DIFF
--- a/.github/workflows/cpal.yml
+++ b/.github/workflows/cpal.yml
@@ -14,6 +14,8 @@ jobs:
       run: sudo apt-get install libasound2-dev
     - name: Install libjack
       run: sudo apt-get install libjack-jackd2-dev libjack-jackd2-0
+    - name: Install dbus
+      run: sudo apt-get install libdbus-1-dev
     - name: Install stable
       uses: dtolnay/rust-toolchain@stable
       with:
@@ -67,6 +69,8 @@ jobs:
       run: sudo apt-get install libasound2-dev
     - name: Install libjack
       run: sudo apt-get install libjack-jackd2-dev libjack-jackd2-0
+    - name: Install dbus
+      run: sudo apt-get install libdbus-1-dev
     - name: Install stable
       uses: dtolnay/rust-toolchain@stable
     - name: Run without features
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,14 +35,14 @@ windows = { version = "0.54.0", features = [
     "Win32_Media_Multimedia",
     "Win32_UI_Shell_PropertiesSystem"
 ]}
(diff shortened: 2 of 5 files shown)


edwloef:
Seems like CI is failing due to dbus not being installed on the linux runners.

edwloef:
Alright, windows is fixed. Should I update the CI and readme to include dbus on linux?

jacksongoode:
I believe this will resolve https://github.com/jpochyla/psst/issues/235 and https://github.com/jpochyla/psst/issues/251. Anything blocking this right now?

marcpabst:
Just as a note, I belive any paramters to `promote_current_thread_to_real_time` are discarded on Windows. That said, can we get this reviewed/merge?

edwloef:
I've added dbus to the CI and mentioned it in the readme. I *think* the wasm failure isn't related to the changes I've made, so as far as I can tell everything should be good to go.

est31:
Can you make this optional (default off)? I don't think we should require an extra C dependency from users unconditionally.

edwloef:
Should it also be made optional on windows as well, or just linux?

jacksongoode:
> Can you make this optional 

Should it be optional? I feel as if this should be baked for Windows since otherwise audio streams can become mangled under high load?

edwloef:
> > Can you make this optional
> 
> Should it be optional? I feel as if this should be baked for Windows since otherwise audio streams can become mangled under high load?

If it's optional I can just leave the solution for people who don't enable the feature. The license (MPL) might be an issue for some people, so I can envision a use case for it being optional, which is what I've just done for now.

Ralith:
If this is a C dependency, we should just vendor equivalent logic. It should only be a few syscalls on any platform. Requiring a feature to be enabled to get the correct behavior is a bad user experience.

edwloef:
`audio_thread_priority` provides a dbus-less implementation as well, but I can't tell whether it actually does anything or is just a stub. Maybe re-implementing the relevant portions with something like zbus would be reasonable, but zbus is humongous and I'm not sure whether that's much better than a C dependency.

jacksongoode:
@est31 Sorry for the ping, but I think this one is ready and looked over quite well :)

est31 (approved):

PR #958, opened 2025-03-01
AudioWorklet based Host for web when Wasm atomics are enabled

kettle11:
This pull request adds an optional feature-flagged `Host` that uses [AudioWorklet](https://developer.mozilla.org/en-US/docs/Web/API/AudioWorklet) on web when the `atomics` feature is enabled. When the `web_audio_worklet` and `wasm-bindgen` features are enabled and nightly Rust is compiled with the `atomics` flag enabled the WebAudioWorklet Host is available.

`AudioWorklet` should be lower latency, lower overhead, and less prone to jank. This implementations works pretty much the same as a native audio thread, allowing the implementation to be simpler than the existing WebAudio Host implementation.

To used shared memory it's also required (for security reasons) that the server hosting the Wasm to be configured with the correct CORS headers. I've added an example that shows how to build with the correct Rust flags and how to set the Cors headers in a test environment.

So far I've tested this in Chrome, Safari, and Firefox with the beep example and it seems to be working fine.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,10 +10,22 @@ edition = "2021"
 rust-version = "1.70"
 
 [features]
+default = []
 asio = ["asio-sys", "num-traits"] # Only available on Windows. See README for setup instructions.
 oboe-shared-stdcxx = ["oboe/shared-stdcxx"] # Only available on Android. See README for what it does.
+# Only available on web when atomics are enabled. See README for what it does.
+web_audio_worklet = [
+    "wasm-bindgen-futures",
+    "web-sys/Blob",
+    "web-sys/BlobPropertyBag",
+    "web-sys/Url",
+    "web-sys/AudioWorklet",
+    "web-sys/AudioWorkletNode",
+    "web-sys/AudioWorkletNodeOptions",
+]
 
 [dependencies]
+wasm-bindgen-futures = {version = "0.4.33", optional = true}
 dasp_sample = "0.11"
 
 [dev-dependencies]
--- a/examples/web-audio-worklet-beep/Cargo.toml
+++ b/examples/web-audio-worklet-beep/Cargo.toml
@@ -1,0 +1,40 @@
+[package]
+name = "web-audio-worklet-beep"
+description = "cpal beep example for WebAssembly on an AudioWorklet"
+version = "0.1.0"
+edition = "2018"
+
+[lib]
+crate-type = ["cdylib"]
+
+[profile.release]
+# This makes the compiled code faster and smaller, but it makes compiling slower,
+# so it's only enabled in release mode.
+lto = true
+
+[features]
+# If you uncomment this line, it will enable `wee_alloc`:
+#default = ["wee_alloc"]
+
+[dependencies]
+cpal = { path = "../..", features = ["wasm-bindgen", "web_audio_worklet"] }
+# `gloo` is a utility crate which improves ergonomics over direct `web-sys` usage.
+gloo = "0.11.0"
+# The `wasm-bindgen` crate provides the bare minimum functionality needed
+# to interact with JavaScript.
+wasm-bindgen = "0.2.45"
+
+# `wee_alloc` is a tiny allocator for wasm that is only ~1K in code size
+# compared to the default allocator's ~10K. However, it is slower than the default
+# allocator, so it's not enabled by default.
+wee_alloc = { version = "0.4.2", optional = true }
+
+# The `console_error_panic_hook` crate provides better debugging of panics by
+# logging them with `console.error`.
+console_error_panic_hook = "0.1.5"
+
+# The `web-sys` crate allows you to interact with the various browser APIs,
+# like the DOM.
+[dependencies.web-sys]
+version = "0.3.22"
+features = ["console", "MouseEvent"]
--- a/examples/web-audio-worklet-beep/src/lib.rs
+++ b/examples/web-audio-worklet-beep/src/lib.rs
@@ -1,0 +1,114 @@
+use std::{cell::Cell, rc::Rc};
+
+use cpal::{
+    traits::{DeviceTrait, HostTrait, StreamTrait},
+    Stream,
+};
+use wasm_bindgen::prelude::*;
+use web_sys::console;
+
+// When the `wee_alloc` feature is enabled, this uses `wee_alloc` as the global
+// allocator.
+//
+// If you don't want to use `wee_alloc`, you can safely delete this.
+#[cfg(feature = "wee_alloc")]
+#[global_allocator]
+static ALLOC: wee_alloc::WeeAlloc = wee_alloc::WeeAlloc::INIT;
+
+// This is like the `main` function, except for JavaScript.
+#[wasm_bindgen(start)]
+pub fn main_js() -> Result<(), JsValue> {
+    // This provides better error messages in debug mode.
+    // It's disabled in release mode, so it doesn't bloat up the file size.
+    #[cfg(debug_assertions)]
+    console_error_panic_hook::set_once();
+
+    let document = gloo::utils::document();
+    let play_button = document.get_element_by_id("play").unwrap();
+    let stop_button = document.get_element_by_id("stop").unwrap();
+
+    // stream needs to be referenced from the "play" and "stop" closures
+    let stream = Rc::new(Cell::new(None));
+
+    // set up play button
+    {
+        let stream = stream.clone();
+        let closure = Closure::<dyn FnMut(_)>::new(move |_event: web_sys::MouseEvent| {
+            stream.set(Some(beep()));
+        });
+        play_button
+            .add_event_listener_with_callback("mousedown", closure.as_ref().unchecked_ref())?;
+        closure.forget();
+    }
+
+    // set up stop button
+    {
+        let closure = Closure::<dyn FnMut(_)>::new(move |_event: web_sys::MouseEvent| {
+            // stop the stream by dropping it
+            stream.take();
+        });
+        stop_button
+            .add_event_listener_with_callback("mousedown", closure.as_ref().unchecked_ref())?;
+        closure.forget();
+    }
+
+    Ok(())
+}
+
+fn beep() -> Stream {
+    let host = cpal::host_from_id(cpal::HostId::WebAudioWorklet)
+        .expect("WebAudioWorklet host not available");
+
+    let device = host
+        .default_output_device()
+        .expect("failed to find a default output device");
+    let config = device.default_output_config().unwrap();
+
+    match config.sample_format() {
+        cpal::SampleFormat::F32 => run::<f32>(&device, &config.into()),
+        cpal::SampleFormat::I16 => run::<i16>(&device, &config.into()),
+        cpal::SampleFormat::U16 => run::<u16>(&device, &config.into()),
+        _ => panic!("unsupported sample format"),
+    }
+}
+
+fn run<T>(device: &cpal::Device, config: &cpal::StreamConfig) -> Stream
+where
+    T: cpal::Sample + cpal::SizedSample + cpal::FromSample<f32>,
+{
+    let sample_rate = config.sample_rate.0 as f32;
+    let channels = config.channels as usize;
+
+    // Produce a sinusoid of maximum amplitude.
+    let mut sample_clock = 0f32;
+    let mut next_value = move || {
+        sample_clock = (sample_clock + 1.0) % sample_rate;
+        (sample_clock * 440.0 * 2.0 * 3.141592 / sample_rate).sin()
+    };
+
+    let err_fn = |err| console::error_1(&format!("an error occurred on stream: {}", err).into());
+
+    let stream = device
+        .build_output_stream(
+            config,
+            move |data: &mut [T], _| write_data(data, channels, &mut next_value),
+            err_fn,
+            None,
+        )
+        .unwrap();
+    stream.play().unwrap();
+    stream
+}
+
+fn write_data<T>(output: &mut [T], channels: usize, next_sample: &mut dyn FnMut() -> f32)
+where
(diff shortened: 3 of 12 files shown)


roderickvd:
Very nice addition! This is a big one, so I'm gonna first trigger a Copilot review to walk through major points, if any, then do a further code review.

Apologies for the late response - new collaborator here and grooming through the backlog now.

kettle11:
Thanks for following up!

I applied some of Copilot's suggestions and also realized that some files needed to build the example were missing so I fixed that as well. Let me know if anything further is needed.

Cargo.toml
```diff
@@ -10,8 +10,19 @@ edition = "2021"
 rust-version = "1.70"
 
 [features]
+default = []
 asio = ["asio-sys", "num-traits"] # Only available on Windows. See README for setup instructions.
 oboe-shared-stdcxx = ["oboe/shared-stdcxx"] # Only available on Android. See README for what it does.
+# Only available on web when atomics are enabled. See README for what it does.
+web_audio_worklet = [
```
roderickvd, line 19: Please document the feature flag in the README.

Cargo.toml
```diff
@@ -65,6 +76,7 @@ web-sys = { version = "0.3.35", features = [ "AudioContext", "AudioContextOption
 
 [target.'cfg(all(target_arch = "wasm32", target_os = "unknown"))'.dependencies]
 wasm-bindgen = { version = "0.2.58", optional = true }
+wasm-bindgen-futures = {version = "0.4.33", optional = true}
```
roderickvd: Style: misses spaces between the curly braces.

examples/web-audio-worklet-beep/.cargo/config.toml
```diff
@@ -0,0 +1,5 @@
+[target.wasm32-unknown-unknown]
+rustflags = ["-C", "target-feature=+atomics"]
+
+[unstable]
+build-std = ["std", "panic_abort"]
```
roderickvd, line 5: Does it have to abort? Could we leave that choice to the user?

.gitignore
```diff
@@ -1,6 +1,8 @@
-/target
-/Cargo.lock
-.cargo/
-.DS_Store
-recorded.wav
-rls*.log
+/target
```
roderickvd, line 1: Why does it give a diff for the first six lines? (don't bother too much, just wondering)

examples/web-audio-worklet-beep/Cargo.toml
```diff
@@ -0,0 +1,40 @@
+[package]
+name = "web-audio-worklet-beep"
```
roderickvd, line 2: Feel free to put your name as author.

examples/web-audio-worklet-beep/src/lib.rs
```diff
@@ -0,0 +1,114 @@
...
+// This is like the `main` function, except for JavaScript.
+#[wasm_bindgen(start)]
+pub fn main_js() -> Result<(), JsValue> {
+    // This provides better error messages in debug mode.
+    // It's disabled in release mode, so it doesn't bloat up the file size.
+    #[cfg(debug_assertions)]
+    console_error_panic_hook::set_once();
+
+    let document = gloo::utils::document();
+    let play_button = document.get_element_by_id("play").unwrap();
```
roderickvd, line 19: As you're return a `Result`, could you remove the `unwrap` here and elsewhere?

examples/web-audio-worklet-beep/src/lib.rs
```diff
@@ -0,0 +1,114 @@
...
+    T: cpal::Sample + cpal::SizedSample + cpal::FromSample<f32>,
+{
+    let sample_rate = config.sample_rate.0 as f32;
+    let channels = config.channels as usize;
+
+    // Produce a sinusoid of maximum amplitude.
+    let mut sample_clock = 0f32;
+    let mut next_value = move || {
+        sample_clock = (sample_clock + 1.0) % sample_rate;
+        (sample_clock * 440.0 * 2.0 * 3.141592 / sample_rate).sin()
```
roderickvd: There's `std::f32::consts::PI`.

src/host/web_audio_worklet/dependent_module.rs
```diff
@@ -0,0 +1,62 @@
...
+//   https://github.com/rustwasm/wasm-bindgen/blob/main/examples/wasm-audio-worklet/src/dependent_module.rs
+//
+// The original code is licensed under either of:
+//   - MIT license (https://opensource.org/licenses/MIT)
+//   - Apache License, Version 2.0 (https://www.apache.org/licenses/LICENSE-2.0)
+// at your option.
+//
+// Copyright (c) 2017-2024 The wasm-bindgen Developers
+//
+// This file incorporates code from the above source under the terms of those licenses.
```
roderickvd: We should be explicit that we opt for the Apache license, since that's what's `cpal` itself is licensed under.

src/host/web_audio_worklet/mod.rs
```diff
@@ -0,0 +1,410 @@
...
+
+const MIN_CHANNELS: u16 = 1;
+const MAX_CHANNELS: u16 = 32;
+const MIN_SAMPLE_RATE: SampleRate = SampleRate(8_000);
+const MAX_SAMPLE_RATE: SampleRate = SampleRate(96_000);
+const DEFAULT_SAMPLE_RATE: SampleRate = SampleRate(44_100);
+const SUPPORTED_SAMPLE_FORMAT: SampleFormat = SampleFormat::F32;
+
+impl Host {
+    pub fn new() -> Result<Self, crate::HostUnavailable> {
```
roderickvd, line 39: Why have a `crate::JostUnavailable` error type when it seems infallible?

src/host/web_audio_worklet/mod.rs
```diff
@@ -0,0 +1,410 @@
...
+pub struct Host;
+
+pub struct Stream {
+    audio_context: web_sys::AudioContext,
+}
+
+pub type SupportedInputConfigs = ::std::vec::IntoIter<SupportedStreamConfigRange>;
+pub type SupportedOutputConfigs = ::std::vec::IntoIter<SupportedStreamConfigRange>;
+
+const MIN_CHANNELS: u16 = 1;
```
roderickvd: You can use the `cpal::ChannelCount` type alias instead of `u16`.

src/host/web_audio_worklet/mod.rs
```diff
@@ -0,0 +1,410 @@
...
+        }
+
+        let config = config.clone();
+
+        let stream_opts = web_sys::AudioContextOptions::new();
+        stream_opts.set_sample_rate(config.sample_rate.0 as f32);
+
+        let audio_context = web_sys::AudioContext::new_with_context_options(&stream_opts).map_err(
+            |err| -> BuildStreamError {
+                let description = format!("{:?}", err);
```
roderickvd: Does this pass clippy? Preferred syntax has become `format!("{err:?}")`.

src/platform/mod.rs
```diff
@@ -662,7 +662,15 @@ mod platform_impl {
         SupportedOutputConfigs as WebAudioSupportedOutputConfigs,
     };
 
-    impl_platform_host!(WebAudio webaudio "WebAudio");
+    #[cfg(feature = "web_audio_worklet")]
```
roderickvd, line 670: Could you clarify this? Instead of doing `impl_platform_host!`, importing the relevant `webaudio` bits yourself when the `web_audio_worklet` feature is enabled?

roderickvd (changes requested):
This looks very, very nice. Super to see such an exhaustive example too.

spookyvision:
> ## Pull Request Overview
LLMs can't create accurate summaries, only plausible looking ones.

> ### Reviewed Changes
LLMs can't review changes either.

👎

roderickvd:
@spookyvision What is the purpose of your comment? @kettle11 processed Copilot’s review comments as far as they were relevant. I do not understand what you are trying to contribute.

.gitignore
kettle11: Honestly I'm not sure. Weird!

src/host/web_audio_worklet/mod.rs
kettle11: Good call out. I've updated it so that it correctly checks if the Host is available first. I also realized that there were cases where AudioWorklets wouldn't be available that weren't being checked for, so I added those checks as well.

src/host/web_audio_worklet/mod.rs
kettle11: For me it passed clippy, but I just updated it anyways.

src/platform/mod.rs
kettle11: So the imports here were clearly wrong but for some reason it's puzzling me as to why it still behaved correctly. I spent 10-ish minutes looking at why it wasn't causing an issue, but I couldn't place it.

Either way I've fixed that line and tested that it's possible to select the different backends and it all seems to be working correctly.

examples/web-audio-worklet-beep/src/lib.rs
kettle11: The thing being unwrapped is an Option and the function error type is a JsValue, so I think it's simplest to just unwrap here.

It matches the other example and the unwrap would only fail if they've modified the associated index.html in which case the unwrap error would be informative.

examples/web-audio-worklet-beep/.cargo/config.toml
kettle11: I think right now when using multiple threads and atomics panic has to be set to abort. I can't remember the exact reason, but if I try to disable the flag there are cryptic build errors. Searching online doesn't turn up anything quickly. I think it's an area that's under documented right now because few people are using multithreaded WebAssembly with Rust.

For now I think leaving this as-is makes the most sense.

kettle11:
OK, I made a series of fixes to respond to comments and caught a few other things in the process. Thanks for the review!

examples/web-audio-worklet-beep/.cargo/config.toml
roderickvd: OK, thanks for looking into it.

roderickvd:
Thanks for the update! In the meantime a few conflicts came up because the platform implementation was changed (but simplified 😄 ) so if you could resolve those?

Then I’ll be out in the coming days so hope to have more time to look into it again over the weekend.

kettle11:
Apologies for the delay. Just got back to this to fix the merge conflicts. Everything should be good to go now!

roderickvd:
Thanks - I wonder what happened because the CI is failing on a lot of platforms now. Could you take another look at that?

kettle11:
I realize the way I did the merge before created a lot of noise, so I've cleaned that up. In the process though I also noticed that this change from a week ago in WasmBindgen may have caused a regression in this PR: https://github.com/wasm-bindgen/wasm-bindgen/pull/4670

I'm investigating that and pinged them to ask about it so this shouldn't get merged until that's resolved.

roderickvd:
Thanks a lot for the follow up on this. Will await your feedback.

kettle11:
OK! Finally this PR should be good to go. 

The fix landed in `wasm-bindgen` a few days ago and I tested and it works correctly now.

examples/web-audio-worklet-beep/src/lib.rs
```diff
@@ -0,0 +1,114 @@
...
+use wasm_bindgen::prelude::*;
+use web_sys::console;
+
+// When the `wee_alloc` feature is enabled, this uses `wee_alloc` as the global
+// allocator.
+//
+// If you don't want to use `wee_alloc`, you can safely delete this.
+#[cfg(feature = "wee_alloc")]
+#[global_allocator]
+static ALLOC: wee_alloc::WeeAlloc = wee_alloc::WeeAlloc::INIT;
```
roderickvd: In another commit I had removed `wee_alloc` because it's no longer maintained and shows up as a vulnerability in security audits. If you would remove it?

examples/web-audio-worklet-beep/Cargo.toml
```diff
@@ -0,0 +1,41 @@
...
+# `gloo` is a utility crate which improves ergonomics over direct `web-sys` usage.
+gloo = "0.11.0"
+# The `wasm-bindgen` crate provides the bare minimum functionality needed
+# to interact with JavaScript.
+wasm-bindgen = "0.2.45"
+
+# `wee_alloc` is a tiny allocator for wasm that is only ~1K in code size
+# compared to the default allocator's ~10K. However, it is slower than the default
+# allocator, so it's not enabled by default.
+wee_alloc = { version = "0.4.2", optional = true }
```
roderickvd: Let's remove it.

Cargo.toml
```diff
@@ -15,6 +15,17 @@ asio = [
     "num-traits",
 ] # Only available on Windows. See README for setup instructions.
 
+# Only available on web when atomics are enabled. See README for what it does.
+web_audio_worklet = [
+    "wasm-bindgen-futures",
```
roderickvd: Let's move to `dep:` syntax.

examples/web-audio-worklet-beep/src/lib.rs
kettle11: Good catch, I don't think wee_alloc even really ever worked properly. I believe it was just a bump allocator and didn't clearly advertise that, but I may be misremembering.

kettle11:
I made the changes.

One thing I realized while looking again at the code is I think I can improve upon the deinterleaving logic that I had implemented. It will work fine but there's probably some room for performance improvements and I'd like to do some experiments and measure it.

That said I think it's OK to make that a future PR, because any changes would be non-breaking and this version should work totally fine as-is.

roderickvd:
Merged, thanks!

kettle11:
Great! Thanks for the iterations on this one.

PR #961, opened 2025-03-12
Migrate from "oboe" to "ndk::audio"

### LINKED ISSUE #793: Remove / add feature for NotSendSyncAcrossAllPlatforms
If NotSendSyncAcrossAllPlatforms is just there in case of some eventual future Android compatibility, why not lock it behind a feature flag? 
Not being able to use streams across threads is a **major** drawback compared to the compatibility with the not yet supported Android AAudio API.

https://github.com/RustAudio/cpal/blob/f063271c232a4447169376e0c2be0c8a50649985/src/platform/mod.rs#L66-L72

### LINKED ISSUE #815: [Android/Arm64] Can't compile cpal
Hi, I cannot compile Cpal on android (with termux) because of an error related to the JNI package

Screenshot_2023-12-19-13-36-53-081_com termux

### LINKED ISSUE #818: Implement `Send` for `Stream`

### LINKED ISSUE #900: Android `failed to load dynamic library: dlopen failed: library "libc++shared.so" not found`
Hi all, 
I'm trying to run `cpal` on an Android device inside my Flutter application. When I run the app, I am greeted with this error:

```
Failed to load dynamic library 'librust_lib_flutter_rust_test.so': 
dlopen failed: library "libc++_shared.so" not found: 
needed by /data/app/~~zyZDDuG_hSM2uJQAOeqDBQ==/com.example.flutter_rust_test-sBfBRTor8WZZGQJgZMG0PA==/lib/arm64/librust_lib_flutter_rust_test.so in namespace clns-4
```
The project builds fine, but I run this code and then the error is thrown:
```rust

  let input_device = host.default_input_device().unwrap();
  let output_device = host.default_output_device().unwrap();

  let input_config = cpal::StreamConfig {
      channels: 1,
      sample_rate: cpal::SampleRate(48000),
      buffer_size: cpal::BufferSize::Default,
  };
  // let output_config: cpal::StreamConfig = input_config.clone();
  let output_config = cpal::StreamConfig {
      channels: 1,
      sample_rate: cpal::SampleRate(48000),
      buffer_size: cpal::BufferSize::Default,
  };

  let input_stream = input_device
      .build_input_stream(
          &input_config,
          move |data: &[f32], _: &_| {
              // do something
          },
          move |err| {
              eprintln!("There was an input error: {:?}", err);
          },
          None,
      )
      .unwrap();
 ```
 
I know that this is the same problem as #720 - and I tried this solution from that issue:
> I'm using this code to initialize the context. Just make it part of your fancy lib.
> 
> ```rust
> use jni::{JNIEnv, JavaVM};
> use std::ffi::c_void;
> 
> #[no_mangle]
> pub extern "C" fn JNI_OnLoad(vm: jni::JavaVM, res: *mut std::os::raw::c_void) -> jni::sys::jint {
>     let env = vm.get_env().unwrap();
>     let vm = vm.get_java_vm_pointer() as *mut c_void;
>     unsafe {
>         ndk_context::initialize_android_context(vm, res);
>     }
>     jni::JNIVersion::V6.into()
> }
> ```

no success here. I'm also not entirely sure where exactly to put this code. I tried putting it inside my `build.rs` and also with my main application code as well with no results.  

I also tried taking issue #563 's advice here:

> This can be solved by enabling the shared-stdcxx feature for oboe
> ```
> oboe = { version = "0.4", features = [ "java-interface", "shared-stdcxx" ] }
> ```
> 
> Then I also had to copy libc++_shared.so into my jniLibs directory (there's likely a way to automate this...)
> 
> _Originally posted by @trobanga in https://github.com/RustAudio/cpal/issues/563#issuecomment-1271637282_

Doing this created some strange errors. I see that it has usage of `IPHONEOS_DEPLOYMENT_TARGET` which wouldn't make sense for compiling to an android application:
``` cargo:warning=In file included from oboe/src/common/Trace.cpp:20:
  cargo:warning=oboe/src/common/OboeDebug.h:21:10: fatal error: 'android/log.h' file not found
  cargo:warning=#include <android/log.h>
  cargo:warning=         ^~~~~~~~~~~~~~~
  cargo:warning=1 error generated.
  cargo:warning=In file included from oboe/src/common/StabilizedCallback.cpp:18:
  cargo:warning=oboe/src/common/AudioClock.h:50:45: error: use of undeclared identifier 'TIMER_ABSTIME'
  cargo:warning=        return 0 - clock_nanosleep(clockId, TIMER_ABSTIME, &time, NULL);
  cargo:warning=                                            ^
  cargo:warning=oboe/src/common/AudioClock.h:67:24: error: use of undeclared identifier 'clock_nanosleep'
  cargo:warning=            return 0 - clock_nanosleep(clockId, 0, &time, NULL);
  cargo:warning=                       ^
  cargo:warning=2 errors generated.
  exit status: 1
  cargo:warning=ToolExecError: Command env -u IPHONEOS_DEPLOYMENT_TARGET "c++" "-O0" "-ffunction-sections" "-fdata-sections" "-fPIC" "-gdwarf-2" "-fno-omit-frame-pointer" "-arch" "arm64" "-static" "-I" "oboe/include" "-I" "oboe/src" "-I" "oboe-ext/include" "-I" "oboe-ext/src" "-Wall" "-Wextra" "-std=c++17" "-Wall" "-Wextra-semi" "-Wshadow" "-Wshadow-field" "-fno-rtti" "-fno-exceptions" "-DOBOE_ENABLE_LOGGING=1" "-o" "/Users/zacharyhaslam/FlutterApplications/flutter_rust_test/rust/target/debug/build/oboe-sys-54157a8a77fee849/out/library/oboe/src/common/StabilizedCallback.o" "-c" "oboe/src/common/StabilizedCallback.cpp" with args "c++" did not execute successfully (status code exit status: 1).
  exit status: 1
  cargo:warning=ToolExecError: Command env -u IPHONEOS_DEPLOYMENT_TARGET "c++" "-O0" "-ffunction-sections" "-fdata-sections" "-fPIC" "-gdwarf-2" "-fno-omit-frame-pointer" "-arch" "arm64" "-static" "-I" "oboe/include" "-I" "oboe/src" "-I" "oboe-ext/include" "-I" "oboe-ext/src" "-Wall" "-Wextra" "-std=c++17" "-Wall" "-Wextra-semi" "-Wshadow" "-Wshadow-field" "-fno-rtti" "-fno-exceptions" "-DOBOE_ENABLE_LOGGING=1" "-o" "/Users/zacharyhaslam/FlutterApplications/flutter_rust_test/rust/target/debug/build/oboe-sys-54157a8a77fee849/out/library/oboe/src/common/Trace.o" "-c" "oboe/src/common/Trace.cpp" with args "c++" did not execute successfully (status code exit status: 1).cargo:warning=oboe-ext/src/AudioStreamCallbackWrapper.cpp:3:10: fatal error: 'android/log.h' file not found
  cargo:warning=#include <android/log.h>
  cargo:warning=         ^~~~~~~~~~~~~~~
  cargo:warning=1 error generated.
  exit status: 0
  exit status: 0
  exit status: 0
  exit status: 1
  cargo:warning=ToolExecError: Command env -u IPHONEOS_DEPLOYMENT_TARGET "c++" "-O0" "-ffunction-sections" "-fdata-sections" "-fPIC" "-gdwarf-2" "-fno-omit-frame-pointer" "-arch" "arm64" "-static" "-I" "oboe/include" "-I" "oboe/src" "-I" "oboe-ext/include" "-I" "oboe-ext/src" "-Wall" "-Wextra" "-std=c++17" "-Wall" "-Wextra-semi" "-Wshadow" "-Wshadow-field" "-fno-rtti" "-fno-exceptions" "-DOBOE_ENABLE_LOGGING=1" "-o" "/Users/zacharyhaslam/FlutterApplications/flutter_rust_test/rust/target/debug/build/oboe-sys-54157a8a77fee849/out/library/oboe-ext/src/AudioStreamCallbackWrapper.o" "-c" "oboe-ext/src/AudioStreamCallbackWrapper.cpp" with args "c++" did not execute successfully (status code exit status: 1).
  --- stderr
  error occurred: Command env -u IPHONEOS_DEPLOYMENT_TARGET "c++" "-O0" "-ffunction-sections" "-fdata-sections" "-fPIC" "-gdwarf-2" "-fno-omit-frame-pointer" "-arch" "arm64" "-static" "-I" "oboe/include" "-I" "oboe/src" "-I" "oboe-ext/include" "-I" "oboe-ext/src" "-Wall" "-Wextra" "-std=c++17" "-Wall" "-Wextra-semi" "-Wshadow" "-Wshadow-field" "-fno-rtti" "-fno-exceptions" "-DOBOE_ENABLE_LOGGING=1" "-o" "/Users/zacharyhaslam/FlutterApplications/flutter_rust_test/rust/target/debug/build/oboe-sys-54157a8a77fee849/out/library/oboe-ext/src/AudioStreamCallbackWrapper.o" "-c" "oboe-ext/src/AudioStreamCallbackWrapper.cpp" with args "c++" did not execute successfully (status code exit status: 1).
 ```
 
However, I'm using `flutter_rust_bridge` which uses a code gen tool that handles most of the compilation for me. I think it automatically compiles for iOS and Android at the same time. I'm not super good at diagnosing configuration errors on Android, so I'm not entirely sure where I could be going wrong here. I also posted [this](https://github.com/fzyzcjy/flutter_rust_bridge/issues/2209) issue on the flutter_rust_bridge github which might provide some more clarity. 

I've seen people mention dragging a `libc++_shared.so` file into a `jniLibs` directory, but neither are present in my project files. I also tried making my own empty `jniLibs` folder to no success either. 

Long story short, no matter what I seem to try, I can't seem to get rid of this error. Any ideas here?

### PR DESCRIPTION
**Pull request by ColorBuffer**
This pull request drops `Oboe` and uses `ndk::audio` (thin safe binding of native AAudio) for android platforms. 

## Pros:
✅ Pure rust implementation, so `libc++_shared.so` is not required anymore. Fixes #900, Fixes #815
✅ **Stream** now could be **Send** and move between threads safely. Fixes #793, Fixes #818
✅ **Stream** can accept wider range of `SampleFormat { I16, F32, I32, ... }`. (in the future, See: [#497](https://github.com/rust-mobile/ndk/pull/497) )
✅ Supported since `ANDROID_API >= 26` witch is **%95** of devices as [android distibution dashboard](https://apilevels.com/) reports.

## Cons:
❌ not supported in `ANDROID_API < 26` witch is **%5** of devices world wide :(

### CODE REVIEW

**Comment by ColorBuffer** [comment]
I need help to configure cargo-apk test to use **android26** toolchain not **android23**

**Comment by est31** [comment]
> I need help to configure cargo-apk test to use android26 toolchain not android23

strange, I've thought we use android 30 in CI since commit 802073d808f6eeb5f300816a97ae3ed83e420372.

**Comment by est31** [comment]
Thanks for filing this! I wonder, could you rename the whole `oboe` directory to `android` maybe? or `android-ndk`?

**Comment by ColorBuffer** [comment]
@est31 what about **aaudio**? it looks nicer and keeps harmony with **CoreAudio**

**Comment by est31** [comment]
```
note: ld.lld: error: unable to find library -laaudio
```

hmm I wonder how one can fix this error.

ColorBuffer:
@est31 
this is very similar to [Linker error in Android Studio while compiling NDK project](https://stackoverflow.com/questions/45316227/linker-error-in-android-studio-while-compiling-ndk-project) witch got fixed by changing **min_sdk_version** to 26 (as expected). but here we have cargo-apk that seems not reading **cargo.toml/min_sdk_version**

ColorBuffer:
@est31 moving android example to a dedicated crate with some apk metadata solved that problm

est31 (approved):
thank you :heart:

Frando:
> ❌ not supported in ANDROID_API < 26 witch is %5 of devices world wide :(

This change has quite big consequences, therefore I'd like to dicuss this: We are currently adding audio functionality to an open-source Android app with several hundred thousands active users, and are seriously consider using cpal, as the app is cross-platform and uses a shared Rust core between all platforms (Linux, Windows, macOS, iOS, Android). The app currently works on all versions from Android 5.0 Lollipop, and has a significant amount of users on old, outdated Android devices. AAudio was only added in Android 8.0 Oreo. This is a significant jump, and it is currently most likely not feasable in the short term for this app to do this change, because it would force several thousand users to either buy new devices or stop using the app. Therefore, this change would mean we'd either have to not use cpal or fork cpal to still include the `oboe` API.

Is there a significant downside to leaving the current `oboe` backend intact, and adding the `AAudio`-only backend as a seperate host, with feature flags to select between the two? The easier-to-build `AAudio` could be enabled by default, whereas `oboe` could be behind a feature flag.

est31:
Thanks for this feedback. I understand that if there is 3.2 billion Android devices, then 5% of that would be 115 million. For apps at your scale, it would still be thousands of devices.

I suppose we could continue maintaining the `oboe` backend upstream behind a feature flag, which would help with containing the issues of that backend to the app devs who really need it. A question would be however for how long.

@Frando would you be interested in becoming the maintainer of the `oboe` backend?

Issue #773, opened 2023-04-04
Can you get the system latency?


tGrothmannFluffy:
Hi there,

I am trying to get the system latency. Basically, I want to know (_roughly_) when the samples I wrote to the output are ending up at the speaker.
I know this is probably highly os and device dependent.

I found [this oboe (android) function](https://docs.rs/oboe-sys/0.4.5/oboe_sys/fn.oboe_AudioStream_calculateLatencyMillis.html) to get this value. But I was wondering if/how I could access that.

I am working on Android and iOS.

morajabi:
I'm interested in this for macOS. Use case is for using acoustic echo cancellation works better with exact latency metrics.

j-n-f:
Also interested in this.

From what I can tell the StreamInstant values in callbacks is just giving `block_size / sample_rate`. The value doesn't match what my driver reports (it's close, but less than the actual values).

From a quick look at ASIO and ALSA they both return this information in frame/sample counts. It would be useful to know what this looks like for the other backends.

It should be possible to use that data to provide more accurate `StreamInstant` values. Since at least ASIO + ALSA provide the latency in frames/samples, it might be nice to include that value as part of the callback info structs.

j-n-f:
Here's what I was able to scrounge up.

My previous comment on putting this latency information at the stream config level may not always be possible. If Oboe can even provide this value, it sounds like it might be dynamically changing block size while a stream is still running.

If some backend can't provide more accurate latencies then the existing heuristic was the best you could do anyways.

This should be enough to start experimenting.

|backend|api|format|comment|
|-|-|-|-|
|ALSA|[snd_pcm_delay](https://www.alsa-project.org/alsa-doc/alsa-lib/group___p_c_m.html#ga012e8b999070e72ab23514f25e7d6482)|frames/samples||
|JACK|???|???|not sure, guessing you just use ALSA to fetch this information|
|WASAPI|[IAudioClient::GetStreamLatency](https://learn.microsoft.com/en-us/windows/win32/api/audioclient/nf-audioclient-iaudioclient-getstreamlatency)|100ns increments||
|ASIO|ASIOGetLatencies (ASIO SDK 2.3, page 14)|samples/frames|must call after ASIOCreateBuffers <br /><br />Driver may notify host (cpal) of `kAsioLatenciesChanged` (ASIO SDK 2.3, page 9)|
[...]

PR #964, opened 2025-03-24
Add method to query latency from `Stream`

marcpabst:
This is initial work to enable getting the current latency in samples from a Stream. The `latency() -> `Option<u32>` method returns `None` by default, so this will already compile with all backends.

Addreses  #773.

### DIFF
--- a/src/host/coreaudio/macos/mod.rs
+++ b/src/host/coreaudio/macos/mod.rs
@@ -33,6 +33,7 @@ use std::fmt;
 use std::mem;
 use std::os::raw::c_char;
 use std::ptr::null;
+use std::rc::Rc;
 use std::slice;
 use std::sync::mpsc::{channel, RecvTimeoutError};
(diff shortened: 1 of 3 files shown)


marcpabst:
I think this can be reviewed now - so far it only implements `latency()` for MacOS Coreaudio.

marcpabst:
Could I get some feedback on this? I'm considering pouring some work into `cpal` (including improvments to th WASAPI backend) but I don't really want to maintain by own fork...

est31 (approved):

roderickvd:
@marcpabst I got added as maintainer and am doing some backlog grooming. Yes there is interest in this. I'd like to ensure that a sensical number of hosts would be supported coming out of the gates. So beyond CoreAudio also Alsa and WASAPI?

roderickvd:
Superseded by #1129 - thanks for bringing this to light.

PR #971, opened 2025-04-25
Fix clippy lints.

chloebrett:
Only fixes for alsa and shared components.

The only controversial one here might be removing the commented-out 24- and 48-bit ints. If anyone has context on that, please let me know.

### DIFF
--- a/src/samples_formats.rs
+++ b/src/samples_formats.rs
@@ -46,8 +46,14 @@ pub enum SampleFormat {
     /// `u16` with a valid range of `u16::MIN..=u16::MAX` with `1 << 15 == 32768` being the origin.
     U16,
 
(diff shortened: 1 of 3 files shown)


src/samples_formats.rs
```diff
@@ -46,15 +46,9 @@ pub enum SampleFormat {
     /// `u16` with a valid range of `u16::MIN..=u16::MAX` with `1 << 15 == 32768` being the origin.
     U16,
 
-    /// `U24` with a valid range of '0..16777216' with `1 << 23 == 8388608` being the origin
```
est31: can you convert these to normal comments instead of doc comments?

est31:
There is also a CI failure, see the links.

src/samples_formats.rs
chloebrett: Done.

PR #974, opened 2025-05-03
Support hidden devices on MacOS

matusfaro:
Allow constructing a CoreAudioDevice (Device on MacOS) given just its ID. This is useful for accessing hidden devices. A hidden device is a device that is not visible to user in system settings, but may be used by applications internally. 

[Example usage](https://github.com/ollisten/ollisten/blob/5c1fc28ef70a15b8a18d457b7cb2f34be91d8a63/src-tauri/src/transcription/cpal_macos_hack.rs#L10-L15):
```rust
pub fn create_cpal_mic_from_id(audio_device_id: AudioDeviceID) -> CoreAudioDevice {
    CoreAudioDevice::new(audio_device_id)
}
```

### DIFF
--- a/.github/workflows/cpal.yml
+++ b/.github/workflows/cpal.yml
@@ -14,6 +14,8 @@ jobs:
       run: sudo apt-get install libasound2-dev
     - name: Install libjack
       run: sudo apt-get install libjack-jackd2-dev libjack-jackd2-0
+    - name: Install dbus
+      run: sudo apt-get install libdbus-1-dev
     - name: Install stable
       uses: dtolnay/rust-toolchain@stable
       with:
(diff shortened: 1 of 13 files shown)


est31:
If you give it an invalid ID, will it crash? or will it return sensible errors?

matusfaro:
@est31 It constructs without error, then panics when you attempt to use it. I don't know how to make that safer except maybe via stronger documentation.

Just to get the right device UID, [I have to call low level C API](https://github.com/ollisten/ollisten/blob/7db472ecb5cc0e0d94a750bf112a6feac700319d/src-tauri/src/audio/macos_core_audio.rs#L286-L325). I guess we could bake in these C APIs to get the device id or verify its existence but that would be a bigger undertaking.

est31:
A panic is finw, just shouldn't segfault or such. Merging

est31 (approved):

PR #978, opened 2025-06-04
Bump to v0.16.0

madsmtm:
I've tried to assemble the changelog by looking at the [full commit range](https://github.com/RustAudio/cpal/compare/7b21251...ed9d643) since v0.15.3, but I might have made a mistake once or twice.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,28 @@
 # Unreleased
 
-- All error enums are now `Clone`.
+# Version 0.16.0 (2024-06-06)
+
+- Migrate from `oboe` to `ndk::audio`. **NOTE:** This raises the minimum Android API version to 26 (Android 8/Oreo).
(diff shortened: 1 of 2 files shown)


est31 (changes requested):
I think we should call it 0.16.0 due to the oboe bump. not everyone was happy about that.

madsmtm:
Fair, I've done that now

est31:
merged it, although we should probably correct the publication date, as it's June 7 already (on some parts of the globe already June 8).

madsmtm:
Filed https://github.com/RustAudio/cpal/pull/979.

Would you mind doing the actual release as well?

est31:
@madsmtm it is published to crates.io now. Apparently something broke in the publish job. https://github.com/RustAudio/cpal/actions/runs/15510765618/job/43671429557

PR #979, opened 2025-06-07
Fix release date for v0.16

madsmtm:
As mentioned in https://github.com/RustAudio/cpal/pull/978#issuecomment-2952782267


edwloef:
Shouldn't that be **2025**-06-07?

madsmtm:
🤦

Filed https://github.com/RustAudio/cpal/pull/980, thanks!

PR #990, opened 2025-07-23
use set_buffer_size_near to calculate fixed alsa buffer size

### LINKED ISSUE #743: ALSA function 'snd_pcm_hw_params_set_buffer_size' failed when setting buffer size
Hi! I'm having an issue as soon as I try specifying a buffer size for my output stream.

The following code runs:

```rust
use cpal::{BufferSize, StreamConfig};

fn main() {

    let mut supported_configs_range = device.supported_output_configs().unwrap();
    let config = supported_configs_range
        .next()
        .unwrap()
        .with_max_sample_rate();

    dbg!(&config);
    let mut c: StreamConfig = config.into();
    // c.buffer_size = BufferSize::Fixed(64);

    let stream = device
        .build_output_stream(
            &c,
            move |_: &mut [f32], _: &cpal::OutputCallbackInfo| {},
            move |_| {},
        )
        .unwrap();

    stream.play().unwrap();
}
```

Furthermore, the `SupportedStreamConfig` I get is as follows:

```
[src/main.rs:15] &config = SupportedStreamConfig {
    channels: 1,
    sample_rate: SampleRate(
        384000,
    ),
    buffer_size: Range {
        min: 3,
        max: 4194304,
    },
    sample_format: I16,
}
```

The above leads me to believe that I can specify any `buffer_size` between 3 and 4194304. Is this correct? Unfortunately, selecting any `Fixed` buffer size leads to failure. For example, if I uncomment the line `c.buffer_size = BufferSize::Fixed(64)`, I get:

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: BackendSpecific { err: BackendSpecificError { description: "ALSA function 'snd_pcm_hw_params_set_buffer_size' failed with error 'EINVAL: Invalid argument'" } }', src/main.rs:25:10
```

Am I doing something wrong or is this a bug? I cant seem to specify any buffer size other than `Default`.

I tried using other values, such as the minimum 3, and the maximum 4194304, with the same results. I also adapted the code to test it with the `master` version of `cpal`, and I got the same error. 

I'm very new to the audio processing world so apologies if I omitted anything. Any guidance would be appreciated. Thanks!

attackgoat:
Fixes #743

Hardware parameters cannot be set to some arbitrary values within the supported range.

Two alternate fixes are:
- Expose the "near" ALSA function to the user
- Return the actual buffer size used

I am not familiar enough with CPAL usage to know if the alternate fixes are needed, but for my use cases they are not.

### DIFF
--- a/src/host/alsa/mod.rs
+++ b/src/host/alsa/mod.rs
@@ -1107,7 +1107,7 @@ fn set_hw_params_from_format(
     match config.buffer_size {
         BufferSize::Fixed(v) => {
             hw_params.set_period_size_near((v / 4) as alsa::pcm::Frames, alsa::ValueOr::Nearest)?;
-            hw_params.set_buffer_size(v as alsa::pcm::Frames)?;
+            hw_params.set_buffer_size_near(v as alsa::pcm::Frames)?;
         }
         BufferSize::Default => {
             // These values together represent a moderate latency and wakeup interval.


roderickvd:
Agreed, this is the right thing to do. This highly related commit in `librespot` actually combines the two approaches you mention: https://github.com/librespot-org/librespot/commit/d5efb8a620554ced4fda731c72046c4b1af53111

roderickvd:
@attackgoat as I became a maintainer today, and agree with your approach, I would like to work with you to getting to a mergable state. Could you update it to:
- take the device default when the "near" function fails? (look at the librespot example for inspiration)
- add a changelog entry

Thanks!

roderickvd:
I see that #917 also does this (and more).

attackgoat:
> I see that #917 also does this (and more).

That PR seems to be the better direction, although it causes me the same issue with large buffers. For now I've addressed the feedback but we may want to close this PR if #917 is cleaned up.

roderickvd:
Thanks a lot @attackgoat. The author of #917 expressed that he is no longer working on it. Would you be willing to merge the two approaches?

src/host/alsa/mod.rs
```diff
@@ -1104,19 +1104,27 @@ fn set_hw_params_from_format(
...
+                .is_err()
+        {
+            buffer_size = BufferSize::Default;
         }
     }
 
+    if let BufferSize::Default = buffer_size {
+        // These values together represent a moderate latency and wakeup interval.
+        // Without them, we are at the mercy of the device
+        hw_params.set_period_time_near(25_000, alsa::ValueOr::Nearest)?;
```
roderickvd: I'm questioning whether an error here should immediately generate a function result. As you've found, there may be reasons why a device returns a different period size and we probably want to use the *actual* period size for the buffer time.

As well - though this is opinionated - I'd probably rather have initialization succeed (so we can play sound) rather than fail hard just because the exact period or buffer time can't be set.

attackgoat: The latest commit handles various failures better and relies on the actual ALSA-provided period _size_ after setting _time_.

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Unreleased
 
 - ALSA(process_output): Pass `silent=true` to `PCM.try_recover`, so it doesn't write to stderr.
+- ALSA: Fix buffer size selection. (error 22)
```
roderickvd: This PR is going to be capable of doing a lot more :wink:
It would be good to concisely but clearly capture the changes to the period size, buffer size, and fallback handling.

src/host/alsa/mod.rs
```diff
@@ -1104,22 +1104,74 @@ fn set_hw_params_from_format(
...
+        period_size
+    } else {
+        return false;
+    };
+
+    // We should not fail if the driver is unhappy here.
+    // `default` pcm sometimes fails here, but there no reason to as we attempt to provide a size or
+    // minimum number of periods.
+    hw_params
+        .set_buffer_size(period_size * PERIOD_COUNT as alsa::pcm::Frames)
```
roderickvd: This fallback is stricter than the primary path: should bed we not also use `set_buffer_size_near` here?

src/host/alsa/mod.rs
```diff
@@ -1104,22 +1104,74 @@ fn set_hw_params_from_format(
...
+
+    Ok(hw_params.can_pause())
+}
+
+/// Returns true if the periods were reasonably set. A false result indicates the device default
+/// configuration is being used.
+fn set_hw_params_periods(hw_params: &alsa::pcm::HwParams, buffer_size: BufferSize) -> bool {
+    const FALLBACK_PERIOD_TIME: u32 = 25_000;
+
+    // When the API is made available, this could rely on snd_pcm_hw_params_get_periods_min and
```
roderickvd: Indeed, let's make a "TODO" as a future reminder.

src/host/alsa/mod.rs
```diff
@@ -1104,22 +1104,74 @@ fn set_hw_params_from_format(
...
+    // snd_pcm_hw_params_get_periods_max
+    const PERIOD_COUNT: u32 = 2;
+
+    /// Returns true if the buffer size was reasonably set.
+    fn set_hw_params_buffer_size(hw_params: &alsa::pcm::HwParams, buffer_size: FrameCount) -> bool {
```
roderickvd: We can make buffer size setting more robust by inspecting the device capabilities and clamping to the supported buffer sizes. Something like:

```rust
fn clamp_frames_to_device_caps(hw_params: &alsa::pcm::HwParams, requested: FrameCount) -> FrameCount {
    let min_buf = hw_params.get_buffer_size_min().unwrap_or(1);
    let max_buf = hw_params.get_buffer_size_max().unwrap_or(Frames::MAX); // usize::MAX? or something more sensical?
    requested.clamp(min_buf, max_buf)
}
```

src/host/alsa/mod.rs
```diff
@@ -1104,22 +1104,74 @@ fn set_hw_params_from_format(
...
+        // Desired period size
+        let period_size = (buffer_size / PERIOD_COUNT) as alsa::pcm::Frames;
+
+        if hw_params
+            .set_period_size_near(period_size, alsa::ValueOr::Greater)
+            .is_err()
+        {
+            return false;
```
roderickvd, line 1201: If setting the period size fails, we still could continue getting the actual period size and tuning the buffer size to it.

src/host/alsa/mod.rs
```diff
@@ -1104,22 +1104,74 @@ fn set_hw_params_from_format(
...
+    const FALLBACK_PERIOD_TIME: u32 = 25_000;
+
+    // When the API is made available, this could rely on snd_pcm_hw_params_get_periods_min and
+    // snd_pcm_hw_params_get_periods_max
+    const PERIOD_COUNT: u32 = 2;
+
+    /// Returns true if the buffer size was reasonably set.
+    fn set_hw_params_buffer_size(hw_params: &alsa::pcm::HwParams, buffer_size: FrameCount) -> bool {
+        // Desired period size
+        let period_size = (buffer_size / PERIOD_COUNT) as alsa::pcm::Frames;
```
roderickvd: Period size we could similarly clamp by inspecting device capabilities.

I know it sounds like `_near` and `ValueOr` would yield something usable in all cases, but from librespot experience I know that not all ALSA drivers do.

roderickvd (changes requested):
Thanks a lot, this is really going in the right direction. Some points to make it even more robust - I've got prior experience where I know that ALSA drivers can be really fickle.

At any time just let me know if and how I can help.

src/host/alsa/mod.rs
attackgoat: I've added this and also updated these other bits to use the same logic:
- Supported stream configuration min/max clamped to valid range
- Buffer size set after using `BufferSize::Default` checked for valid range

src/host/alsa/mod.rs
attackgoat: If this path fails, the fallback will use period time to calculate a valid buffer size. The last commit (4da16f1) has moved this all around a bit.

src/host/alsa/mod.rs
```diff
@@ -1104,22 +1135,95 @@ fn set_hw_params_from_format(
...
+        };
+
+        // Desired period size
+        let period_size = {
+            let (min_period_size, max_period_size) = hw_params_period_size_min_max(hw_params);
+            (buffer_size / PERIOD_COUNT).clamp(min_period_size, max_period_size)
+        };
+
+        // Actual period size
+        let Ok(period_size) =
```
roderickvd, line 1198: As final polish this would be even more resilient:

```rust
// Try to set desired period size
let period_size = if let Ok(size) = hw_params.set_period_size_near(period_size as _, alsa::ValueOr::Greater) {
    size  // Use the size we successfully set
} else {
    // If setting fails, try to get whatever period size the device currently has
    if let Ok(current_size) = hw_params.get_period_size() {
        current_size  // Work with device's current period size
    } else {
        return false;  // Only fail if we can't even query the current size
    }
};
```

roderickvd (changes requested):
Well implemented with one final bit of polish! 👏

src/host/alsa/mod.rs
roderickvd: Actually, though I marked the review as “changes requested”, keen to know what you think. It has the upside of pretty much always working, but the downside of potentially ignoring the user’s request.

attackgoat: This was been incorporated in [b710e04](https://github.com/RustAudio/cpal/pull/990/commits/b710e0464a5731d34b3cd917442da0c7e09544cc), but after rethinking it I don't think this check is required.

Reasoning: The sub-function `set_hw_params_buffer_size` takes a buffer size and attempts to fit the desired number of periods. If this fails, the fallback in the outer function already gets the period size and attempts to set valid parameters using a period-time approach, thereby handling the proposed check.

Also:
- I realized `snd_pcm_hw_params_set_periods` was only being called in the absolute fallback case, so the desired count of two was never actually respected (fixed)
- It's possible to make `set_hw_params_periods` infallible: if it fails it is already ignored anyways (not fixed)

roderickvd: I've been thinking about this as well - thanks for sticking through this difficult undertaking of getting the period / buffer system right.

We need to have a behavior contract that's clear to what a user would reasonable expect:
- `BufferSize::Fixed(n)` → "I need around `n` frames, fail if you can't get close"
- `BufferSize::Default` → "Give me something reasonable for this hardware"

Right now we might be _too_ accommodating for `Fixed(n)` and succeed even when we request 128 frames and fall back to 2048. By analogy, we also fail opening a 48 kHz stream when a device only supports 44.1 or 96 kHz.

Should be easy to change, right?

```rust
  if let BufferSize::Fixed(val) = buffer_size {
      if set_hw_params_buffer_size(hw_params, period_count, val) {
          return true;
      }
      // Currently falls through to fallback
  }
```

⬇️ 

```rust
  if let BufferSize::Fixed(val) = buffer_size {
      if set_hw_params_buffer_size(hw_params, period_count, val) {
          return true;
      }
      return false; // Fail immediately - don't use fallback
  }
```

attackgoat: Sounds good - the code has been adjusted to return `BackendSpecificError` in case it cannot set the requested buffer size (default or fixed).

In the future I think it would be helpful to return a better error type that programs could inspect and use to make decisions, because the string-based error is only helpful for the programmer. The `log` crate would be handy here.

roderickvd:
Great. Going to merge this now. If you've got further improvements, we can add them later. This already is a big improvement, thanks so much.

PR #993, opened 2025-07-30
Fix warning in MacOS CoreAudio — `audio_device_id < 0`

wgibbs-rs:
After running `cargo build` and `cargo clippy` I observed a warning for code testing if a u32 variable was less than 0. 

```
pub fn new(audio_device_id: AudioDeviceID) -> Self {
        Device {
            audio_device_id,
            is_default: audio_device_id < 0;
        }
    }
```

This doesn't make sense, however, as a u32 cannot be below 0. For readability-sake, and because I didn't have anything else to do, I've changed this from `audio_device_id < 0` to `false`.

### DIFF
--- a/src/host/coreaudio/macos/mod.rs
+++ b/src/host/coreaudio/macos/mod.rs
@@ -166,7 +166,11 @@ impl Device {
     pub fn new(audio_device_id: AudioDeviceID) -> Self {
         Device {
             audio_device_id,
-            is_default: audio_device_id < 0,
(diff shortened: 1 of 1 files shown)


roderickvd (changes requested):
Thanks for the clippy fix. For clarity, let's remove the first three lines of code comments, and just keep the TODO.

wgibbs-rs:
Done! Let me know what else I can do to help.

roderickvd:
🚀 first one!

PR #995, opened 2025-07-31
feat(android): add `input_preset` to `StreamConfig`

TD-Sky:
Users can enable AEC (Acoustic Echo Canceler) via setting `VOICE_COMMUNICATION` for recording audio on Android (refer to [here](https://source.android.com/docs/core/audio/implement-pre-processing)), which is a powerful and important feature for developing Android applications. However, I find no ways to enable this feature in **cpal** at present. So I add it here.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,6 +11,7 @@ rust-version = "1.70"
 
 [features]
 asio = ["asio-sys", "num-traits"] # Only available on Windows. See README for setup instructions.
+android-input-preset = ["ndk/api-level-28"]
 
 # Deprecated, the `oboe` backend has been removed
 oboe-shared-stdcxx = []
@@ -83,7 +84,7 @@ js-sys = { version = "0.3.35" }
 web-sys = { version = "0.3.35", features = [ "AudioContext", "AudioContextOptions", "AudioBuffer", "AudioBufferSourceNode", "AudioNode",  "AudioDestinationNode", "Window", "AudioContextState"] }
 
 [target.'cfg(target_os = "android")'.dependencies]
-ndk = { version = "0.9", default-features = false, features = ["audio", "api-level-26"]}
+ndk = { version = "0.9", default-features = false, features = ["audio", "api-level-26"] }
 ndk-context = "0.1"
 jni = "0.21"
 num-derive = "0.4"
--- a/src/host/aaudio/mod.rs
+++ b/src/host/aaudio/mod.rs
@@ -1,4 +1,3 @@
-use std::cell::RefCell;
 use std::cmp;
 use std::convert::TryInto;
 use std::time::{Duration, Instant};
@@ -14,7 +13,7 @@ use crate::{
     BackendSpecificError, BufferSize, BuildStreamError, Data, DefaultStreamConfigError,
     DeviceNameError, DevicesError, InputCallbackInfo, InputStreamTimestamp, OutputCallbackInfo,
     OutputStreamTimestamp, PauseStreamError, PlayStreamError, SampleFormat, SampleRate,
-    SizedSample, StreamConfig, StreamError, SupportedBufferSize, SupportedStreamConfig,
+    StreamConfig, StreamError, SupportedBufferSize, SupportedStreamConfig,
     SupportedStreamConfigRange, SupportedStreamConfigsError,
 };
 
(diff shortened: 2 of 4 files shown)


wgibbs-rs:
Looking at the Github Actions results, it seems you left out some platform specific flags. Unfortunately, I can't edit your code, but it may be best to ensure this doesn't fail other systems. 

Take a look at this:

MacOS: `error[E0433]: failed to resolve: use of unresolved module or unlinked crate "ndk"`

Would it be possible to go back and prevent other platforms from attempting to use this code, or fix the cross-platform issues?

TD-Sky:
> Looking at the Github Actions results, it seems you left out some platform specific flags. Unfortunately, I can't edit your code, but it may be best to ensure this doesn't fail other systems.
> 
> Take a look at this:
> 
> MacOS: `error[E0433]: failed to resolve: use of unresolved module or unlinked crate "ndk"`
> 
> Would it be possible to go back and prevent other platforms from attempting to use this code, or fix the cross-platform issues?

All fixed.

Cargo.toml
```diff
@@ -11,6 +11,7 @@ rust-version = "1.70"
 
 [features]
 asio = ["asio-sys", "num-traits"] # Only available on Windows. See README for setup instructions.
+android-input-preset = ["ndk/api-level-28"]
```
roderickvd, line 14: I see this requires `api-level-28` where the normal feature requires `api-level-26`. That OK?

src/lib.rs
```diff
@@ -410,6 +412,8 @@ impl SupportedStreamConfig {
             channels: self.channels,
             sample_rate: self.sample_rate,
             buffer_size: BufferSize::Default,
+            #[cfg(all(target_os = "android", feature = "android-input-preset"))]
```
roderickvd, line 415: Would it make sense to rename the feature flag to something more specific like "android-voice-recognition" or "android-aec"? Because it seems like there are more input presets that one could choose, if not today then in the future.

roderickvd (changes requested):
Nice addition. If you would take a look at the review comments and add a changelog entry?

src/host/aaudio/mod.rs
```diff
@@ -14,7 +13,7 @@ use crate::{
     BackendSpecificError, BufferSize, BuildStreamError, Data, DefaultStreamConfigError,
     DeviceNameError, DevicesError, InputCallbackInfo, InputStreamTimestamp, OutputCallbackInfo,
     OutputStreamTimestamp, PauseStreamError, PlayStreamError, SampleFormat, SampleRate,
-    SizedSample, StreamConfig, StreamError, SupportedBufferSize, SupportedStreamConfig,
+    StreamConfig, StreamError, SupportedBufferSize, SupportedStreamConfig,
```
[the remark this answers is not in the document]
roderickvd: No, we appreciate opportunistic refactoring.

src/lib.rs
```diff
@@ -410,6 +412,8 @@ impl SupportedStreamConfig {
...
+            input_preset: platform::AudioInputPreset::VoiceRecognition,
```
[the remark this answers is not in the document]
roderickvd: No, Copilot is missing the feature gate.

Cargo.toml
TD-Sky: I don't see any conflicts between the features in `ndk` document and didn't find any errors about api levels when I used the branch of this PR in my own project.

src/lib.rs
TD-Sky: I think it better to keep the name because AEC is not the only function of the flag.

roderickvd: That's what I was trying to say. Looking at the [AudioInputPreset enum](https://docs.rs/ndk/latest/ndk/audio/enum.AudioInputPreset.html) there are six variants today. How can we make the feature flag clearly convey *which* preset gets enabled? 

Edit: what would be a good way going forward to support the other variants as well?

Cargo.toml
roderickvd: Well, if it works, it works. As it _does_ bump the minimum Android version, it would be a good thing to document. Would you document the feature flag, as well as add an entry to the changelog?

roderickvd:
This is a candidate for opt-in with #1010 together with the other audio input presets.

roderickvd:
To land through the extension API proposed in #1334

PR #996, opened 2025-07-31
properly detect `is_default` in `Device::new()` for MacOS

wgibbs-rs:
I've introduced proper detection if a `Device::new()` is actually default to the system.

Previously, I hardcoded `is_default` to be false, regardless of the device ID. Now, the function will gather both the default input and output device IDs, and compare them to this device. If this device matches at least one device's ID, and no platform specific error occurred behind the scenes, we return true, otherwise we return false.

If we'd prefer to now return `Result<Self, BackendSpecificError>` I can implement this, though I didn't as I didn't have time. Just let me know if you'd prefer that, now that these errors *can* occur.

### DIFF
--- a/src/host/coreaudio/macos/enumerate.rs
+++ b/src/host/coreaudio/macos/enumerate.rs
@@ -105,7 +105,7 @@ pub fn default_input_device() -> Option<Device> {
             NonNull::from(&mut audio_device_id).cast(),
         )
     };
-    if status != kAudioHardwareNoError as i32 {
+    if status != kAudioHardwareNoError {
         return None;
     }
 
@@ -135,7 +135,7 @@ pub fn default_output_device() -> Option<Device> {
             NonNull::from(&mut audio_device_id).cast(),
         )
     };
-    if status != kAudioHardwareNoError as i32 {
+    if status != kAudioHardwareNoError {
         return None;
     }
 
--- a/src/host/coreaudio/macos/enumerate.rs
+++ b/src/host/coreaudio/macos/enumerate.rs
@@ -105,7 +105,7 @@ pub fn default_input_device() -> Option<Device> {
             NonNull::from(&mut audio_device_id).cast(),
         )
     };
-    if status != kAudioHardwareNoError {
+    if status != kAudioHardwareNoError as i32 {
         return None;
     }
 
@@ -135,7 +135,7 @@ pub fn default_output_device() -> Option<Device> {
             NonNull::from(&mut audio_device_id).cast(),
         )
     };
-    if status != kAudioHardwareNoError {
+    if status != kAudioHardwareNoError as i32 {
         return None;
     }
 
--- a/src/host/coreaudio/macos/mod.rs
+++ b/src/host/coreaudio/macos/mod.rs
@@ -24,7 +24,8 @@ use objc2_core_audio::{
     kAudioObjectPropertyScopeGlobal, kAudioObjectPropertyScopeInput,
     kAudioObjectPropertyScopeOutput, AudioDeviceID, AudioObjectGetPropertyData,
     AudioObjectGetPropertyDataSize, AudioObjectID, AudioObjectPropertyAddress,
-    AudioObjectPropertyScope, AudioObjectSetPropertyData,
+    AudioObjectPropertyScope, AudioObjectSetPropertyData, kAudioHardwarePropertyDefaultInputDevice,
+    kAudioObjectSystemObject, kAudioHardwarePropertyDefaultOutputDevice
 };
 use objc2_core_audio_types::{
     AudioBuffer, AudioBufferList, AudioStreamBasicDescription, AudioValueRange,
@@ -164,10 +165,48 @@ impl Device {
     /// Construct a new device given its ID.
     /// Useful for constructing hidden devices.
     pub fn new(audio_device_id: AudioDeviceID) -> Self {
+
+        // Get default input device ID.
+        let mut property_address = AudioObjectPropertyAddress {
+            mSelector: kAudioHardwarePropertyDefaultInputDevice,
+            mScope: kAudioObjectPropertyScopeGlobal,
+            mElement: kAudioObjectPropertyElementMaster,
+        };
+
+        let mut default_input_device_id: AudioDeviceID = 0;
+        let data_size = mem::size_of::<AudioDeviceID>() as u32;
+        let input_status = unsafe {
+            AudioObjectGetPropertyData(
+                kAudioObjectSystemObject as AudioObjectID,
+                NonNull::from(&property_address),
+                0,
+                null(),
+                NonNull::from(&data_size),
+                NonNull::from(&mut default_input_device_id).cast(),
+            )
+        };
+
+        // Get default output device ID.
+        property_address.mSelector = kAudioHardwarePropertyDefaultOutputDevice;
+
+        let mut default_output_device_id: AudioDeviceID = 0;
+        let output_status = unsafe {
+            AudioObjectGetPropertyData(
+                kAudioObjectSystemObject as AudioObjectID,
+                NonNull::from(&property_address),
+                0,
+                null(),
+                NonNull::from(&data_size),
+                NonNull::from(&mut default_output_device_id).cast(),
+            )
+        };
+
+
         Device {
             audio_device_id,
-            // TODO: This could be made to detect the default device properly.
-            is_default: false,
+            is_default: 
+                (default_input_device_id == audio_device_id && input_status == 0) || 
+                (default_output_device_id == audio_device_id && output_status == 0)
         }
     }
 
--- a/src/host/coreaudio/macos/mod.rs
+++ b/src/host/coreaudio/macos/mod.rs
@@ -20,10 +20,9 @@ use objc2_core_audio::{
     kAudioDevicePropertyAvailableNominalSampleRates, kAudioDevicePropertyBufferFrameSize,
     kAudioDevicePropertyBufferFrameSizeRange, kAudioDevicePropertyDeviceIsAlive,
     kAudioDevicePropertyNominalSampleRate, kAudioDevicePropertyStreamConfiguration,
-    kAudioDevicePropertyStreamFormat, kAudioHardwarePropertyDefaultInputDevice,
-    kAudioHardwarePropertyDefaultOutputDevice, kAudioObjectPropertyElementMaster,
+    kAudioDevicePropertyStreamFormat, kAudioObjectPropertyElementMaster,
     kAudioObjectPropertyScopeGlobal, kAudioObjectPropertyScopeInput,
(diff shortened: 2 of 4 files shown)


src/host/coreaudio/macos/enumerate.rs
```diff
@@ -105,7 +105,7 @@ pub fn default_input_device() -> Option<Device> {
             NonNull::from(&mut audio_device_id).cast(),
         )
     };
-    if status != kAudioHardwareNoError as i32 {
+    if status != kAudioHardwareNoError {
```
roderickvd: I like the opportunistic refactoring. I see that there are a whole bunch of clippy warnings that the CI doesn't check for; I'll submit a PR separately. Sorry if that causes any merge conflicts.

src/host/coreaudio/macos/mod.rs
```diff
@@ -164,10 +165,44 @@ impl Device {
     /// Construct a new device given its ID.
     /// Useful for constructing hidden devices.
     pub fn new(audio_device_id: AudioDeviceID) -> Self {
+        // Get default input device ID.
```
roderickvd: There's similar logic in `enumerate.rs`, see the `default_output_device` function. We should unify that.

src/host/coreaudio/macos/mod.rs
```diff
@@ -164,10 +165,44 @@ impl Device {
...
+                null(),
+                NonNull::from(&data_size),
+                NonNull::from(&mut default_output_device_id).cast(),
+            )
+        };
         Device {
             audio_device_id,
-            // TODO: This could be made to detect the default device properly.
-            is_default: false,
+            is_default: (default_input_device_id == audio_device_id && input_status == 0)
```
roderickvd: Style nitpick: generally when there's a bit more logic involved, like is the case here, I prefer that we do `let is_default = ...` somewhere above. Also, we can return `Self { ... }` instead of repeating the struct name like `Device { ... }`.

src/samples_formats.rs
```diff
@@ -46,15 +46,13 @@ pub enum SampleFormat {
     /// `u16` with a valid range of `u16::MIN..=u16::MAX` with `1 << 15 == 32768` being the origin.
     U16,
 
-    /// `U24` with a valid range of '0..16777216' with `1 << 23 == 8388608` being the origin
+    // /// `U24` with a valid range of '0..16777216' with `1 << 23 == 8388608` being the origin
```
roderickvd: Double code comment block.

src/samples_formats.rs
```diff
@@ -46,15 +46,13 @@ pub enum SampleFormat {
...
     // U24,
-
     /// `u32` with a valid range of `u32::MIN..=u32::MAX` with `1 << 31` being the origin.
     U32,
 
-    /// `U48` with a valid range of '0..(1 << 48)' with `1 << 47` being the origin
+    // /// `U48` with a valid range of '0..(1 << 48)' with `1 << 47` being the origin
```
roderickvd: Double code comment block.

src/host/coreaudio/mod.rs
```diff
@@ -26,7 +26,6 @@ pub use self::macos::{
 };
 
 /// Common helper methods used by both macOS and iOS
-
```
roderickvd: The comment on line 28 isn't actually a code comment for the function below, rather the start of a file section. The newline should probably remain, and the `///` shortened to `//`. I'll tackle it in my clippy fix branch too.

src/host/coreaudio/macos/enumerate.rs
wgibbs-rs: I put it back to how it was, and left it alone! Thanks for catching that.

src/host/coreaudio/macos/mod.rs
wgibbs-rs: I completely rewrote what I had to use these functions. It's a *lot* simpler now, I have to say.

src/host/coreaudio/macos/mod.rs
wgibbs-rs: Done!

src/samples_formats.rs
wgibbs-rs: Fixed.

src/samples_formats.rs
wgibbs-rs: Fixed.

src/host/coreaudio/mod.rs
wgibbs-rs: Ok! I kept it the way it was.

wgibbs-rs:
@roderickvd 

Thanks for the helpful reviews on my PR. I've gone through each one, and fixed it or left it for you to change in a future PR.

Also, I did not change

wgibbs-rs:
Shoot! Hit "Close with commit." I'll see if I can get it back...

Thanks for the help, that was really insightful. I fixed each remark you had, and made sure it still works on my machine. I rewrote the code I had, including the functions from `enumerate.rs`.

Also, I did not change the return type to `Result<Self, BackendSpecificError>` as that would be unnecessary with these new changes. As the function currently is, no BackendSpecificError would be possible.

Let me know what you think, and I can update anything else you need.

wgibbs-rs:
(Is this reopened now!?)

src/host/coreaudio/macos/mod.rs
roderickvd: Good. I have the feeling we can do better still than doing `unwrap_or(false)` twice. Copilot review below also hints at that. Maybe something like:

```rust
fn is_default_device(audio_device_id: &AudioDeviceId) -> bool {
    let is_default_input = default_input_device()
        .map_or(false, |d| d.audio_device_id == *audio_device_id);
    
    let is_default_output = default_output_device()
        .map_or(false, |d| d.audio_device_id == *audio_device_id);
    
    is_default_input || is_default_output
}
```

src/host/coreaudio/macos/mod.rs
```diff
@@ -164,10 +164,19 @@ impl Device {
     /// Construct a new device given its ID.
     /// Useful for constructing hidden devices.
     pub fn new(audio_device_id: AudioDeviceID) -> Self {
-        Device {
+
+        let is_default =
```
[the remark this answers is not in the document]
roderickvd: This is something to ponder. If we cache it, it won't pick up on changes in default devices. Anyway it's questionable how often these methods are called. If we *do* decide to cache it then this could get you started:

```rust
use std::cell::OnceCell;

struct AudioDeviceCache {
    input: OnceCell<Option<AudioDevice>>,
    output: OnceCell<Option<AudioDevice>>,
}

impl AudioDeviceCache {
    fn new() -> Self {
        Self {
            input: OnceCell::new(),
            output: OnceCell::new(),
        }
    }
    
    fn default_input(&self) -> Option<&AudioDevice> {
        self.input.get_or_init(|| default_input_device()).as_ref()
    }
}
```

I doubt whether we should, what do you think?

src/host/coreaudio/macos/mod.rs
```diff
@@ -164,10 +164,19 @@ impl Device {
...
-        Device {
+
+        let is_default = 
+            default_input_device()
+                .map(|d| d.audio_device_id == audio_device_id)
+                .unwrap_or(false)
+            ||
+            default_output_device()
+                .map(|d| d.audio_device_id == audio_device_id)
+                .unwrap_or(false);
```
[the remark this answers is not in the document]
roderickvd: No, we're not gonna print output in the library.

roderickvd:
Yes, it's reopened :smile:

If you could resolve the merge conflict and look at Copilot's comments? I doubt the point about caching, but agree with that the helper function would be a cherry on top.

src/host/coreaudio/macos/mod.rs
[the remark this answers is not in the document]
wgibbs-rs: Hm, that's a really interesting point.

I wasn't sure *how* much of an impact this would be, so I tried stress-testing `cpal` for this.

`Time elapsed in calling Device::new 1 million times: 36.66s`

While this might not be the *best* way to test this, it puts it into perspective—if using cpal normally results in 1-2 devices, and rarely exceeds ~10 (I used 10 in a DAW I contributed to), then I don't think that sort of latency really makes a difference.

What could also make a difference is caching data—though its impact on RAM will still also be extremely small.

But if caching sounds right, and that's a path we want to go on for `cpal` overall, then starting here might also be the right choice.

wgibbs-rs:
I created the helper function, and gave my 2-cents on caching the default devices.

Just let me know if there's anything else I should do!

roderickvd:
Having second thoughts about this `is_default` struct field. We're only using it in `build_input_stream_raw` and `build_output_stream_raw` which both call `audio_unit_from_device`. So why not remove the field altogether, and instead check whether the device is the default device in those functions?

This seems better because it:
- prevents race conditions
- saves some memory

Example race condition:
1. default device is `Foo`
2. `cpal` enumerates devices
3.  default device is changed to `Bar`
4. `cpal` opens `Foo` as default `<-- INCORRECT`

wgibbs-rs:
I like this idea.

I wasn't *100%* sure what you meant, so I interpreted it as removing `is_default` entirely, and instead using a function to determine this on-demand.

I went back and replaced all instances of `is_default` with `is_default_device` or removed them entirely. The code builds correctly, and everything seems alright with this new method.

How does this sound? If I misinterpreted what you said, just let me know and I'll fix it!

src/host/coreaudio/mod.rs
```diff
@@ -25,7 +25,8 @@ pub use self::macos::{
     Device, Host, Stream,
 };
 
-/// Common helper methods used by both macOS and iOS
+
+// Common helper methods used by both macOS and iOS
```
roderickvd: Keep the newline between the comment and the next function - this is only a file section header.

src/host/coreaudio/macos/mod.rs
```diff
@@ -159,26 +159,25 @@ impl DeviceTrait for Device {
 #[derive(Clone, PartialEq, Eq)]
 pub struct Device {
     pub(crate) audio_device_id: AudioDeviceID,
-    is_default: bool,
 }
 
+fn is_default_device(device: Device) -> bool {
```
roderickvd: Can you take a `&Device` here so we don't need to clone in `audio_unit_from_device`?

wgibbs-rs:
I fixed the `device.clone()` issues, and the function now takes a `&Device`. I also added a new entry to `CHANGELOG.md`!

Also, I updated my rust edition to the latest version. Now, a warning appears. This is not related to my branch specifically, so I will leave it as is.

```
warning: struct `NotSendSyncAcrossAllPlatforms` is never constructed
   --> src/platform/mod.rs:759:8
    |
759 | struct NotSendSyncAcrossAllPlatforms(std::marker::PhantomData<*mut ()>);
    |        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
```

 If there's anything else you need, just let me know!

roderickvd:
Merged, thanks!

PR #999, opened 2025-08-12
Update Dockerfile to modern syntax and include dependencies

wgibbs-rs:
I found that using Docker was a pain on Debian Linux, as it had no documentation and did not already include the required files, like the Rust toolchain, curl, and build tools.

I've updated the `Dockerfile` to both use modern Docker syntax ( ENV key=var ) and to automatically install the Rust toolchain, curl (required for Rust), and the build-essentials needed for CC to work.

I also updated the `CHANGELOG.md` file to include these changes.

I've tested this on my machine, and found no issues (besides taking a while on the first build).

Let me know if there's anything else I need to do!

### DIFF
--- a/.github/workflows/cpal.yml
+++ b/.github/workflows/cpal.yml
@@ -78,32 +78,27 @@ jobs:
     - name: Run all features
       run: cargo test --all --all-features --verbose
 
-  linux-check-and-test-armv7:
-    runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


wgibbs-rs:
Hold on- I'm trying to fix the Github Actions runner, as it will need an update with the new Dockerfile.

wgibbs-rs:
nevermind.... I'll be back!

PR #1002, opened 2025-08-14
remove unused `NotSendSyncAcrossAllPlatforms` struct and impl

wgibbs-rs:
After updating to Rust vs 1.89.0, a new `struct 'NotSendSyncAcrossAllPlatforms' is never constructed` warning appears when building on any platform.

I've removed this code, and the project builds, tests, and example "beep" runs correctly on both MacOS and arm64 Linux.

According to @roderickvd , this was code likely left over from `oboe` when moving to `ndk` on Android.


roderickvd:
Merged, thanks!

PR #1003, opened 2025-08-16
Support loopback recording on macOS

SuperKenVery:
This pr adds support for recording system output audio on macOS.

For usage example, please see `src/host/coreaudio/macos/mod.rs` `test::test_record_output`.

To test this, use `cargo run --example record_wav --device default-output`

Testing requirements:
- You're on macOS 14.2 or later. 
- The app has permission for recording system audio. Note the "app" means the app that starts `cargo run --example`, e.g. VSCode or Zed.
- Don't build with nix. It comes with an old macOS SDK and lacks required symbols.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,10 @@ edition = "2021"
 rust-version = "1.70"
 
 [features]
-asio = ["asio-sys", "num-traits"] # Only available on Windows. See README for setup instructions.
+asio = [
+    "asio-sys",
+    "num-traits",
+] # Only available on Windows. See README for setup instructions.
 
 # Deprecated, the `oboe` backend has been removed
 oboe-shared-stdcxx = []
@@ -36,8 +39,8 @@ windows = { version = "0.54.0", features = [
     "Win32_System_SystemServices",
     "Win32_System_Variant",
     "Win32_Media_Multimedia",
-    "Win32_UI_Shell_PropertiesSystem"
-]}
+    "Win32_UI_Shell_PropertiesSystem",
+] }
 audio_thread_priority = { version = "0.33.0", optional = true }
 asio-sys = { version = "0.2", path = "asio-sys", optional = true }
 num-traits = { version = "0.2.6", optional = true }
@@ -60,6 +63,8 @@ objc2-core-audio = { version = "0.3.1", default-features = false, features = [
     "std",
     "AudioHardware",
     "AudioHardwareDeprecated",
+    "objc2",
+    "objc2-foundation",
 ] }
 objc2-audio-toolbox = { version = "0.3.1", default-features = false, features = [
     "std",
@@ -70,20 +75,44 @@ objc2-core-audio-types = { version = "0.3.1", default-features = false, features
     "std",
     "CoreAudioBaseTypes",
 ] }
+objc2-core-foundation = { version = "0.3.1" }
+objc2-foundation = { version = "0.3.1" }
+objc2 = { version = "0.6.2" }
 
 [target.'cfg(target_os = "emscripten")'.dependencies]
 wasm-bindgen = { version = "0.2.89" }
 wasm-bindgen-futures = "0.4.33"
 js-sys = { version = "0.3.35" }
-web-sys = { version = "0.3.35", features = [ "AudioContext", "AudioContextOptions", "AudioBuffer", "AudioBufferSourceNode", "AudioNode",  "AudioDestinationNode", "Window", "AudioContextState"] }
+web-sys = { version = "0.3.35", features = [
+    "AudioContext",
+    "AudioContextOptions",
+    "AudioBuffer",
+    "AudioBufferSourceNode",
+    "AudioNode",
+    "AudioDestinationNode",
+    "Window",
+    "AudioContextState",
+] }
 
 [target.'cfg(all(target_arch = "wasm32", target_os = "unknown"))'.dependencies]
 wasm-bindgen = { version = "0.2.58", optional = true }
 js-sys = { version = "0.3.35" }
-web-sys = { version = "0.3.35", features = [ "AudioContext", "AudioContextOptions", "AudioBuffer", "AudioBufferSourceNode", "AudioNode",  "AudioDestinationNode", "Window", "AudioContextState"] }
+web-sys = { version = "0.3.35", features = [
+    "AudioContext",
+    "AudioContextOptions",
+    "AudioBuffer",
+    "AudioBufferSourceNode",
+    "AudioNode",
+    "AudioDestinationNode",
+    "Window",
+    "AudioContextState",
+] }
 
 [target.'cfg(target_os = "android")'.dependencies]
-ndk = { version = "0.9", default-features = false, features = ["audio", "api-level-26"]}
+ndk = { version = "0.9", default-features = false, features = [
+    "audio",
+    "api-level-26",
+] }
 ndk-context = "0.1"
 jni = "0.21"
 num-derive = "0.4"
--- a/src/host/coreaudio/macos/device.rs
+++ b/src/host/coreaudio/macos/device.rs
@@ -1,0 +1,1101 @@
+#![allow(deprecated)]
+use super::OSStatus;
+use super::Stream;
+use super::{asbd_from_config, check_os_status, frames_to_duration, host_time_to_stream_instant};
+use crate::host::coreaudio::macos::StreamInner;
+use crate::traits::{DeviceTrait, HostTrait, StreamTrait};
+use crate::{
+    BackendSpecificError, BufferSize, BuildStreamError, ChannelCount, Data,
+    DefaultStreamConfigError, DeviceNameError, DevicesError, InputCallbackInfo, OutputCallbackInfo,
+    PauseStreamError, PlayStreamError, SampleFormat, SampleRate, StreamConfig, StreamError,
+    SupportedBufferSize, SupportedStreamConfig, SupportedStreamConfigRange,
+    SupportedStreamConfigsError,
+};
+use coreaudio::audio_unit::render_callback::{self, data};
+use coreaudio::audio_unit::{AudioUnit, Element, Scope};
+use coreaudio::error::audio_unit;
+use objc2::rc::{autoreleasepool, Retained};
+use objc2::AnyThread;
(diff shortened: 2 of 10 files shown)


SuperKenVery:
This pr might seem a little bit messy, but most of the new code should be at `src/host/coreaudio/macos/loopback.rs`.

I also extracted some code from`mod.rs` to `device.rs`, and added some tests in `mod.rs`.

For `Cargo.toml`, the main change was adding some features in `objc2-xxx` and/or add new `objc2-xxx` dependencies. However, the file was automatically formatted by Zed so it looks like there're lots of changes. But formatting it is anyway a good thing, so I didn't bother reverting that.

SuperKenVery:
It seems that, in the CI, `test_record_output` test isn't properly finishing. Not sure why, it runes fine on my machine...

Well, CI is on macOS 14.7 while I'm on 15.5. Can anyone on macOS 14 try this?

wgibbs-rs:
There are some features not enabled by default. Given the kind of errors produced, I doubt this is a system-specific problem. Make sure you run the proper build commands, run unit tests, and try an example (like "beep"). To properly test this on your system, try using the following commands:

`cargo test --all --no-default-features --verbose`
`cargo test --all --all-features --verbose`

What could have changed that impacted `wasm`? I see the error around this is a time-out issue, so maybe try fixing MacOS and see if that resolves the `wasm` issue.

SuperKenVery:
@wgibbs-rs Thanks for your suggestions!

Both of the following runs flawlessly on my machine:
```
cargo test --all --no-default-features --verbose
cargo test --all --all-features --verbose
```

The wasm error is probably a network issue, it didn't even get to the point to run my code. It failed at "install stable".

roderickvd:
Re-ran the failed tests and they’re OK now. Edit: no you’re right, the macOS test doesn’t finish. I only have an up to date macOS too. 

Thanks for your contribution! This seems like a worthwhile feature to have. I am away now and should be able to look more into it in a few days. I’ll start by triggering a Copilot review and see what it comes up with.

roderickvd:
Any ideas how we could get it to work on macOS 14?

SuperKenVery:
I think it might not be very difficult, the only problem is I don't have a macOS 14 machine.

When I get time, I could probably try to debug this in a VM.

roderickvd:
Maybe it's because the CI is asking the (non-existent) user whether microphone access is allowed? Maybe have it time out?

SuperKenVery:
> the CI is asking the (non-existent) user whether microphone access is allowed

I think that might be the case. But why does recording test pass? It should also ask for microphone access...

SuperKenVery:
Ok. I just fired up a macOS 14 VM. 

It does ask for microphone permission. But after granting the permission, it does ran without problems. So, not sure what happened on CI 🤔

roderickvd:
I'm no expert but maybe the problem is that loopback requires different permissions: screen capture, instead of microphone for the recording test.

I'd be OK with skipping such tests conditionally on the CI:

```rust
  #[test]
  fn test_foo() {
      if std::env::var("CI").is_ok() {
          println!("Skipping test_foo in CI environment due to permissions");
          return;
      }
      // ...
  }
```

simlay:
Huh. I'm getting some odd behavior with the example. I've spent probably 20+ minutes trying to narrow it down but I'm more confused than I was initially. I'm on macOS 15.6.1, I've swapped between xcode 26 beta 6 and xcode 16.4 via `xcode-select --switch` and back. At first I thought it wasn't working as expected because my sample audio was a shakira youtube video in firefox (not trying to copy, just lazy).

I had it working in a non famous youtube music video and when I play local media. I'm not a huge audio person but the audio quality sounded like "It was under water". I wanted to provide the input recorded audio and the output from the loopback but somehow in the last 15 minutes something changed and now I don't hear anything when I play `recorded.wav`. I wish I had something more specific to be helpful.

SuperKenVery:
@simlay Thanks for reporting!

About the youtube video, I think it might be related to DRM, but I'm not sure. If you could provide its URL I may find out something.

About the sound quality, I'm not sure what's happening. Something that comes to my mind:
1. According to [this](https://gist.github.com/directmusic/7d653806c24fe5bb8166d12a9f4422de), macOS may have bugs handling outputs with a lot of channels:
```cpp
        // Create a tap description for all processes without a specific output.
        // Note: I believe there is a bug in the CoreAudio Tap API. If the
        // default output device has 2 output channels this works as expected.
        // But if you have a device with 4 output channels then the volume of
        // the resulting buffer will be halved. You can extrapolate this to any
        // number of channels.

        // This bug is also present in ScreenCaptureKit if you use it on macOS
        // 14.2 or later. The bug is not present in macOS 14.1 or earlier.

        // A workaround for this issue could be to increase the volume of the
        // audio in the output by the number of channels. Or simply use the
        // other API to tap a specific output.
```
2. Make sure you are recording from "default-output" device, and not from your microphone.

About playing the wav file, well, that's really strange! Hope you would figure it out~

SuperKenVery:
@roderickvd 

> loopback requires different permissions: screen capture

In my case (macOS 15.6.1) there are two sections in the screen capture permissions, "Screen capture and system audio" and "system audio only" (I translated from Chinese, might not be word accurate). It should require the "system audio only" permission.

However on macOS 14, the OS only asks for the microphone permission when running the tests.

I'm personally OK with skipping it on CI, but I'm curious why the microphone tests pass on CI? I didn't see any permission granting thing on the workflow file...

roderickvd:
I don’t know and can only guess that it has something to do with these permissions. Fixing the CI without skipping the test would be great but if we can’t win the fight, then let’s spend our energy elsewhere.

SuperKenVery:
> then let’s spend our energy elsewhere

Agreed. I've now disabled the check on CI and all checks have passed 😄

src/host/coreaudio/macos/device.rs
```diff
@@ -0,0 +1,884 @@
...
+                NonNull::from(&property_address),
+                0,
+                null(),
+                NonNull::from(&data_size),
+                NonNull::new(ranges.as_mut_ptr()).unwrap().cast(),
+            )
+        };
+        coreaudio::Error::from_os_status(status)?;
+        let ranges: *mut AudioValueRange = ranges.as_mut_ptr() as *mut _;
+        let ranges: &'static [AudioValueRange] = unsafe { slice::from_raw_parts(ranges, n_ranges) };
```
roderickvd: Should this be `&'static` or just `&`?

src/host/coreaudio/macos/loopback.rs
```diff
@@ -0,0 +1,250 @@
...
+            aggregate_device: Device::new(aggregate_device_id),
+        })
+    }
+}
+
+impl Drop for LoopbackDevice {
+    fn drop(&mut self) {
+        unsafe {
+            let status = AudioHardwareDestroyAggregateDevice(self.aggregate_device.audio_device_id);
+            check_os_status(status).unwrap();
```
roderickvd: To prevent panics during drop, would it be OK to ignore the result? `let _ = check_os_status(status);`

Cargo.toml
```diff
@@ -70,20 +75,44 @@ objc2-core-audio-types = { version = "0.3.1", default-features = false, features
     "std",
     "CoreAudioBaseTypes",
 ] }
+objc2-core-foundation = { version = "0.3.1" }
```
roderickvd, line 78: Is it necessary to pull them in explicitly? A few lines above they are already selected as features.

src/host/coreaudio/macos/device.rs
SuperKenVery: This code isn't really written by me... I just moved it to another file.

But looking at the code, it seems it should be `&` as it's a locally-declared `Vec`. But maybe it didn't return this pointer so it didn't cause any problem.

src/host/coreaudio/macos/loopback.rs
SuperKenVery: Indeed, resource leakage is probably better than panic.

Cargo.toml
SuperKenVery: This is my first time using `objc2` for rust, so I might be wrong. But I referenced some types in `objc2-core-foundation` explicitly: 

https://github.com/SuperKenVery/cpal/blob/5f6696376aa5219cb97feb202858e2dc9e1afc49/src/host/coreaudio/macos/loopback.rs#L18-L23

```rust
use objc2_core_foundation::{
    kCFAllocatorDefault, kCFTypeArrayCallBacks, kCFTypeDictionaryKeyCallBacks,
    kCFTypeDictionaryValueCallBacks, CFArray, CFDictionary, CFMutableDictionary, CFRetained,
    CFString, CFStringCreateWithCString,
};
use objc2_foundation::{ns_string, NSArray, NSNumber, NSString};
```

I am not aware of any way that I could use these without using `objc2-core-foundation`, or not using these types at all...

roderickvd: You're right.

src/host/coreaudio/macos/device.rs
roderickvd: Would you check?

SuperKenVery: Yes sure. Will do that will I get back to my computer~

SuperKenVery: I think the `'static`s are indeed unnecessary. I've removed them and ran `cargo test`, and everything seems fine.

roderickvd:
Perfect! This deserves mention in the changelog and we're good to merge!

roderickvd:
Busy rebasing on your branch... gonna merge this when the CI passes. Thanks!

zheng93775:
This feature is exactly what I need, and I'm very grateful that someone implemented it. My question is whether the code has already been merged into master, and if Cargo has not yet updated to release the latest code? Currently, cpal = "0.16.0" does not support this feature; when can we expect it to be released on Cargo?

SuperKenVery:
It has been merged into master, but we would need a release for it to appear in cargo registry. If you don't want to wait, you could use this github repo (rather than cargo registry) as the package source, like this:
```toml
[dependencies]
cpal = { git = "https://github.com/RustAudio/cpal.git" }
```
(not verified, but should work)

zheng93775:
Thank you very much, it works very well.

PR #957, opened 2025-02-27
PulseAudio support

colinmarc:
Fixes #259.

I finally got around to this after promising to PR it in #259 a year ago. :sweat_smile: 

This is based on https://github.com/colinmarc/pulseaudio-rs/pull/2, which obviously needs to land before this. However, I developed both PRs in parallel. If you're feeling generous and would like to review that as well, I would welcome any feedback.

The first commit is unrelated and a bit opinionated, but it seemed nicer. Let me know if I should move that to a separate PR or just drop it.

PR #1004, opened 2025-08-24
Use associated types instead of magical names in platform_impl

colinmarc:
This reduces the exported surface area of each host implementation, and makes the `impl_platform_host` macro more robust.

This was originally part of #957, but it was requested that I pull it out.


roderickvd:
Nice improvement for robustness and maintainability. Thank you, merging.

Issue #981, opened 2025-06-07
Looking for maintainers


est31:
## A bit of history of cpal
@tomaka has created the `cpal` library a over decade ago[^1], as a library to provide an abstraction layer over the different audio backends of the operating systems. In fact, @tomaka is one of the giants on whose shoulders the Rust community is standing, having created multiple projects, many of which have become key libraries for their categories, like rodio, glium, glutin, winit.

In 2018 (I think, I don't have precise records), @tomaka has added @mitchmindtree as a co-maintainer. Eventually, in 2019[^2], it was moved into the RustAudio organization. It was due to @tomaka having less time to dedicate for his old projects as he focused on his new job.

## A bit of history on me

I have been fascinated with audio since the start of my involvement in Rust. In fact, some of my first projects were an ogg and vorbis decoder ([ogg](https://github.com/RustAudio/ogg) and [lewton](https://github.com/RustAudio/lewton)). A bit after their creation, I have moved the projects to the RustAudio org.

The first PR I merged in cpal was in August 2019, 6bf00f176a02f4cd06af3550dee6597e7da11d0e. Initially @mitchmindtree and me maintained cpal together, but as he stopped maintaining, I was on my own.

Of the projects mentioned above, I've also become maintainer of glium and rodio.

I have a full time job in Rust now that I really love, and it needs a lot of my time. So my situation is similar to the one @tomaka was in.

## cpal maintenance now

In fact, I'd say that over the last 2-3 years, I haven't been a good steward of cpal. I've left PRs unreviewed, didn't respond to people making PRs, etc. People became frustrated, closed PRs, etc. And, of course, I'm not paid for cpal, especially not by the folks opening PRs, so they can't expect anything, but their emotions are also justified. I'd become frustrated too if nobody read my PR for months and years. I'm not happy about this, and I'm sorry!

## call for maintainers

As explained, the need for a maintainer exists for a longer time already. But today, with the 0.16 release, I want make an explicit public call for maintainers of cpal. Please reach out to me via the rust-lang zulip (est31) or via this github thread.

## maintaining cpal

Maintaining cpal is not easy. A lot of people rely on it so the ability to do damage is there. PRs might not contain actually good solutions to a problem. It's very low level code. If you are a single maintainer, you need to be expert for audio libraries for multiple operating systems. Ideally the burden is shared among a group where each person can specialize. In fact, part of why I am not a good maintainer is that I know surprisingly little about them.

[...]

PR #1006, opened 2025-08-29
asio-sys: Fixes to get cross-compiling under mingw64

JerwuQu:


roderickvd:
Thank you! Would you top this off with an update to the changelog? Good to merge then.

JerwuQu:
Done!

Completely off topic, but there was a suggestion to split out the asio-sys crate to be separate from cpal since it may be easier to maintain that way. Thoughts?

roderickvd:
> Done!

Great, will merge now.

> Completely off topic, but there was a suggestion to split out the asio-sys crate to be separate from cpal since it may be easier to maintain that way. Thoughts?

Where was that suggestion?

JerwuQu:
Awesome, thanks!

The suggestion was [over here](https://github.com/RustAudio/cpal/issues/981#issuecomment-2954054713) at the bottom.

roderickvd:
> The suggestion was [over here](https://github.com/RustAudio/cpal/issues/981#issuecomment-2954054713) at the bottom.

I see. I think it's a good way to go. To those that are interested in carrying longer term maintenance of it too, I propose that they set up a repo extracting `asio-sys` and put in a PR with `cpal` to extract it.

PR #1009, opened 2025-09-07
fix: improve ALSA period configuration for device compatibility

roderickvd:
Use time-based period configuration with set_period_time_near for better device compatibility. Calculate period time based on buffer size and period count, and fall back to safe defaults if constraints are unavailable.

## Problem

#990 broke playback on one of my USB devices with the following error immediately on opening the stream:

```text
A backend-specific error has occurred: `alsa::poll()` returned POLLERR
```

This PR fixes opening the device, but still runs into `POLLERR` after several minutes of playback when the system is under load. This is when requesting `BufferSize::Fixed(4410)` for 100 ms on 44.1 kHz.

## Way Forward

I submitted PR https://github.com/diwic/alsa-rs/pull/137 to expose more Alsa getters and setters relevant to periods and buffers. For now I'm putting this PR in draft in hopes of getting that one through not too long from now.

cc: @attackgoat


roderickvd:
As the current code in `master` is now unstable for some ALSA devices, I am going to merge this now. Let's collaborate on https://github.com/RustAudio/cpal/pull/1010 to be able to programmatically set the number of periods.

PR #1014, opened 2025-09-23
feat: stable Device::id()

xephyris:
This PR adds a `Device::id()` command that returns the OS native device id on supported operating systems.
~~Currently, only macOS and Windows (WASAPI) are supported.~~

### Function

By returning a `Device::id()`, it will be easier to ensure that the same device gets connected on application startup. This also allows for greater extension capability for other libraries to extend upon `cpal` function.

### Architecture

The OS specific device ids are all stored in a `DeviceId` enum, that allows for support for the various data types that different operating systems and APIs use to store device ids (e.g. `u32` for macOS and `String` for WASAPI.

``` rust
#[derive(Clone, Debug, Eq, PartialEq)]
pub enum DeviceId {
    CoreAudio(String),
    WASAPI(String),
    // ..
}
```

To handle any errors that might be outputted in when running `Device::id()`, there is also a new `DeviceIdError` enum
```rust
#[derive(Clone, Debug, PartialEq, Eq, Hash)]
pub enum DeviceIdError {
    /// See the [`BackendSpecificError`] docs for more information about this error variant.
    BackendSpecific { err: BackendSpecificError },
    UnsupportedOS,
    ParseError,
}
```

Calling `Device::id()` will return a `Result<DeviceId, DeviceIdError>`. If the OS is supported and the device id is obtained, it will return a `DeviceId`. If the OS is unsupported, a `DeviceIdError::UnsupportedOS` error will be returned.

### Current Status
~~Currently, support for Windows WASAPI `String` Ids has been added, along with support for macOS `u32` Ids. I am working to add support for `ALSA` and maybe `pulseaudio` (once that gets merged).~~

For other APIs, I do not have the hardware to test or program for them, ~~so they currently return a `DeviceIdError::UnsupportedOS` error.~~

### Testing
`Device::id()` for macOS should return the ~~standard `u32`~~ device uid that can be used anywhere. 
For Windows, to convert the Rust `String` type to a `PWSTR`, the following code can be used
```rust
// device_id is the id returned by cpal
 let mut id = format!("{}\0", device_id).encode_utf16().collect::<Vec<u16>>(); 
 let pwstr = PWSTR(id.as_mut_ptr());
```


src/lib.rs
```diff
@@ -219,6 +219,15 @@ where
 /// The desired number of frames for the hardware buffer.
 pub type FrameCount = u32;
 
+/// The device ID of the audio device, on supported OSs
+/// Currently only supports macOS and Windows (WASAPI)
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub enum DeviceId {
+    MacOS(u32),
```
roderickvd: For macOS, it's better to use the `kAudioDevicePropertyDeviceUID` instead of the AudioDeviceID, because the device UID doesn't change on reboot.

Also better to rename to `CoreAudio` as there may be multiple hosts on the macOS platform.

src/lib.rs
```diff
@@ -219,6 +219,15 @@ where
 /// The desired number of frames for the hardware buffer.
 pub type FrameCount = u32;
 
+/// The device ID of the audio device, on supported OSs
+/// Currently only supports macOS and Windows (WASAPI)
+#[derive(Clone, Debug, Eq, PartialEq)]
```
roderickvd: You could add a variant `AAudio(i32)` with an implementation in `src/host/aaudio/mod.rs` like:

```rust
fn id(&self) -> Result<DeviceId, DeviceIdError> {
    match &self.0 {
        None => Ok(DeviceId::aaudio(-1)), // Default device
        Some(info) => Ok(DeviceId::aaudio(info.id)),
    }
}
```

And a variant `Jack(string)` with in `src/host/jack/mod.rs`:

```rust
fn id(&self) -> Result<DeviceId, DeviceIdError> {
    Ok(DeviceId::jack(self.name.clone()))
}
```

src/host/null/mod.rs
```diff
@@ -47,6 +47,11 @@ impl DeviceTrait for Device {
         Ok("null".to_owned())
     }
 
+    #[inline]
+    fn id(&self) -> Result<DeviceId, DeviceIdError> {
+        Err(DeviceIdError::UnsupportedOS)
```
roderickvd: This could simply return `Ok(DeviceId::Null)`.

src/traits.rs
```diff
@@ -89,6 +89,9 @@ pub trait DeviceTrait {
     /// The human-readable name of the device.
     fn name(&self) -> Result<String, DeviceNameError>;
 
+    /// The device-id of the device. (For supported OSs only)
+    fn id(&self) -> Result<DeviceId, DeviceIdError>;
```
roderickvd, line 102: Then on the `HostTrait`, it would be desired to have:

```rust
fn device_by_id(&self, id: &DeviceId) -> Option<Self::Device> {
     self.devices().ok()?.find(|device| {
        device.device_id().ok().as_ref() == Some(id)
    })
}
```

(or something smarter)

src/lib.rs
```diff
@@ -219,6 +219,15 @@ where
...
+pub enum DeviceId {
+    MacOS(u32),
+    Windows(String),
```
roderickvd: Let's stick to `Wasapi` as there could be multiple hosts on the Windows platform.

src/lib.rs
roderickvd: It'd be great to also implement `FromStr` and `Display`:

```rust
impl fmt::Display for DeviceId {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        match self {
            DeviceId::Wasapi(guid) => write!(f, "wasapi:{}", guid),
            DeviceId::CoreAudio(uid) => write!(f, "coreaudio:{}", uid),
            DeviceId::Alsa(pcm_id) => write!(f, "alsa:{}", pcm_id),
            DeviceId::AAudio(id) => write!(f, "aaudio:{}", id),
            DeviceId::Jack(name) => write!(f, "jack:{}", name),
        }
    }
}

impl FromStr for DeviceId {
    type Err = DeviceIdError;

    fn from_str(s: &str) -> Result<Self, Self::Err> {
        let (platform, data) = s.split_once(':')
            .ok_or(DeviceIdError:: ParseError)?;

        match platform {
            "wasapi" => Ok(DeviceId::Wasapi(data.to_string())),
            "coreaudio" => Ok(DeviceId::CoreAudio(data.to_string())),
            "alsa" => Ok(DeviceId::Alsa(data.to_string())),
            "aaudio" => {
                let id = data.parse().map_err(|_| DeviceIdError:: ParseError)?;
                Ok(DeviceId::AAudio(id))
            }
            "jack" => Ok(DeviceId::Jack(data.to_string())),
            _ => Err(DeviceIdError::UnsupportedPlatform),
        }
    }
}
```

roderickvd (changes requested):
This is great! Your proposal _literally_ crossed paths as I was working on something also. No matter, all the better I could provide some review comments with actual code suggestions.

Please also add a changelog, it's worth it 😄

xephyris:
Hello @roderickvd !

I have implemented most of the changes requested and renamed the enum variants to the suggested names.

For `jack` and `aaudio`, I just copied the code that you provided, and because I cannot test it, there may or may not be some errors.

I have also added `alsa` support which currently just returns the same content as name, because I was unable to find any other specific identifiers that could work better.

I will work on changing the macOS CoreAudio type to `kAudioDevicePropertyDeviceUID` in the coming days.

The changelog has also been updated 😄.

src/lib.rs
yury: Also if you unplug mic and plug it back in. u32 value will be new, but UID string will be the same.

roderickvd:
Cool stuff let me know when you're ready for me to take another look.

xephyris:
I have finished the changes and now `DeviceId::CoreAudio` has a `String` type instead of `u32`. 

One thing that I noticed was that there was already a `uid()` function in `loopback.rs` for `Device`, however I wasn't sure if it should be moved into `device.rs` or not as it seems to rely on some types inside of `loopback.rs`. 

The code I'm currently using is basically 90% the same as the `uid()` function, just with a different return type. If needed, I can attempt moving the `uid()` function over into `device.rs`

src/host/coreaudio/macos/device.rs
```diff
@@ -392,6 +403,37 @@ impl Device {
...
+            AudioObjectGetPropertyData(
+                self.audio_device_id,
+                NonNull::from(&property_address),
+                0,
+                null(),
+                NonNull::from(&data_size),
+                NonNull::from(&mut name).cast(),
+            )
+        };
+        if status == 0 {
```
roderickvd: Instead of checking `status == 0` would it make sense to `check_os_status(status)?` consistent with the rest of the code?

Then instead (or additionally) we should return an error `if uid.is_null()`.

src/host/coreaudio/macos/device.rs
```diff
@@ -392,6 +403,37 @@ impl Device {
         })
     }
 
+    fn id(&self) -> Result<DeviceId, DeviceIdError> {
+        let property_address = AudioObjectPropertyAddress {
+            mSelector: kAudioDevicePropertyDeviceUID,
+            mScope: kAudioObjectPropertyScopeGlobal,
+            mElement: kAudioObjectPropertyElementMain,
+        };
+        let mut name: CFStringRef = std::ptr::null_mut();
```
roderickvd: Is this the `name` or `uid`?

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # Unreleased
 
+- Add `DeviceId` enum supporting most APIs that returns a system device id for the audio device, allowing it to be reselected using the `HostTrait::device_by_id()` function
```
roderickvd: ⬇️ 

```markdown
- Add `HostTrait::id` method that returns a stable audio device ID.
- Add `HostTrait::device_by_id` to select a device by its stable ID.
```

src/host/asio/device.rs
```diff
@@ -54,6 +56,10 @@ impl Device {
         Ok(self.driver.name().to_string())
     }
 
+    fn id(&self) -> Result<DeviceId, DeviceIdError> {
+        Err(DeviceIdError::UnsupportedPlatform)
```
roderickvd: Not an ASIO user but wouldn't this work:

```rust
Ok(DeviceId::ASIO(self.driver.name().to_string()))
```

src/host/coreaudio/ios/mod.rs
```diff
@@ -74,6 +75,10 @@ impl Device {
         Ok("Default Device".to_owned())
     }
 
+    fn id(&self) -> Result<DeviceId, DeviceIdError> {
+        Err(DeviceIdError::UnsupportedPlatform)
```
roderickvd: I think it would be possible to get the input/output name through `AVAudioSession::shared_instance().current_route().outputs/inputs` but right now we always seem to only enumerate and always select the default device. Would it make sense to hardcode the default device here as a value?

src/host/emscripten/mod.rs
```diff
@@ -69,6 +69,11 @@ impl Device {
         Ok("Default Device".to_owned())
     }
 
+    #[inline]
+    fn id(&self) -> Result<DeviceId, DeviceIdError> {
+        Err(DeviceIdError::UnsupportedPlatform)
```
roderickvd: Same, would it make sense to always return `default`?

src/host/webaudio/mod.rs
```diff
@@ -84,6 +85,11 @@ impl Device {
         Ok("Default Device".to_owned())
     }
 
+    #[inline]
+    fn id(&self) -> Result<DeviceId, DeviceIdError> {
+        Err(DeviceIdError::UnsupportedPlatform)
```
roderickvd: Same.

src/lib.rs
```diff
@@ -219,6 +219,51 @@ where
...
+        match platform {
+            "wasapi" => Ok(DeviceId::WASAPI(data.to_string())),
+            "coreaudio" => Ok(DeviceId::CoreAudio(data.to_string())),
+            "alsa" => Ok(DeviceId::ALSA(data.to_string())),
+            "aaudio" => {
+                let id = data.parse().map_err(|_| DeviceIdError::ParseError)?;
+                Ok(DeviceId::AAudio(id))
+            }
+            "jack" => Ok(DeviceId::Jack(data.to_string())),
+            _ => Err(DeviceIdError::UnsupportedPlatform),
```
roderickvd: This may require syncing.

roderickvd (changes requested):
Some ideas and one point about the macOS implementation.

xephyris:
I have updated the `ASIO` configuration and fixed the naming for macOS.

I have also made `emscripten`, `webaudio`, and `iOS` return `Ok(DeviceId::"API_NAME"("default"))` instead of an error. 

Currently, I think it is fine as all three of those APIs seem to only return the default device, however I think if they were expanded to be able to select audio devices, it would be better to leave them returning `DeviceIdError::UnsupportedPlatform` than `default`, as if someone were to save the `id` while a different device is selected, it would incorrectly select the device.

As for the syncing issue, I'm unsure about what you mean, would you care to elaborate?

src/lib.rs
xephyris: I'm unsure about what you mean by this, would you care to elaborate?

roderickvd:
From your PR description, which I edited to show the progress made since:

> For Windows, to convert the Rust `String` type to a `PWSTR`, the following code can be used

Is that still necessary?

src/lib.rs
roderickvd: Sorry I meant to say that by now you’ve added more platforms that support it, so instead of matching on `_` I propose to include all variants and your code will be “in sync”.

xephyris:
Thanks for the update @roderickvd ! I'll try getting the code updated soon for another review.

> Is that still necessary?

Currently I believe it is still necessary to convert it to a `PWSTR` because to get the `IMMDevice` from the `IMMDeviceEnumerator`, it requires a `PWSTR` at least from what I recall. 

This is primarily for if you are using the `Device::id()` to extend upon `cpal` functionality like I'm currently working on `cpvc` and adding the volume controls. If you are just trying to re-select a device from `cpal`, then that code to convert to `PWSTR` isn't necessary.

roderickvd:
But it already does `pwstr.to_string()` right in `src/host/wasapi/device.rs` line 332?

I don't have a Windows machine myself, so am trying to understand.

xephyris:
So right now I'm having it return the `Device::id()` as a `String` guid for Windows and if you iterate with `device_by_id(id)` to get the device, it will work fine.

The snippet about `PWSTR` that I included is mainly for if you want to access a Windows WASAPI audio device yourself using the `id` provided by `cpal`. 

That code is more about external access using the `id`s from `cpal`, not if you are only using `cpal` to access and adjust audio devices.

src/lib.rs
```diff
@@ -272,6 +272,7 @@ impl std::str::FromStr for DeviceId {
             "webaudio" => Ok(DeviceId::WebAudio(data.to_string())),
             "emscripten" => Ok(DeviceId::Emscripten(data.to_string())),
             "ios" => Ok(DeviceId::IOS(data.to_string())),
+            "null" => Ok(DeviceId::Null),
             _ => Err(DeviceIdError::UnsupportedPlatform),
```
roderickvd: Let's remove the `_` catch-all. This will force us updating this method every time we add or remove a host.

xephyris: Unfortunately, I can't find a way to remove the `_` catch-all. This is an `&str` match so it gives me a non-exhaustive error if I do try to remove the `_` catch-all.

Here is the error below 
```
non-exhaustive patterns: `&_` not covered
the matched value is of type `&str`
`&str` cannot be matched exhaustively, so a wildcard `_` is necessary rustc[Click for full compiler diagnostic]
lib.rs(275, 41): ensure that all possible cases are being handled by adding a match arm with a wildcard pattern or an explicit pattern as shown: `,
            &_ => todo!()`
```

roderickvd: Ah yes. Let’s indeed make it a todo.

xephyris: So is there anything I should do at this point? Or should I mark this as resolved?

roderickvd: I think it’d be good to have: `&_ => todo!(“implement DeviceId::FromStr for {platform}”)`

xephyris: Done!

src/lib.rs
```diff
@@ -220,6 +220,64 @@ where
...
+            DeviceId::Null => write!(f, "null"),
+        }
+    }
+}
+
+impl std::str::FromStr for DeviceId {
+    type Err = DeviceIdError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let (platform, data) = s.split_once(':').ok_or(DeviceIdError::ParseError)?;
```
Klemen2: The null will only get parsed if it has ':' in the string so just "null" will be ParseError. The best way would probably be to add an extra if above for it since it's the only exception.

src/lib.rs
```diff
@@ -220,6 +220,64 @@ where
 /// one frame contains two samples (left and right channels).
 pub type FrameCount = u32;
 
+/// The device ID of the audio device, on supported OSs
+/// Currently only supports macOS and Windows (WASAPI)
+#[derive(Clone, Debug, Eq, PartialEq)]
```
Klemen2: Were DeviceIdError and DeviceId derives switched? I don't really see a reason why would an error type implement a Hash, but DeviceId wouldn't

src/lib.rs
xephyris: I have adjusted the format of `null` to `null:null` instead as that seems to have better consistency than adding an if for this specific scenario. This way, the match statement also covers all cases without requiring a specific separate case for `null`

src/lib.rs
xephyris: I originally based my implementation off of `DeviceName` and `DeviceNameError`, so this slipped through without me noticing, they are now switched. Thanks for catching this!

src/host/coreaudio/macos/device.rs
```diff
@@ -44,6 +48,9 @@ use std::time::{Duration, Instant};
 
 use super::property_listener::AudioObjectPropertyListener;
 use coreaudio::audio_unit::macos_helpers::get_device_name;
+
+type CFStringRef = *mut std::os::raw::c_void;
```
roderickvd: Could we remove this in favor of plain `CFString` and be less unsafe?

```rust
  fn id(&self) -> Result<DeviceId, DeviceIdError> {
      let property_address = AudioObjectPropertyAddress {
          mSelector: kAudioDevicePropertyDeviceUID,
          mScope: kAudioObjectPropertyScopeGlobal,
          mElement: kAudioObjectPropertyElementMain,
      };

      // CFString is retained by the audio object, use wrap_under_get_rule
      let mut uid: *const CFString = std::ptr::null();
      let data_size = size_of::<*const CFString>() as u32;

      // SAFETY: AudioObjectGetPropertyData is documented to write a CFString pointer
      // for kAudioDevicePropertyDeviceUID. We check the status code before use.
      let status = unsafe {
          AudioObjectGetPropertyData(
              self.audio_device_id,
              NonNull::from(&property_address),
              0,
              null(),
              NonNull::from(&data_size),
              NonNull::from(&mut uid).cast(),
          )
      };

      check_os_status(status)?;

      if uid.is_null() {
          return Err(DeviceIdError::BackendSpecific {
              err: BackendSpecificError {
                  description: "Device UID is null".to_string(),
              },
          });
      }

      // SAFETY: We verified uid is non-null and the status was successful
      let uid_string = unsafe { CFString::wrap_under_get_rule(uid).to_string() };
      Ok(DeviceId::CoreAudio(uid_string))
  }
```

src/host/wasapi/device.rs
```diff
@@ -322,6 +326,18 @@ impl Device {
         }
     }
 
+    fn id(&self) -> Result<DeviceId, DeviceIdError> {
+        unsafe {
+            match self.device.GetId() {
+                Ok(pwstr) => match pwstr.to_string() {
```
roderickvd, line 332: Instead of silently discarding, we could:

```rust
  fn id(&self) -> Result<DeviceId, DeviceIdError> {
      // SAFETY: Calling GetId on a valid IMMDevice is safe
      unsafe {
          let pwstr = self.device.GetId().map_err(|e| DeviceIdError::BackendSpecific {
              err: e.into(),
          })?;

          let id_str = pwstr.to_string().map_err(|e| DeviceIdError::BackendSpecific {
              err: BackendSpecificError { description: format!("Failed to convert device ID to string: {}", e) },
          })?;

          Ok(DeviceId::WASAPI(id_str))
      }
  }
```

src/host/coreaudio/ios/mod.rs
```diff
@@ -88,6 +89,10 @@ impl Device {
         Ok("Default Device".to_owned())
     }
 
+    fn id(&self) -> Result<DeviceId, DeviceIdError> {
+        Ok(DeviceId::IOS("default".to_string()))
```
roderickvd, line 75: "default" differs from "Default Device" in `name`. Maybe use shared const?

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 # Unreleased
 
+- Add `HostTrait::id` method that returns a stable audio device ID.
```
roderickvd: Should be `DeviceTrait`.

roderickvd (commented):
Thanks for sticking through 👍 this is a pretty important feature so let's get it right!

src/host/coreaudio/ios/mod.rs
xephyris: I'm slightly unsure about this change, as in most conventions an id would not have spaces inside of it, so I feel like `"default"` does a better job of this. `DeviceId`s and `name`s also do not intermix so I don't believe this will be much of an issue either.

Of course if the general consensus is that `"Default Device"` is better than `"default"`, I would be happy to change it.

roderickvd: Good rationale, let’s keep it like it this then. 

How do you feel this’ll be for ALSA? Spaces in device names (e.g. for USB devices) would not be impossible.

xephyris: I think if we are to maintain no spaces in the `id`, then we could maybe just replace all the spaces with `-` dashes or `_` underscores. 

From what I've seen and researched though, it seems like the `hw:CARD="CARD_ID",DEV="DEVICE_NUM"` style in `self.pcm_id` cannot contain spaces, as they are not permitted in card ids. Do you have any examples of such a case with spaces, or does the `alsa` implementation in `cpal` use card names not card ids?

My research in this is very basic, so what I'm saying may not be completely accurate.

src/host/coreaudio/macos/device.rs
```diff
@@ -411,11 +411,11 @@ impl Device {
             mElement: kAudioObjectPropertyElementMain,
         };
 
-        // CFString is retained by the audio object, use wrap_under_get_rule    
+        // CFString is retained by the audio object, use wrap_under_get_rule
```
yury: Quote from docs: 

>A CFString that contains a persistent identifier for the AudioDevice. An
AudioDevice's UID is persistent across boots. The content of the UID string
is a black box and may contain information that is unique to a particular
instance of an AudioDevice's hardware or unique to the CPU. Therefore they
are not suitable for passing between CPUs or for identifying similar models
of hardware. **The caller is responsible for releasing the returned CFObject.**

yury: @xephyris it is opposite. Even if uid is retained by device it will create copy when you get it.

src/host/coreaudio/ios/mod.rs
roderickvd: It's possible that USB DACs show up with spaces, e.g. "Schiit Unison Modi Multi 2".

src/host/coreaudio/macos/device.rs
xephyris: Thanks @yury.  I have fixed the issue and replaced `wrap_under_get_rule` with `wrap_under_create_rule`. Would you mind sharing a link to wherever you found that information? I can't seem these docs online from my searches.

yury: it is doc comments in 'CoreAudio/AudioHardwareBase.h' header.
Screenshot 2025-10-14 at 6 56 58 AM

Also wrote [a test](https://github.com/yury/cidre/blob/0571d6e61e59d69b80fc37a325cb02af180cad9b/cidre/src/core_audio/hardware.rs#L1923-L1932) to make sure pointers are unique

src/host/coreaudio/ios/mod.rs
xephyris: I don't have a USB DAC I can test with, but asking Copilot, this is what it said the card id would be for a USB DAC 
`/proc/asound/cards`
```
 1 [USB            ]: USB-Audio - USB Audio DAC
                      USB Audio DAC at usb-0000:00:1d.0-1, full speed
```

It seems like `alsa` just trims the device name for the first word. If this is the case, we could also do something similar, but one issue that might occur is that if two devices with the same starting word are plugged in, `alsa` will just assign ids based on the first device detected.

Do you have any suggestions as to resolve this?

roderickvd: I asked on Discord and here are some interesting data points: https://discord.com/channels/590254806208217089/672897096826748948/1427737852271136778

> It looks like they do the equivalent of 
> `name.split(" ").last()`

So that'd mean that there'd be no spaces in the PCM ID, indeed.

roderickvd:
Thanks, I’ll review later as I’m away for a few days.

xephyris:
Hello @roderickvd, 

Just wondering if you were available to review the updates, as I was hoping to get the `device_ids` integrated as soon as possible.

If it comes at an inconvenient time, I'm happy to wait.

src/lib.rs
```diff
@@ -223,6 +223,65 @@ where
 /// one frame contains two samples (left and right channels).
 pub type FrameCount = u32;
 
+/// The device ID of the audio device, on supported OSs
+/// Currently only supports macOS and Windows (WASAPI)
```
roderickvd: This is now expanded to other hosts as well, and could read:

```rust
/// A stable identifier for an audio device across all supported platforms.
///
/// Device IDs remain stable across application restarts and can be serialized using `Display`/`FromStr`.
```

src/host/coreaudio/macos/device.rs
```diff
@@ -355,6 +364,44 @@ impl Device {
...
+                NonNull::from(&property_address),
+                0,
+                null(),
+                NonNull::from(&data_size),
+                NonNull::from(&mut uid).cast(),
+            )
+        };
+        check_os_status(status)?;
+
+        // SAFETY: We verified uid is non-null and the status was successful
```
roderickvd: Slightly more accurate:

```rust
// SAFETY: Status was successful, meaning the API call succeeded.
// We now check if the returned uid is non-null before using it.
```

src/lib.rs
```diff
@@ -223,6 +223,65 @@ where
...
+            "aaudio" => {
+                let id = data.parse().map_err(|_| DeviceIdError::ParseError)?;
+                Ok(DeviceId::AAudio(id))
+            }
+            "jack" => Ok(DeviceId::Jack(data.to_string())),
+            "webaudio" => Ok(DeviceId::WebAudio(data.to_string())),
+            "emscripten" => Ok(DeviceId::Emscripten(data.to_string())),
+            "ios" => Ok(DeviceId::IOS(data.to_string())),
+            "null" => Ok(DeviceId::Null),
+            &_ => todo!("implement DeviceId::FromStr for {platform}"),
```
roderickvd, line 291: I know we discussed this earlier but I find myself doubting here.
Should we panic here or return `DeviceIdError::UnsupportedPlatform`?

roderickvd (changes requested):
Thanks for the ping and sorry for the wait. Yes lots of work done, thanks for that. Here's a few hopefully small ones.

src/lib.rs
xephyris: I think a `DeviceIdError::UnsuportedPlatform` would be better in this case as it would leave more flexibility for the end user, if they want their application to panic if the API is unsupported that option is allowed, but we also give them the option to accept it as an error instead of crashing their app.

roderickvd:
Thanks for all the work and updating with the custom host support. I wonder if we can support custom hosts better in the future, but that can be added later if necessary.

Eager to get this out with v0.17 and see the community response!

Issue #902, opened 2024-07-27
Android won't adhere to fixed buffer size


zhpixel517:
Hi all. I've created this pretty simple program that just listens to audio from the microphone and plays it back out to the speakers. All it does is record a sample, pass it to a ring buffer, and the output callback will copy that to the output. 

Take a look:

```rust
pub fn loopback_audio() {
    const BUFFER_SIZE: usize = 480;
    const RINGBUFFER_SIZE: usize = BUFFER_SIZE * 2;

    let (mut prod, mut cons) = RingBuffer::<f32>::new(RINGBUFFER_SIZE);
    let host = cpal::default_host();

    let input_device = host.default_input_device().unwrap();
    let output_device = host.default_output_device().unwrap();

    let input_config = cpal::StreamConfig {
        channels: 1,
        sample_rate: cpal::SampleRate(48000),
        buffer_size: cpal::BufferSize::Default
    };
    let output_config: cpal::StreamConfig = input_config.clone();

    let input_stream = input_device
        .build_input_stream(
            &input_config,
            move |data: &[f32], _: &_| {
                if let Ok(chunk) = prod.write_chunk_uninit(BUFFER_SIZE) {
                    chunk.fill_from_iter(data.to_owned()); // push to ring buffer
                }
            },
            move |err| {
                eprintln!("There was an input error: {:?}", err);
            },
            None,
        )
        .unwrap();

    let output_stream = output_device
        .build_output_stream(
            &output_config,
            move |data: &mut [f32], _: &cpal::OutputCallbackInfo| {
                debug!("Entered callback");
                let available = cons.slots();
                let to_read = data.len().min(available);

                match cons.read_chunk(to_read) {
                    Ok(chunk) => {
                        debug!("Read a chunk from the ringbuffer");
                        let (first, second) = chunk.as_slices(); //sort of like first half, second half, because rb is circular
                        let mid = first.len();  
                        data[..mid].copy_from_slice(first);
                        if second.len() != 0 {
                            data[mid..].copy_from_slice(second);
                        }
                        debug!("Copied all");
                        chunk.commit_all();
                        debug!("Committed all");
                    }
                    Err(error) => debug!("{} ", error),
                }
            },
            move |err| {
                debug!("There was an output error: {:?}", err);
            },
            None,
        )
        .unwrap();

    input_stream.play().unwrap();
    output_stream.play().unwrap();

    std::thread::sleep(Duration::from_secs(150));
}
```
[...]

PR #1010, opened 2025-09-11
feat!: add StreamConfigBuilder with cross-platform configuration support

roderickvd:
# Note This PR is stale

_I am no longer pursuing this PR, opting for a better API. I'm keeping this PR around for now only as a reminder of the various bells and whistles we can expose with some better API: #1334 _

---

This PR introduces an `StreamConfigBuilder` with support for platform-specific audio configuration options. This addresses recurring requests for fine-grained control over backend-specific settings while maintaining cross-platform compatibility. At the same time, it improves ergonomics of stream creation.

# Motivation

There is a recurring interest in exposing platform-specific audio configuration options:

- **#990** - ALSA buffer size configuration issues requiring `set_buffer_size_near`
- **#995** - Android `input_preset` configuration for AEC support
- **#987** - WASAPI raw audio stream configuration needs
- **#917** - ALSA period/buffer size configuration problems

Users have consistently needed platform-specific control over:
- **ALSA**: Period counts, access types (RW vs MMap; interleaved vs. non-interleaved)
- **JACK**: Client names, automatic port connections, server startup
- **WASAPI**: Exclusive vs shared mode

Not yet added:
- **Android**: Input presets for AEC and noise cancellation

Currently, these options are either impossible to configure or would require hard-coded defaults or introduction of environment variables.

# Solution

All existing APIs remain unchanged. The new builder and platform configuration are purely additive:

```rust
use cpal::{StreamConfigBuilder, SampleRate, SampleFormat, BufferSize};

// - builds on all platforms
// - offers more ergonomic stream creation
// - allows platform-specific options
let stream = device.default_output_config()?
    .with_buffer_size(BufferSize::Fixed(256))
    .on_alsa(|alsa| alsa.periods(2))                  // No-op on non-Linux
    .on_wasapi(|wasapi| wasapi.exclusive_mode(true))  // No-op on non-Windows  
    .on_jack(|jack| jack.client_name("app"))          // No-op without JACK
    .build_output_stream(/* callbacks */)?;
```

As opportunistic refactoring, the following feature flags were re-added:

- `jack` - Enable JACK audio backend support
- `audio_thread_priority` - Enable real-time thread priority for ALSA/WASAPI

And documentation was updated for readability and correctness.

PR #1015, opened 2025-09-24
AAudio AudioPerformanceMode::LowLatency

joe-wd:
Current cpal implementation results in noisy audio on Android. Please consider defaulting to AudioPerformanceMode::LowLatency.

I forked and made this change as suggested in https://github.com/RustAudio/cpal/issues/902, and it resolved the noise issues. This should be the default.

### DIFF
--- a/src/host/aaudio/mod.rs
+++ b/src/host/aaudio/mod.rs
@@ -406,7 +406,8 @@ impl DeviceTrait for Device {
         let builder = ndk::audio::AudioStreamBuilder::new()?
             .direction(ndk::audio::AudioDirection::Input)
             .channel_count(channel_count)
-            .format(format);
+            .format(format)
+            .performance_mode(ndk::audio::AudioPerformanceMode::LowLatency);
 
         build_input_stream(
             self,
@@ -455,7 +456,8 @@ impl DeviceTrait for Device {
         let builder = ndk::audio::AudioStreamBuilder::new()?
             .direction(ndk::audio::AudioDirection::Output)
             .channel_count(channel_count)
-            .format(format);
+            .format(format)
+            .performance_mode(ndk::audio::AudioPerformanceMode::LowLatency);
 
         build_output_stream(
             self,


roderickvd:
Thank you for the PR! I'd be curious to hear more details about the "noisy audio" issue you're experiencing. What does it sound like exactly? And which buffer size and sample rate are you using?

I understand that low latency mode trades battery efficiency for reduced audio delay. While your fix may work as a workaround, I suspect there might be an underlying issue that should work properly even in default performance mode. We could explore adding low latency as a configurable option in #1010 rather than making it the default.

A few potential root causes I'm considering:
- The data callback makes system calls for both `callback` and `capture` timestamps, which aren't real-time safe
- Buffer size variations could occur with certain sample rates like 44100 Hz
- We don't appear to be using `frames_per_data_callback()`, which the [docs](https://developer.android.com/ndk/reference/group/audio#aaudiostreambuilder_setframesperdatacallback)
indicates is necessary for consistent callback buffer sizes: _"Note that numFrames can vary unless 
AAudioStreamBuilder_setFramesPerDataCallback() is called."_

Could you test an alternative fix? In `src/host/aaudio/mod.rs`, try modifying the `configure_for_device` function around line 218:

```rust
match &config.buffer_size {
    BufferSize::Default => builder,
    BufferSize::Fixed(size) => builder
        .frames_per_data_callback(*size as i32)
        .buffer_capacity_in_frames((*size * 2) as i32), // Double-buffering
}
```

If this resolves the noise problem, we could consider implementing low latency and other performance modes as an opt-in feature instead of the default behavior in #1010.

joe-wd:
Thank you for responding and considering the issue! I tested your suggested change and it does make a significant difference -- I'd say it gets us about 90% of the way there. We will adopt it this change as well.

Our general use case is high-end audio playback, sample rate 48000, 2-channel (stereo). The majority of the time we're simply copying directly from a ring buffer to the output buffer. When transitioning from one track to the next, we're performing a Hann window crossfade, so we're copying from two different ring buffers, applying gain, and mixing into the final buffer. All of that should be handled easily by any modern Android device. I verified that the ring buffers keep up (our reads from the ring are never waiting for data).

Without your suggested change, we found that the callback buffer sizes on Android are all over the place, ranging from 12 (really!) to 4000+, and suspected that this was the main problem. Requesting a fixed buffer size made no difference. With your change, the buffer sizes are consistent and as requested. 

Regarding noise, our QA and users reported popping and crackly output, which suggests that we were missing the output buffer deadline, probably when the buffer size is very small. One difficulty here is that the callback passes an `OutputCallbackInfo` with the `OutputStreamTimestamp`:

```
pub struct OutputStreamTimestamp {
    /// The instant the stream's data callback was invoked.
    pub callback: StreamInstant,
    /// The predicted instant that data written will be delivered to the device for playback.
    ///
    /// E.g. The instant data will be played by a DAC.
    pub playback: StreamInstant,
}
```

On Android, the `callback` instant is not a reasonable value (it is some value far in the past), so we can't confidently know the real deadline. But I assume we were missing it.

Again, thanks for replying and the suggested change. For our work, we want to keep AudioPerformanceMode::LowLatency as well, so I hope you'll consider it at least as an opt-in.

roderickvd:
> Thank you for responding and considering the issue! I tested your suggested change and it does make a significant difference -- I'd say it gets us about 90% of the way there. We will adopt it this change as well.

Great, thanks for the feedback. Let's extract it into a separate PR.

> I verified that the ring buffers keep up (our reads from the ring are never waiting for data).

That's a clear data point. So there's no zero padding of the input / output buffers then that could cause audible glitches.

> Without your suggested change, we found that the callback buffer sizes on Android are all over the place, ranging from 12 (really!) to 4000+, and suspected that this was the main problem. Requesting a fixed buffer size made no difference. With your change, the buffer sizes are consistent and as requested.

What's the buffer size you're setting? 
Does noise become better / worse when you request higher / lower buffer sizes?
Finally, as a quick test, what happens if you do `buffer_capacity_in_frames((*size * 4) as i32)` instead of `* 2`?

> One difficulty here is that the callback passes an `OutputCallbackInfo` with the `OutputStreamTimestamp`:
> 
> On Android, the `callback` instant is not a reasonable value (it is some value far in the past), so we can't confidently know the real deadline. But I assume we were missing it.

The origin of `StreamInstant` is not guaranteed to be related to UNIX epoch time or something like it. In this case, the `callback` stream instant is the duration that has elapsed since its creation.

roderickvd:
Superseded by https://github.com/RustAudio/cpal/pull/1025

roderickvd:
With #1187. low latency mode will be set when the `realtime` feature is enabled. In the future, I'd also like to support power-saving mode using trait extensions or something.

PR #1021, opened 2025-09-27
feat: make Stream implement Send on macOS

roderickvd:
`Stream` can now be moved between threads safely on macOS.

- Replaced `Rc<RefCell<>>` with `Arc<Mutex<>>` for thread-safe access to stream state
- Extracted device disconnection handling into a separate `DisconnectManager` to isolate non-`Send` `AudioObjectPropertyListener`
- Removed unnecessary `unsafe impl Send` implementations that are now automatically derived

The mutex is only locked during `play()` and `pause()` operations, so performance impact is negligible. The audio callback is completely unaffected.

### DIFF
--- a/src/host/coreaudio/macos/mod.rs
+++ b/src/host/coreaudio/macos/mod.rs
@@ -7,11 +7,14 @@ use crate::traits::{HostTrait, StreamTrait};
 use crate::{BackendSpecificError, DevicesError, PauseStreamError, PlayStreamError};
 use coreaudio::audio_unit::AudioUnit;
 use objc2_core_audio::AudioDeviceID;
-use std::cell::RefCell;
-use std::rc::Rc;
+use std::sync::{Arc, Mutex, Weak};
 
 pub use self::enumerate::{default_input_device, default_output_device, Devices};
 
+use objc2_core_audio::{
+    kAudioDevicePropertyDeviceIsAlive, kAudioObjectPropertyElementMain,
+    kAudioObjectPropertyScopeGlobal, AudioObjectPropertyAddress,
+};
 use property_listener::AudioObjectPropertyListener;
 
 mod device;
@@ -52,11 +55,85 @@ impl HostTrait for Host {
     }
 }
 
+/// Type alias for the error callback to reduce complexity
+type ErrorCallback = Box<dyn FnMut(crate::StreamError) + Send + 'static>;
+
+/// Manages device disconnection listener separately from StreamInner to allow Stream to be Send.
+/// The AudioObjectPropertyListener contains non-Send closures, so we keep it separate and
+/// manage its lifetime independently.
+struct DisconnectManager {
+    _listener: AudioObjectPropertyListener,
+}
+
+impl DisconnectManager {
+    /// Create a new DisconnectManager that monitors device disconnection
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure:
+    /// - `device_id` is a valid AudioDeviceID
+    /// - The error callback is Send + 'static
+    fn new(
+        device_id: AudioDeviceID,
+        stream_weak: Weak<Mutex<StreamInner>>,
+        error_callback: Arc<Mutex<ErrorCallback>>,
+    ) -> Result<Arc<Self>, crate::BuildStreamError> {
+        let property_address = AudioObjectPropertyAddress {
+            mSelector: kAudioDevicePropertyDeviceIsAlive,
+            mScope: kAudioObjectPropertyScopeGlobal,
+            mElement: kAudioObjectPropertyElementMain,
+        };
+
+        let callback = {
+            let stream_weak = stream_weak.clone();
+            let error_callback = error_callback.clone();
+            move || {
+                // Check if stream still exists
+                if let Some(stream_arc) = stream_weak.upgrade() {
+                    // First, try to pause the stream to stop playback
+                    match stream_arc.lock() {
+                        Ok(mut stream_inner) => {
+                            let _ = stream_inner.pause();
+                        }
+                        Err(_) => {
+                            // Could not acquire lock. This can occur if there are
+                            // overlapping locks, if the stream is already in use, or if a panic
+                            // occurred during a previous lock. Still notify about device
+                            // disconnection even if we can't pause.
+                        }
+                    }
+
+                    // Call the error callback to notify about device disconnection
+                    if let Ok(mut cb) = error_callback.lock() {
+                        cb(crate::StreamError::DeviceNotAvailable);
+                    }
+                }
+            }
+        };
+
+        let listener = AudioObjectPropertyListener::new(device_id, property_address, callback)
+            .map_err(|e| {
+                let description = format!("Failed to create disconnect listener: {e}");
+                crate::BuildStreamError::BackendSpecific {
+                    err: crate::BackendSpecificError { description },
+                }
+            })?;
+
+        Ok(Arc::new(DisconnectManager {
+            _listener: listener,
+        }))
+    }
+}
+
+// SAFETY: DisconnectManager is only used to keep the AudioObjectPropertyListener alive
+// and is never actually accessed across threads. The AudioObjectPropertyListener
+// itself handles thread safety for the CoreAudio callback internally.
+unsafe impl Send for DisconnectManager {}
+unsafe impl Sync for DisconnectManager {}
+
 struct StreamInner {
     playing: bool,
     audio_unit: AudioUnit,
-    /// Manage the lifetime of the closure that handles device disconnection.
-    _disconnect_listener: Option<AudioObjectPropertyListener>,
     // Track the device with which the audio unit was spawned.
     //
     // We must do this so that we can avoid changing the device sample rate if there is already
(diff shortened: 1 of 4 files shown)


src/host/coreaudio/macos/mod.rs
```diff
@@ -52,11 +55,85 @@ impl HostTrait for Host {
...
+}
+
+impl DisconnectManager {
+    /// Create a new DisconnectManager that monitors device disconnection
+    ///
+    /// # Safety
+    ///
+    /// The caller must ensure:
+    /// - `device_id` is a valid AudioDeviceID
+    /// - The error callback is Send + 'static
```
yara-blue: Doesnt the compiler do that? Given the bounds on ErrorCallback are Send + 'static

src/host/coreaudio/macos/mod.rs
```diff
@@ -96,33 +173,98 @@ impl StreamInner {
...
 
 impl StreamTrait for Stream {
     fn play(&self) -> Result<(), PlayStreamError> {
-        let mut stream = self.inner.borrow_mut();
+        let mut stream = self
+            .inner
+            .lock()
+            .map_err(|_| PlayStreamError::BackendSpecific {
+                err: BackendSpecificError {
+                    description: "Failed to acquire stream lock".to_string(),
```
yara-blue: Assuming std::sync::Mutex we can only fail to lock because a thread holding this code panicked. That should never happen: lets document that in the error description. Something along the lines of `A cpal operation holding this stream panicked, this is a bug`.

yara-blue: Same for pause

src/host/coreaudio/macos/mod.rs
```diff
@@ -96,33 +173,98 @@ impl StreamInner {
 
 #[derive(Clone)]
 pub struct Stream {
-    inner: Rc<RefCell<StreamInner>>,
+    inner: Arc<Mutex<StreamInner>>,
+    // Manages the device disconnection listener separately to allow Stream to be Send.
+    // The DisconnectManager contains the non-Send AudioObjectPropertyListener.
+    _disconnect_manager: Arc<DisconnectManager>,
```
yara-blue: If this `Stream` drops after being moved to another thread then `_disconnect_manager` will be dropped on another thread too. Is it safe to drop `DisconnectManager` from another thread?

yara-blue: I guess that depends if its safe to drop AudioObjectPropertyListener from another thread

yara-blue: It does do things on drop: https://github.com/RustAudio/cpal/blob/87b7036a85d180e075dbd81720eb5818ccbabf7b/src/host/coreaudio/macos/property_listener.rs#L70

yara-blue: Reviewing this has reminded me how much I hate Apple's documentation.

yara-blue: Anyway I strongly suspect this can not be dropped safely from another thread. In which case we can not make `Stream` implement `Send`.

src/host/coreaudio/macos/mod.rs
roderickvd: Yep, removed in [3b25497](https://github.com/RustAudio/cpal/pull/1021/commits/3b254971cb330ca4d99f033d96ca566528bce485).

src/host/coreaudio/macos/mod.rs
roderickvd: > Anyway I strongly suspect this can not be dropped safely from another thread. In which case we can not make `Stream` implement `Send`.

Moved to dedicated thread in [3b25497](https://github.com/RustAudio/cpal/pull/1021/commits/3b254971cb330ca4d99f033d96ca566528bce485).

> Reviewing this has reminded me how much I hate Apple's documentation.

Yes, it is bizarre.

roderickvd:
@dvdsk good catch. Let's be more conservative.  I've reworked this to use a dedicated thread.

While I was at it, I also killed the `Clone` impl on `Stream`. That's just confusing, and ALSA and WASAPI don't have it either. Should be much cleaner now.

src/host/coreaudio/macos/mod.rs
```diff
@@ -52,11 +55,86 @@ impl HostTrait for Host {
...
+                // Check if stream still exists
+                if let Some(stream_arc) = stream_weak_clone.upgrade() {
+                    // First, try to pause the stream to stop playback
+                    match stream_arc.lock() {
+                        Ok(mut stream_inner) => {
+                            let _ = stream_inner.pause();
+                        }
+                        Err(_) => {
+                            // Could not acquire lock. This can occur if there are
+                            // overlapping locks, if the stream is already in use, or if a panic
```
yara-blue: this sounds like this could deadlock? Should we do `try_lock` instead then?

yara-blue: looking at the other locks it should not.

src/host/coreaudio/macos/mod.rs
```diff
@@ -52,11 +55,86 @@ impl HostTrait for Host {
     }
 }
 
+/// Type alias for the error callback to reduce complexity
+type ErrorCallback = Box<dyn FnMut(crate::StreamError) + Send + 'static>;
+
+/// Manages device disconnection listener on a dedicated thread to ensure the
+/// AudioObjectPropertyListener is always created and dropped on the same thread.
+/// This avoids potential threading issues with CoreAudio APIs.
```
yara-blue, line 87: so the purpose of this struct is to pause the stream on disconnect?

src/host/coreaudio/macos/mod.rs
roderickvd: This was a semi-correct (so: incorrect 😆) carry-over from the previous code. In 77ba291 I've added some defensive coding to deal with:
- callbacks already being invoked (no full reentrancy)
- the lock being poisoned

src/host/coreaudio/macos/mod.rs
roderickvd: That and to propagate a `DeviceNotAvailable` error. Clarified in 77ba291.

roderickvd:
This should be in a good state now - thanks @dvdsk for the necessary reviews. I'll rebase and if you agree, we can merge.

edwloef:
It would be nice if various backends could be tested on their send-ness, to make sure this doesn't accidentally regress. Something on the lines of
```rs
const fn assert_send<T: Send>() {}
const _: () = assert_send::<Stream>();
```
maybe, so that a non-send stream is a compile fail?

roderickvd:
> It would be nice if various backends could be tested on their send-ness, to make sure this doesn't accidentally regress. Something on the lines of
> 
> ```rust
> const fn assert_send<T: Send>() {}
> const _: () = assert_send::<Stream>();
> ```

Adding that in [1d117d3](https://github.com/RustAudio/cpal/pull/1021/commits/1d117d32caf414798f13ec94c96f01f4cf3aacae) was pretty revealing: WASAPI and WASM streams aren't `Send`, either. I'm gonna see if that's easy enough to fix to include with this PR.

edwloef:
> > It would be nice if various backends could be tested on their send-ness, to make sure this doesn't accidentally regress. Something on the lines of
> > 
> > ```rust
> > const fn assert_send<T: Send>() {}
> > const _: () = assert_send::<Stream>();
> > ```
> 
> Adding that in [1d117d3](https://github.com/RustAudio/cpal/pull/1021/commits/1d117d32caf414798f13ec94c96f01f4cf3aacae) was pretty revealing: WASAPI and WASM streams aren't `Send`, either. I'm gonna see if that's easy enough to fix to include with this PR.

On WASM this is probably fine since there's no real concept of threading as far as I'm aware, but the WASAPI stream being `Send` would be nice.

roderickvd:
> On WASM this is probably fine since there's no real concept of threading as far as I'm aware, but the WASAPI stream being `Send` would be nice.

You're right. Turned out to be easy for both, so added it here.

roderickvd:
Merging this now because I want to start preparing the next release. Please create an issue if something's up with it.

ghenry:
Hi all,

What's the best way to use this on macOS as I've just hit the following:

```
Note: required for `std::sync::Arc<std::sync::Mutex<rodio::cpal::host::coreaudio::macos::StreamInner>>` to implement `Sync`
```

as I do:

https://codeberg.org/OneTalker/OneTalker/src/branch/main/src/main.rs#L203


```rust
async fn play_sound(sound: String) -> Result<(Arc<Sink>, Arc<OutputStream>), Error> {
    // Get an output stream handle to the default physical sound device.
    // Note that the playback stops when the stream_handle is dropped.
    let stream_handle = rodio::OutputStreamBuilder::open_default_stream().map_err(|e| {
        Error::PlaybackError(format!("Failed to open default audio output stream: {e}"))
    })?;

    // Load a sound from a file, using a path relative to Cargo.toml
    let file = BufReader::new(
        File::open(audio_path().join(sound))
            .map_err(|e| Error::PlaybackError(format!("Failed to open audio file: {e}")))?,
    );
    // Note that the playback stops when the sink is dropped
    let sink = rodio::play(stream_handle.mixer(), file)
        .map_err(|e| Error::PlaybackError(format!("Failed to play audio file: {e}")))?;

    Ok((Arc::new(sink), Arc::new(stream_handle)))
}
```

which is fine on Linux :-)

Thanks!

roderickvd:
@ghenry I see that you're pulling in Rodio 0.16 through Rodio 0.21. Until Rodio 0.22 is released, you need Rodio master to pull in cpal 0.17.

ghenry:
Cool. Will try that tomorrow. Thanks!

ghenry:
Working fine on Linux - [changes](https://codeberg.org/OneTalker/OneTalker/commit/90bc7a88f8dc82cb484635af7b5f118bfeab7ca8). Test mac tomorrow. Only thing I had to figure out was using the `nz!` macro by searching for `SamplesBuffer::new` as I was:

```
    let buf = SamplesBuffer::new(1, 16000, samples); 
```

which is now:

```
    let buf = SamplesBuffer::new(nz!(1), nz!(16000), samples); 
```

Others may hit that too. I see that's in the CHANGELOG as:

```
Added a macro nz! which facilitates creating NonZero's for SampleRate and ChannelCount.
```

but might be worth highlighting the `SamplesBuffer::new()` bit in the [Changed](https://github.com/RustAudio/rodio/blob/master/CHANGELOG.md#changed) part. I'll do a PR for you.

PR #1013, opened 2025-09-21
refactor(alsa): optimize period configuration and audio callback performance

roderickvd:
Improves audio callback performance through device-aware buffering, pre-allocation and period-aligned writes. This results in more consistent latency control and improved hot path performance.

Also fixes a DC offset issue with unsigned formats by proper silence (equilibrium) handling.

## Period Configuration Strategy
Queries device capabilities to determine which approach to use:
1. **Double buffering**: Uses 2 periods when user buffer fits within device limits
2. **Multi-period**: Calculates optimal period distribution for larger buffers
3. **Fallback**: Uses buffer-size-only for maximum device compatibility

### Audio Quality Improvements
- **Proper silence handling**: Fills buffers with correct equilibrium values for unsigned sample formats instead of zeros (prevents DC offset and pops)
- **Full period writes**: Always writes complete periods to ALSA for consistent block-based processing
- **Proper stream draining**: Drain audio sink on drop when still playing, essential for common patterns like Rodio's `sink.sleep_until_end()`

## Performance & Real-Time Improvements
- **Zero allocations during playback**: Pre-allocate all buffers (`transfer_buffer`, `silence_template`, `descriptors`) at stream creation
- **Cache hot path values**: Store `period_frames` and `period_samples` in `StreamInner` to avoid repeated calculations
- **Consistent buffer sizes**: Period-aligned writes eliminate variable-size buffer handling overhead

Thanks to @abique for the [discussions](https://github.com/RustAudio/cpal/commit/ea415979b080362a73811b0d84ce28eee8b1364e#r165664795)

Supersedes:
- https://github.com/RustAudio/cpal/pull/990

PR #1022, opened 2025-09-27
refactor(coreaudio): optimize buffer configuration

roderickvd:
Configure CoreAudio device buffer to ensure applications receive callbacks with exactly the buffer size they requested. This provides the same predictable buffer behavior that PR #1013 proposes for ALSA.

## Behavior

- Before: Requesting 256 frames → receiving 512 frame callbacks, ~512 frame latency
- After: Requesting 256 frames → receiving 256 frame callbacks, ~256 frame latency

Latency is determined by the callback buffer size before transfer to the device hardware buffer. CoreAudio's buffer frame size corresponds to ALSA's "period" concept, with CoreAudio automatically implementing double buffering.


roderickvd:
Another part of the campaign on consistent buffer sizes. When done, I'm thinking of doing an alpha release and reaching out on Discord.

PR #1026, opened 2025-09-29
feat(ios): add AVAudioSession integration and refactor implementation

roderickvd:
Part of the campaign to implement fixed buffer sizes across all hosts.

- Implemented `BufferSize::Fixed` using `AVAudioSession`
- Query actual sample rates, channel counts, and buffer size ranges from `AVAudioSession`
- Consolidated input/output stream setup logic and extracted common buffer processing
- Added `#[inline]` to hot-path callback functions

Admission: this came from pure code review - I don't have iOS development hardware myself. @NightBlaze - since you're an iOS user, would you take this for a spin?

### DIFF
--- a/src/host/coreaudio/ios/mod.rs
+++ b/src/host/coreaudio/ios/mod.rs
@@ -52,7 +52,7 @@ impl HostTrait for Host {
     }
 
     fn devices(&self) -> Result<Self::Devices, DevicesError> {
-        Devices::new()
+        Ok(Devices::new())
     }
 
     fn default_input_device(&self) -> Option<Self::Device> {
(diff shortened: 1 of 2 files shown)


NightBlaze:
@roderickvd I think Copilot could handle this task better than I can. I've only worked with audio on iOS at a really basic level and I'm not familiar with a lot of the details.

roderickvd:
If you would, it’d already be a great help if you could confirm that it actually works 😄 without regressions.

src/host/coreaudio/ios/mod.rs
```diff
@@ -472,3 +325,250 @@ fn get_device_buffer_frame_size(audio_unit: &AudioUnit) -> Result<usize, coreaud
         Element::Output,
     )
 }
+
+/// Get the actual buffer size from AVAudioSession.
+///
+/// This queries the current IO buffer duration from AVAudioSession and converts
+/// it to frames based on the current sample rate.
+fn get_audio_session_buffer_size() -> Result<usize, coreaudio::Error> {
```
yury: Why Result? It seems never fail.

NightBlaze:
Tested on sim and real device, everything works as expected.

roderickvd:
> Tested on sim and real device, everything works as expected.

Thanks, that's super helpful. I'll process that other comment and will go for merge.

src/host/coreaudio/ios/mod.rs
roderickvd: You are right. I changed that here and in many other places in fd78285.

PR #754, opened 2023-02-24
WASAPI: Improve how default devices are created to support automatic stream routing

sidit77:
Use `ActivateAudioInterfaceAsync` to create default input/output devices to support [automatic stream routing](https://learn.microsoft.com/en-us/windows/win32/coreaudio/automatic-stream-routing).

Related issues: #740

Open questions:
- [ ] Fallback for windows versions <1607
- [ ] Implementation of device name/eq
- [ ] What happens when the is no default device (can this even happen?)

PR #1027, opened 2025-10-01
WASAPI: Use `ActivateAudioInterfaceAsync` with virtual device IDs for default devices

### LINKED ISSUE #740: Respect Windows output device selection
## Current behavior

I am using `rodio` to play audio on Windows 11. When I change the output device on Windows, the audio continues to play on the old device.

cpal

There is [a similar issue](https://github.com/RustAudio/rodio/issues/327) for macOS. But that is fixed in https://github.com/RustAudio/rodio/issues/327.

## The behavior I expected

On Windows, if user creates a device with `Device::default_output_device()` and plays audio through it, when user changes Windows's output device, the audio should play on the new device.

## Document:

https://learn.microsoft.com/en-us/windows/win32/coreaudio/relevant-device-notifications-for-stream-routing

https://learn.microsoft.com/en-us/windows/win32/coreaudio/device-events

1. Add `implement` feature to `windows` crate:

    ```toml
    [target.'cfg(target_os = "windows")'.dependencies]
    windows = { ..., features = [..., "implement"] }
    ```

2. Implement [IMMNotificationClient](https://learn.microsoft.com/en-us/windows/win32/api/mmdeviceapi/nn-mmdeviceapi-immnotificationclient):

    ```rust
    #[implement(windows::Win32::Media::Audio::IMMNotificationClient)]
    struct IMMNotificationClientWrapper {}

    impl windows::Win32::Media::Audio::IMMNotificationClient_Impl for IMMNotificationClientWrapper {
        ...

        fn OnDefaultDeviceChanged(&self, flow: EDataFlow, role: ERole, pwstrdefaultdeviceid: &PCWSTR, ) -> Result<(), windows::core::Error> {
            todo!("Handle default device changed event");
        }
    }
    ```

3. Register the `IMMNotificationClientWrapper` in `src\host\wasapi\device.rs`:
   
    ```rust
    ENUMERATOR.0.RegisterEndpointNotificationCallback(IMM_NOTIFICATION_CLIENT_WRAPPER);
    ```

philpax:
Fixes #740 (mostly). Updated version of #754 that aims to address the issues mentioned in that PR.

## Background

When you currently play some audio on the default device using `cpal`, the default device is fetched once, and then a stream is created from that device to play audio. However, changing the default device will lead to that stream continuing to play on the original device, and removing the device entirely (e.g. unplugging headphones) will result in the stream dying as it no longer has a device to output to.

I spent several hours trying to fix this properly - that is, trying to update the device and rebuild the stream when it changes - and stopped when I realised that a significant amount of the WASAPI stream code would need to be revised to handle a changing device. For those curious, you can see my [experimental work here](https://github.com/RustAudio/cpal/compare/master...philpax:cpal:audio-device-fix). The goal was to replicate the flow of this [Chromium code](https://chromium.googlesource.com/chromium/chromium/+/4cecfb2ac11410fc21f86d96802c9da6acd90bcd/media/audio/win/audio_low_latency_output_win.cc), which implements this logic.

## The Imperfect Fix

However, you don't actually need to do any of that if you don't care about supporting <Windows 8. Windows 8 introduced [`ActivateAudioInterfaceAsync`](https://learn.microsoft.com/en-us/windows/win32/api/mmdeviceapi/nf-mmdeviceapi-activateaudiointerfaceasync), and with it, [virtual device interfaces](https://learn.microsoft.com/en-us/windows/win32/coreaudio/devinterface-xxx-guids). `ActivateAudioInterfaceAsync` can be used instead of `Audio::IMMDevice::Activate` to generate a `Audio::IAudioClient` that will automatically reroute audio for you if the device changes.

This is what the previous PR did, and it works great. On top of that PR, this PR:
- brings it up to date
- pushes these changes into a default-on feature
- makes `name` use the current default device
- does some minor cleanup

## Ramifications

I've added a default-on feature, `wasapi-virtual-default-devices`, that adds support for this. Users that need to support older versions of Windows can turn this feature off, and it will behave as the current status quo does. (I'm not actually sure what happens if you have this feature on for an older Windows: in an ideal world, it does nothing, and newer Windows users can enjoy virtual devices. Please let me know if you've tested this.)

One downside of this is that I had to bump the minimum `windows` version to 0.61. This is because 0.61 removed [the previously-explicitly-required `implement` feature in 0.61](https://github.com/microsoft/windows-rs/pull/3333), and I can't see any way to conditionally include that feature for older versions of `windows`. `cargo` will refuse to resolve the `windows` dependency with a feature that doesn't exist for the version it's resolving to. I'm happy to drop it back down if anyone can think of a workaround for this.


src/host/wasapi/device.rs
```diff
@@ -266,15 +273,70 @@ unsafe fn format_from_waveformatex_ptr(
...
+    let (sender, receiver) = std::sync::mpsc::channel();
+    let completion: Audio::IActivateAudioInterfaceCompletionHandler =
+        CompletionHandler(sender).into();
+    Audio::ActivateAudioInterfaceAsync(
+        deviceinterfacepath,
+        &Audio::IAudioClient::IID,
+        None,
+        &completion,
+    )?;
+    let result = receiver.recv_timeout(Duration::from_secs(2)).unwrap()?;
```
roderickvd: Why 2 seconds?

How about having it return `Err`?

```rust
let result = receiver.recv_timeout(Duration::from_secs(2))
    .map_err(|_| windows::core::Error::new(
        Audio::AUDCLNT_E_DEVICE_INVALIDATED,
        "timeout waiting for audio interface activation"
    ))??;
```

src/host/wasapi/device.rs
```diff
@@ -343,10 +435,43 @@ impl Device {
...
+        #[cfg(feature = "wasapi-virtual-default-devices")]
         let audio_client: Audio::IAudioClient = unsafe {
-            // can fail if the device has been disconnected since we enumerated it, or if
-            // the device doesn't support playback for some reason
-            self.device.Activate(Com::CLSCTX_ALL, None)?
+            match &self.device {
+                DeviceType::DefaultOutput => {
+                    let default_audio = Com::StringFromIID(&Audio::DEVINTERFACE_AUDIO_RENDER)?;
+                    let result = activate_audio_interface_sync(default_audio);
+                    Com::CoTaskMemFree(Some(default_audio.as_ptr() as _));
```
roderickvd: Good that you're freeing before returning the error. An safer / more defensive pattern like with `IdRAII` could be:

```rust
struct ComString(windows::core::PWSTR);
impl Drop for ComString {
    fn drop(&mut self) {
        unsafe { Com::CoTaskMemFree(Some(self.0.as_ptr() as _)) }
    }
}

DeviceType::DefaultOutput => {
    let default_audio = ComString(Com::StringFromIID(&Audio::DEVINTERFACE_AUDIO_RENDER)?);
    activate_audio_interface_sync(default_audio.0)?
}
```

Cargo.toml
```diff
@@ -10,11 +10,20 @@ edition = "2021"
...
     "asio-sys",
     "num-traits",
 ] # Only available on Windows. See README for setup instructions.
 
+# Enable virtual default devices for WASAPI, so that audio will be
+# automatically rerouted when the default input or output device is changed.
+#
+# Note that this only works on Windows 8 and above. It is turned on by default,
+# but consider turning it off if you are supporting an older version of Windows.
+wasapi-virtual-default-devices = []
```
roderickvd: First, a documentation suggestion:

```toml
# Enable virtual default devices for WASAPI. When enabled:
# - Audio automatically reroutes when the default device changes
# - Streams survive device changes (e.g., plugging in headphones)
# - Requires Windows 8 or later
#
# Disable this feature if supporting Windows 7 or earlier.
```

Second, I wonder if we should invert the feature and rename it to `windows-legacy` (disabled by default). I'm not so sure if "virtual default devices" clearly communicates its intention. What do you think?

src/host/wasapi/device.rs
```diff
@@ -266,15 +273,70 @@ unsafe fn format_from_waveformatex_ptr(
...
+    result.cast()
+}
+
 unsafe impl Send for Device {}
 unsafe impl Sync for Device {}
 
 impl Device {
     pub fn name(&self) -> Result<String, DeviceNameError> {
+        let device = self
+            .immdevice()
```
roderickvd, line 418: How much of a performance hit does this take when enumerating? Would it make sense to cache it during enumeration or is that a negligible thing?

Cargo.toml
```diff
@@ -44,6 +58,8 @@ windows = { version = ">=0.58, <=0.62", features = [
     "Win32_Media_Multimedia",
     "Win32_UI_Shell_PropertiesSystem",
 ] }
+# Explicitly depend on windows-core for use with the `windows::core::implement` macro.
+windows-core = "*"
```
roderickvd: `cargo publish` will reject wildcard versions, please specify what we need.

roderickvd (changes requested):
Thank you! I don't really see a way around the `windows` crate version problem. How much of a problem that is, kind of depends on the community... don't know how much it'll hurt to bump the minimum to v0.61. If anyone feels strongly about that, do comment here.

src/host/wasapi/device.rs
philpax: 2 seconds was from the previous PR, but it's a good question; I'm not sure if there's a better value or strategy we could use here. Tempted to leave it as-is and comment that it's a failsafe timeout and that activation should happen "instantly" otherwise. (Not sure when it would fail: no device at all present, maybe?)

Returning `Err` is a good suggestion, will do!

src/host/wasapi/device.rs
philpax: `windows-rs` might have something for this, I'll dig around (I know it has `Owned` and `HSTRING`). Failing that, though, I'll get a `ComString` in.

Cargo.toml
philpax: Good shout on the docs, will do first.

I think inverting the feature would be more ergonomic, but [the Cargo docs](https://doc.rust-lang.org/cargo/reference/features.html#feature-unification) suggest that features should always be additive:

> A consequence of this is that features should be additive. That is, enabling a feature should not disable functionality, and it should usually be safe to enable any combination of features. A feature should not introduce a [SemVer-incompatible change](https://doc.rust-lang.org/cargo/reference/features.html#semver-compatibility).

I _think_ it's probably fine - I don't see a way for this to cause a conflict with another package, as you'd want this to be controlled at the application level anyway - but I'm not sure. For now, I'll just action the doc change, but I'm open to the inversion if you think it's compatible with the norms of the ecosystem.

Feature name: Hmm, yeah, I'm not sold on it either; it was what first came to mind to cover the concept of activating the output with a virtual default device, but it's not the most self-descriptive thing. Maybe `wasapi-default-device-autorouting` or something? (I'm using `wasapi` because it looks like `cpal` supports ASIO as well, but I'm happy to change to `windows` if that sounds better)

src/host/wasapi/device.rs
philpax: I haven't checked, but I'd imagine it should be pretty negligible. I did consider caching it, but opted against it because the device could change, and we can't detect that without setting up a `IMMNotificationClient` (which is a bit tedious, as you can see from https://github.com/RustAudio/cpal/compare/master...philpax:cpal:audio-device-fix#diff-ce1e8d650aace89ec56c28e4525e838476cffdbe51fe4447d8181936b5ec6724).

My gut feeling is that this is probably fine for now, and it can be revisited if it ends up being a problem.

Cargo.toml
philpax: Hmm... yes, that is a problem. I'll see what I can do; the problem I saw was that the CI would fail to resolve `windows-core` to the correct version, but that might be fine with the bodge I put in there. Will test.

philpax: Well, this is a problem. `cargo` will not resolve `windows-core` to the same version of `windows`'s `windows-core` if I match the version constraints or use a lax constraint like `^0.61`.

I can see three options here:
1. Restrict the version to a single known version (0.61 or 0.62)
2. Ask `windows-rs` to add support for referencing `windows::core`, instead of `windows_core`, somehow: https://github.com/microsoft/windows-rs/issues/3568#issuecomment-2789677229. Even if this happened, it would have to be another `windows-rs` release.
3. Ask a Cargo wizard if there are any solutions for getting `*`-like behaviour in a way that can still be `cargo publish`-ed

I'll ask around regarding 3, and ask in that `windows-rs` issue for future support, but 1 may be necessary in the meantime :S

roderickvd: Thanks for asking in `windows-rs`, how did that go?

The frequent API breakage of these `windows` crates sure is a problem when trying to support a ~12 month window of them. I'm not a Windows user myself, and wouldn't mind "bumping" the Minimum Supported Windows Version. But from what I gather from the community here, historical support is a desired thing to have.

Cargo.toml
roderickvd: I think you can still argue it's additive: it *enables* legacy support.

src/host/wasapi/device.rs
roderickvd: Sure let's leave it then.

Cargo.toml
yara-blue: > But from what I gather from the community here, historical support is a desired thing to have.

Yeah especially with how Microsoft is treating their users on windows 11 we will want to support windows 10 as long as we can. Thank you soo much for your work in that regard philpax and roderick. I know its a pain to support te windows platform, but the impact is so high. Especially for gamedev (bevy).

Cargo.toml
philpax: Fair enough, will action!

Cargo.toml
philpax: `windows-rs` doesn't have any immediate solutions; they're planning on refactoring their crate structure, but as far as I can tell, the result would have a similar problem. (It also wouldn't help us with our historical support of crate versions, even if they did make a change: we'd still have to bump to that version onwards)

> Yeah especially with how Microsoft is treating their users on windows 11 we will want to support windows 10 as long as we can.

Hmm, I'm pretty sure newer versions of the `windows` crate support older Windows versions, it's mostly just that you're likely to have multiple versions of the crate without some mechanism to synchronise them (e.g. windows v0.58 and windows v0.62, etc)

---

Speaking of, I was thinking of another solution: [catpuccin-egui](https://github.com/catppuccin/egui/blob/main/Cargo.toml) uses individual features for each supported version of `egui`, so the user can control which version of `egui` is used by switching out the feature in use.

We could do something similar for both `windows` and `windows-core`, so that the user can select which version to target, defaulting to either the earliest or the latest supported by us (e.g. default on `windows-0-62`). I think this would also solve the issue with `implement`, so we could extend the range of support once more.

Of course, this comes with the downside that the user needs to opt into an older version of `windows`-as-used-by-`cpal`-as-maybe-used-by-`rodio`, but I suspect it's the only viable solution given the constraints at hand (outside of locking down to one `windows` version and potentially bloating users' builds). Thoughts? Do we know of any users who have strong feelings on the matter?

philpax: No, wait, just realised that wouldn't work... the `implement` macro would still look for a global `windows_core`, bypassing any aliases we set up. What a headache 😭

roderickvd: 😢 that's a real pity. There's been an express request from the community to support multiple `windows` crate versions... how best to continue?

philpax: The only solution I can think of is an incredibly ugly one: publish multiple crates (containing just the relevant code) for each supported `windows` version that wrap up the correct `windows` and `windows-core` versions and feature-gate them, so that they're paired together but otherwise isolated.

Anything that depends on the `implement` macro (and thus `windows-core`) will have this issue; I don't see a way around it, unfortunately :/

We could consider avoiding the side-dependency issue by rewriting all Windows-related code to only depend on `windows-core`, but I imagine this will be quite a headache, and it would still restrict the minimum version to 0.61.

roderickvd: Ouch 😓 Anyway the path forward seems to be anticipating v0.61+ right?
Thinking out of the box, anything we could do with `build.rs`?

philpax: Sorry for the late reply. With the current setup, we'd have to pick a `windows` and `windows-core` and stick to them; the only way to avoid this would be to use the multi-crate solution I suggested, as that would be sufficiently decoupled to avoid any issues with Cargo.

I can't think of anything we could do with `build.rs`; the issue is with how Cargo does package resolution, which means any interventions we do need to be done before that stage.

We could noodle around with a `windows-core` -only solution that completely removes the `windows` dependency, but that would still limit us to >=0.61 due to the `implement` feature being removed, so I'm not convinced it's worth the effort to do that.

Honestly, I think the pragmatic choice would be a policy of following the latest `windows`, and letting users deal with the package duplication/having to update their own `windows`. I'm not really sure there's much else that can be done that would be a realistic solution.

roderickvd: > Honestly, I think the pragmatic choice would be a policy of following the latest `windows`, and letting users deal with the package duplication/having to update their own `windows`.

I'm hesitant to do that, because more than a few Windows-using community members had specifically asked for broader crate support. Not being a Windows user myself, I'm following the community voice here.

philpax: Hmm... yeah, we're stuck between a rock and a hard place here. I can't test this, but it might be possible to make the `windows-core` dependency optional and to only turn it on when this feature is turned on, and then document that you need to ensure that the versions of the packages match? It's pretty messy, but I'm all out of ideas now :frowning_face:

philpax:
I've merged in the latest `cpal`, pinned to `windows` 0.62.2, and inverted the feature, as requested two months ago :sweat_smile:

I don't currently have a Windows machine at hand, so I verified compilation through cross-compilation.

Still not sure what to do about the bumping of the `windows` crate; are there any particular users we can ask about the level of disruption it'd cause them? afaict, the worst it'll do is duplicate crates or require them to pull in 0.62.2 for interop - that's not great, but it's not a showstopper, I think

roderickvd:
Superseded by #1183 - credits to @philpax and @sidit77 for their lion's share!

PR #1032, opened 2025-10-13
feat: Add support for custom hosts, devices, and streams

melody-rs:
This PR adds support for writing and using custom `HostTrait`, `DeviceTrait`, and `StreamTrait` implementations

This was already sort of possible before, but `cpal::Data` has no public constructor so writing a custom device implementation wasn't possible.
Actually using the custom host wasn't doable either, as crates like `rodio` only accept a `cpal::Host`/`cpal::Device`, which doesn't support `Host`s defined outside `cpal`

### Why

Currently, adding support for platforms that `cpal` doesn't support requires forking `cpal`. 
Depending on the platform, it might not be possible/desirable to upstream support into `cpal` either

### How

I partially modeled this PR off of [wgpu's custom device support](https://docs.rs/wgpu/latest/wgpu/custom/index.html) and [erased-serde](https://lib.rs/crates/erased-serde).

Custom hosts require everything to be `'static`, which is annoying, but relaxing that requirement would've meant introducing lifetimes in a lot of places that didn't have them previously

There's also the requirement that custom devices and supported configs be `Clone` as well, because of these derives in `platform/mod.rs`:
https://github.com/RustAudio/cpal/blob/2b7dc14503c932e2984c3f9848bc354b137f2032/src/platform/mod.rs#L90-L96
https://github.com/RustAudio/cpal/blob/2b7dc14503c932e2984c3f9848bc354b137f2032/src/platform/mod.rs#L122-L128
https://github.com/RustAudio/cpal/blob/2b7dc14503c932e2984c3f9848bc354b137f2032/src/platform/mod.rs#L130-L136
Removing those derives would make the code for custom devices simpler

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,8 +15,9 @@ asio = [
     "num-traits",
 ] # Only available on Windows. See README for setup instructions.
 
+# Support for user-defined custom hosts, devices, and streams
+# See examples/custom.rs for usage
 custom = []
-
 
 [dependencies]
 dasp_sample = "0.11"


roderickvd:
Thanks for the PR! I'm off for a little while and will look at it later. It's an interesting feature - however curious for what kind of hosts one would want to add proprietary support instead of contributing it to cpal?

melody-rs:
Any hosts written against proprietary code (anything console-specific for example) couldn't be upstreamed for legal reasons

There's definitely other reasons not to upstream a host aside from that, certain hosts could be very niche or be designed for very specific use-cases!

I wrote this PR after I tried to make a custom cpal backend for the 3ds using [libctru](https://github.com/devkitPro/libctru) and couldn't. In order to upstream it, I'd have to make the host production ready, and after it'd be upstreamed the cpal maintainers would have to maintain a host for a very niche platform

src/host/custom/mod.rs
```diff
@@ -0,0 +1,414 @@
...
+    fn supports_output(&self) -> bool {
+        <T as DeviceTrait>::supports_output(self)
+    }
+
+    fn supported_input_configs(&self) -> Result<SupportedConfigs, SupportedStreamConfigsError> {
+        <T as DeviceTrait>::supported_input_configs(self).map(supported_configs_to_erased)
+    }
+
+    fn supported_output_configs(&self) -> Result<SupportedConfigs, SupportedStreamConfigsError> {
+        <T as DeviceTrait>::supported_input_configs(self).map(supported_configs_to_erased)
```
roderickvd: ➡️ `supported_output_configs` (instead of _input_ configs)

Cargo.toml
```diff
@@ -15,6 +15,10 @@ asio = [
     "num-traits",
 ] # Only available on Windows. See README for setup instructions.
 
+custom = []
+
+default = ["custom"]
```
roderickvd: Probably shouldn't be the default.

src/host/custom/mod.rs
melody-rs: Oops! Dunno how I missed that

melody-rs:
I tried to make the example generally show how `Host` impls should be written while not getting into the details too much

roderickvd:
Merged. Thank you!

Issue #1029, opened 2025-10-08
High latency on pipewire-alsa with BufferSize::Default


the-drunk-coder:
The following commit can lead to long output delays when using the ALSA host with the default device and `BufferSize::Default` setting.

commit id: b11f8f1e1103f8415ec932bd0ba11af21475e8fd 
commit message: fix: simplify ALSA buffer size and period configuration logic

**Environment:** 

* Intel ThinkPad, `sof-hda-dsp` module
* Arch Linux
* Pipewire
* Pipewire drop-in replacements for ALSA, JACK, etc.

**The symptoms:** 
 
(imagine a drum-machine-style program as a context here)

* Before: using `BufferSize::Default` would lead to an effective buffer size of 1536, now it ends up being 512 (system blocksize)
* Audio thread is definitely running, callback is called.
* I can see data sent through a lock-free queue to my frontend arriving within an expected timeframe (like, peak meters etc move as expected)
* Audio, on the other hand, arrives to the headphones with about 15 seconds delay

abique:
Did you use an alsa plugin that was talking to pipewire, or you opened the alsa device exclusively?

the-drunk-coder:
> Did you use an alsa plugin that was talking to pipewire, or you opened the alsa device exclusively?

the pipewire alsa plugin, i think ... as mentioned before, I have `pipewire-alsa` installed, so my assumption is that it will talk to pipewire.

abique:
Yeah, so your latency settings in cpal are potentially irrelevant.
Did you configure the quantum correctly in pipewire?

Anyway, you should release the device from pipewire, and use it directly with alsa and then report.

the-drunk-coder:
Pipewire quantum is set to 512.

But, there's definitely something breaking here, isn't it? 

**BEFORE** the mentioned commit, with the default buffer size (IIRC `BufferSize::Fixed` had no effect), the latency was comparatively high (1536 samples), but within reason.

**AFTER** the mentioned commit, all other things being equal, I can suddenly detect the "correct" buffer size (the quantum configured in pipewire) in the audio callback even without requesting `BufferSize::Fixed`, but the latency is *incredibly* high ... it's not like we're talking about a few milliseconds here, more like 15-20 seconds.

No other sound-producing program shows that behaviour. Choosing JACK as a host and setting `BufferSize::Fixed` to the pipewire quantum produces low latency, as expected (i.e. when both are set to 128). 

[...]

PR #1033, opened 2025-10-14
fix(alsa): buffering behavior on PipeWire-ALSA

roderickvd:
Fixes high latency with `BufferSize::Default` on PipeWire-ALSA by avoiding period count constraints that cause pathologically large periods.

Fixes
- #1029
- #1034 
- #1036

### DIFF
--- a/src/host/alsa/mod.rs
+++ b/src/host/alsa/mod.rs
@@ -1150,9 +1150,9 @@ impl StreamTrait for Stream {
     }
 }
 
-// Overly safe clamp because alsa Frames are i64
+// Overly safe clamp because alsa Frames are i64 (64-bit) or i32 (32-bit)
 fn clamp_frame_count(buffer_size: alsa::pcm::Frames) -> FrameCount {
-    buffer_size.clamp(1, FrameCount::MAX as _) as _
+    buffer_size.clamp(1, FrameCount::MAX as alsa::pcm::Frames) as FrameCount
 }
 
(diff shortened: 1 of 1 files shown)


roderickvd:
Note to self: drop should wait for two periods worth of time, not the full buffer.

src/host/alsa/mod.rs
```diff
@@ -1287,22 +1277,25 @@ fn set_hw_params_from_format(
...
 
-    // Configure period size based on buffer size request
-    // When BufferSize::Fixed(x) is specified, we request a period size of x frames
-    // to achieve approximately x-sized callbacks. ALSA may adjust this to the nearest
-    // supported value based on hardware constraints.
+    // When BufferSize::Fixed(x) is specified, we configure double-buffering with
+    // buffer_size = 2x and period_size = x. This provides consistent low-latency
+    // behavior across different ALSA implementations and hardware.
     if let BufferSize::Fixed(buffer_frames) = config.buffer_size {
+        hw_params.set_buffer_size_near((2 * buffer_frames) as _)?;
```
jwagner: Might be better to use into here so there is a compile time error in case the types become incompatible.

roderickvd: Thanks for looking along. I'm thinking to explicitly cast as `alsa::pcm::Frames` instead. Rationale: on 32-bit platforms, `Frames` will be `i32` and that'd wouldn't compile with `into()`.

roderickvd:
Positive feedback over at #1036 so merging this. Took a while to get this playing nice with PipeWire-ALSA 😓 
Thanks for all the testing so far, and please test again before we get v0.17 out in the wild!

Issue #890, opened 2024-06-12
Min buffer size range detected on Android is huge (~1sec)


kuviman:
While figuring out a mysterious audio delay on android when using `web-audio-api` crate - https://github.com/orottier/web-audio-api-rs/issues/515

I found that cpal reports a huge min buffer size (when using `Device::default_output_config`)

It looks like the min buffer size is calculated using java api [AudioTrack.getMinBufferSize](https://developer.android.com/reference/android/media/AudioTrack#getMinBufferSize(int,%20int,%20int))

Trying using that API directly in a sample Kotlin app like this:

```kt
val size = AudioTrack.getMinBufferSize(44100, AudioFormat.CHANNEL_OUT_STEREO, AudioFormat.ENCODING_PCM_FLOAT);
```

reports the same huge buffer size (~40000, a bit random every time).

Tested on Google Pixel 7 Pro

Since `web-audio-api` was clamping buffer size to the reported range I got a ~1 sec delay.

On one hand the issue is not in cpal itself, but whatever the reason is for this huge min buffer size.

On the other hand, I thought that maybe using java apis here doesnt make much sense since the actually used audio api is oboe the c++ library.

I'm not sure what to do with any of that but thought you guys might know about my findings here.

In `web-audio-api` the solution was to just use `Default` buffer size without trying to calculate it based on reported ranges.

PR #1039, opened 2025-10-27
fix(aaudio): Correctly detect minimum buffer size

will3942:
AAudio buffer size minimum should be detected using the PROPERTY_OUTPUT_FRAMES_PER_BUFFER on the AudioManager not using AudioTrack.

Android documentation reference: https://developer.android.com/ndk/guides/audio/audio-latency#buffer-size

This allows for a sensible buffer size minimum to be used in comparison to the 1 second+ buffer sizes reported by AudioTrack.

Tested on a Samsung Galaxy Tab A9: `SupportedStreamConfigRange { channels: 1, min_sample_rate: SampleRate(5512), max_sample_rate: SampleRate(5512), buffer_size: Range { min: 256, max: 2147483647 }, sample_format: I16 }`

Closes: https://github.com/RustAudio/cpal/issues/890

### DIFF
--- a/src/host/aaudio/java_interface/audio_manager.rs
+++ b/src/host/aaudio/java_interface/audio_manager.rs
@@ -27,5 +27,7 @@ fn get_frames_per_buffer<'j>(env: &mut JNIEnv<'j>, context: &JObject<'j>) -> JRe
     let frames_per_buffer_string = String::from(env.get_string(&frames_per_buffer)?);
 
     // TODO: Use jni::errors::Error::ParseFailed instead of jni::errors::Error::JniCall once jni > v0.21.1 is released
-    frames_per_buffer_string.parse::<i32>().map_err(|e| jni::errors::Error::JniCall(jni::errors::JniError::Unknown))
+    frames_per_buffer_string
+        .parse::<i32>()
+        .map_err(|e| jni::errors::Error::JniCall(jni::errors::JniError::Unknown))
 }


src/host/aaudio/mod.rs
```diff
@@ -91,17 +89,13 @@ impl HostTrait for Host {
...
-    android_format: i32,
-) -> SupportedBufferSize {
-    let min_buffer_size = if is_output {
-        get_audio_track_min_buffer_size(sample_rate, channel_mask, android_format)
+fn buffer_size_range() -> SupportedBufferSize {
+    let min_buffer_size = if let Ok(min_buffer_size) = AudioManager::get_frames_per_buffer() {
+        min_buffer_size
     } else {
-        get_audio_record_min_buffer_size(sample_rate, channel_mask, android_format)
+        0
```
roderickvd: I think this could be refactored into something like:

```rust
if let Ok(min_buffer_size) = AudioManager::get_frames_per_buffer() {
    SupportedBufferSize::Range {
        min: min_buffer_size as u32,
        max: i32::MAX as u32,
    }
} else {
    SupportedBufferSize::Unknown
}
```

src/host/aaudio/mod.rs
```diff
@@ -112,26 +106,16 @@ fn buffer_size_range_for_params(
...
 
     let mut output = Vec::with_capacity(SAMPLE_RATES.len() * CHANNEL_MASKS.len() * FORMATS.len());
     for sample_format in &FORMATS {
-        let android_format = if *sample_format == SampleFormat::I16 {
-            android_media::ENCODING_PCM_16BIT
-        } else {
-            android_media::ENCODING_PCM_FLOAT
-        };
-        for (mask_idx, channel_mask) in CHANNEL_MASKS.iter().enumerate() {
+        for (mask_idx, _) in CHANNEL_MASKS.iter().enumerate() {
```
roderickvd: Do we still need `CHANNEL_MASKS`? Because currently the elements that that array contains aren't used anymore.

src/host/aaudio/mod.rs
```diff
@@ -112,26 +106,16 @@ fn buffer_size_range_for_params(
...
             let channel_count = mask_idx + 1;
```
roderickvd: ...but maybe they should because while this line works, it's kind of hackish (I know, not your code, nor was it mine 😉). It at some point a 5.1 or whatever mode is introduced, we should use a mapping instead of doing + 1.

src/host/aaudio/java_interface/audio_manager.rs
```diff
@@ -0,0 +1,34 @@
+use super::{
+    utils::{
+        get_context, get_property, get_system_service, with_attached, JNIEnv, JObject, JResult,
+    },
+    AudioManager, Context,
+};
+
+impl AudioManager {
+    /**
+     * Get the frames per buffer using Android Java API
```
roderickvd: Style nitpick: please use Rustdoc-standard `//` comments instead of Javadoc-style `/** .. */`.

src/host/aaudio/java_interface/audio_manager.rs
```diff
@@ -0,0 +1,34 @@
...
+    }
+}
+
+fn get_frames_per_buffer<'j>(env: &mut JNIEnv<'j>, context: &JObject<'j>) -> JResult<i32> {
+    let audio_manager = get_system_service(env, context, Context::AUDIO_SERVICE)?;
+
+    let frames_per_buffer = get_property(
+        env,
+        &audio_manager,
+        AudioManager::PROPERTY_OUTPUT_FRAMES_PER_BUFFER,
```
roderickvd, line 24: Question: is this also valid for input devices? Though I see that `AudioManager::PROPERTY_INPUT_FRAMES_PER_BUFFER` does not exist.

src/host/aaudio/mod.rs
```diff
@@ -112,26 +106,16 @@ fn buffer_size_range_for_params(
...
 
-fn default_supported_configs(is_output: bool) -> VecIntoIter<SupportedStreamConfigRange> {
+fn default_supported_configs() -> VecIntoIter<SupportedStreamConfigRange> {
     // Have to "brute force" the parameter combinations with getMinBufferSize
     const FORMATS: [SampleFormat; 2] = [SampleFormat::I16, SampleFormat::F32];
 
     let mut output = Vec::with_capacity(SAMPLE_RATES.len() * CHANNEL_MASKS.len() * FORMATS.len());
     for sample_format in &FORMATS {
-        let android_format = if *sample_format == SampleFormat::I16 {
-            android_media::ENCODING_PCM_16BIT
```
roderickvd, line 122: In `android_media.rs` can these constants now be removed?

will3942:
> Thanks for the contribution!

Thanks for the review - will address all feedback in the morning. The Q around input is an interesting one, I couldn't see anything different for input in their docs but also I agree the property is named after output so...

src/host/aaudio/java_interface/audio_manager.rs
will3942: As far as I can see, yes - the docs I linked in the PR description reference `PROPERTY_OUTPUT_FRAMES_PER_BUFFER` in the context of input.

src/host/aaudio/mod.rs
will3942: Good shout - implemented.

src/host/aaudio/mod.rs
will3942: Yeah this wasn't ideal before but I've reworked this now to use a mapping so that we can support more channel configurations in the future.

src/host/aaudio/java_interface/audio_manager.rs
will3942: Done - thanks!

src/host/aaudio/mod.rs
will3942: Yes - removed.

will3942:
This works well now - longer term I think we should look at a few feature requests:
1. Allow querying the configuration that the stream is using - e.g. if you use BufferSize::Default, you can see the buffer size the underlying host has used. AAudio supports this, I'm sure others do too.
2. Edit the stream configuration (where supported) after construction - AAudio supports/recommends this for buffer size tuning
3. Bringing back Oboe support - Oboe does some magic around buffer size / latency setting that we could leverage. The Android docs for ndk/aaudio recommend this.

src/host/aaudio/java_interface/audio_manager.rs
```diff
@@ -0,0 +1,32 @@
...
+    )?;
+
+    let frames_per_buffer_string = String::from(env.get_string(&frames_per_buffer)?);
+
+    frames_per_buffer_string
+        .parse::<i32>()
+        .map_err(|e| jni::errors::Error::JniCall(jni::errors::JniError::Unknown))
```
roderickvd, line 32: Would it make sense to retain the error message and return something like `jni::errors::Error::ParseError(format!("Failed to parse frames per buffer: {}", e))`?

src/host/aaudio/android_media.rs
```diff
@@ -6,8 +6,6 @@ use self::jni::Executor;
 use self::jni::{errors::Result as JResult, JNIEnv, JavaVM};
 
 // constants from android.media.AudioFormat
-pub const ENCODING_PCM_16BIT: i32 = 2;
-pub const ENCODING_PCM_FLOAT: i32 = 4;
 pub const CHANNEL_OUT_MONO: i32 = 4;
```
roderickvd: This and `CHANNEL_OUT_STEREO` could now be removed?

src/host/aaudio/android_media.rs
roderickvd, line 1: This module is now so small that it may make sense to inline what was left.

roderickvd:
> This works well now - longer term I think we should look at a few feature requests:
> 
> 1. Allow querying the configuration that the stream is using - e.g. if you use BufferSize::Default, you can see the buffer size the underlying host has used. AAudio supports this, I'm sure others do too.

Yes. #964 also added querying latency - we could go from there.

> 2. Edit the stream configuration (where supported) after construction - AAudio supports/recommends this for buffer size tuning

That's interesting for mobile platforms. I believe iOS does too?
Question is how this could work elegantly in cpal.

> 3. Bringing back Oboe support - Oboe does some magic around buffer size / latency setting that we could leverage. The Android docs for ndk/aaudio recommend this.

You know why Oboe was removed in the first place? (I don't.)

will3942:
> Yes. #964 also added querying latency - we could go from there.

Agreed. Documeted as an issue / feature request - https://github.com/RustAudio/cpal/issues/1042

> That's interesting for mobile platforms. I believe iOS does too? Question is how this could work elegantly in cpal.

This could work as functions on the Stream trait? Although I'm not sure if this is universally supported functionality across platforms.

> You know why Oboe was removed in the first place? (I don't.)

Pros/cons where detailed here - https://github.com/RustAudio/cpal/pull/961

src/host/aaudio/java_interface/audio_manager.rs
will3942: I've fixed this, the master version of `jni` is nicer, to do this with the current supported version I have had to add `combine` as a dependency which I don't love. Let me know what you think? I can't see a great solution here for passing the error back through.

will3942:
@roderickvd addressed the nits - let me know your feedback on the last one.

roderickvd:
> > That's interesting for mobile platforms. I believe iOS does too? Question is how this could work elegantly in cpal.
> 
> This could work as functions on the Stream trait? Although I'm not sure if this is universally supported functionality across platforms.

We'd have to map out the API and judge how it'd impact the various hosts. Preferably, it'd keep things simple.

The ALSA host would also be a candidate, because we can play the same game: requesting a large buffer with X periods but only using the first Y periods of it, and dynamically tuning Y. But, to the point of keeping things simple, we'd have to play around to see how easy it is to do that in the poll cycle.

> > You know why Oboe was removed in the first place? (I don't.)
> 
> Pros/cons where detailed here - #961

Ah, thanks, could've looked for that myself 😊 That makes it seem like there's almost no cons and all pros, yet you refer to something that Oboe does by itself and NDK doesn't? If we can implement dynamic buffers without the weight of Oboe, that'd also be nice, no?

> @roderickvd addressed the nits - let me know your feedback on the last one.

👍 I'll take a look.

src/host/aaudio/java_interface/audio_manager.rs
roderickvd: Agreed, if we can prevent pulling in another dependency then let's put in that effort. How about `JniCall(JniError::Other(String))`?

will3942: `JniError::Other(jint)` is the signature so it won't accept a string.

We only really have ParseFailed with the additional dependency or FieldAlreadySet(String)

will3942: @roderickvd The crate jni will get released at some point which will allow us to use ParseFailed(String) without an external dependency. In the meantime I think we have three choices:

1) Pull in combine dependency and use ParseFailed
2) Use FieldAlreadySet(String)
3) Add a TODO to update to ParseFailed(String) on jni release and just return the error without the string.

I'm leaning towards 3 or 2 (also with a TODO).

What's your preference? I can then implement this - I'd love to get this across the line with the next release so we don't have to run a patch for cpal.

will3942: I've updated this and gone with 3 - this feels the best in my mind, let me know if you want me to change.

roderickvd: Makes sense. Let's not spend too much time over it. If you'd want to track it and deal with it later?

src/host/aaudio/mod.rs
```diff
@@ -7,30 +6,30 @@ use std::vec::IntoIter as VecIntoIter;
...
 // TODO: more channels available in native AAudio
-const CHANNEL_MASKS: [i32; 2] = [
-    android_media::CHANNEL_OUT_MONO,
-    android_media::CHANNEL_OUT_STEREO,
-];
+// Maps channel masks to their corresponding channel counts
+const CHANNEL_CONFIGS: [(i32, u16); 2] = [(CHANNEL_OUT_MONO, 1), (CHANNEL_OUT_STEREO, 2)];
 
 const SAMPLE_RATES: [i32; 13] = [
     5512, 8000, 11025, 16000, 22050, 32000, 44100, 48000, 64000, 88200, 96000, 176_400, 192_000,
```
roderickvd, line 35: I know this isn't yours, but now I read it I think that it should also support 12 and 24 kHz.

roderickvd:
Thanks for your PR and efforts. I'm gonna merge this now, and add the 12/24 kHz after manually. Let me know if I'm wrong.

PR #1041, opened 2025-11-09
add missing packed(4) for asio time structs

iftach-cogntiv:
Some structs in the asio bindings for asio-sys created by bindgen are packed(4). The mod.rs implementation of these structs in asio-sys/src/bindings are missing this representation flag. This fix adds it.
Without  the packed(4) representation flag, in rust versions > 1.88.0 compilation fails in debug mode due to validation on access to memory address aligned to 8. Specifically this can be recreated with Sound Blaster's asio driver.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
 # Unreleased
-- ASIO: Add packed(4) to representation of asio time structs in bindings
+
 - Add `Sample::bits_per_sample` method.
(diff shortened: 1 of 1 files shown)


roderickvd:
Thank you. Would you add an entry to the changelog?

iftach-cogntiv:
> Thank you. Would you add an entry to the changelog?

Added

PR #1048, opened 2025-11-14
Fix a build break for the AudioWorklet backend

kettle11:
This fixes a build break in the recently added AudioWorklet backend. [11dea8d](https://github.com/RustAudio/cpal/commit/11dea8d3c729ea9a57405364edc270183ebf8632) added an ID that each backend needs to implement and that wasn't part of the AudioWorklet backend yet. The sequencing was sort of unlucky: I signed off on the AudioWorklet PR, the ID change merged, and then the AudioWorklet PR merged already broken within the span of a few hours.

This does indicate that the AudioWorklet backend should be in the build tests to avoid this in the future.


roderickvd:
Absolutely. Thanks for the quick turnaround time on this one.

src/host/web_audio_worklet/mod.rs
```diff
@@ -96,6 +98,11 @@ impl DeviceTrait for Device {
         Ok("Default Device".to_owned())
     }
 
+    #[inline]
```
roderickvd, line 101: This `inline` pragma and most (all?) others can go because this code isn't in any hot path. But no need to do that in this PR.

PR #1049, opened 2025-11-14
Add the AudioWorklet backend to the Github CI

kettle11:
As mentioned in https://github.com/RustAudio/cpal/pull/1048 it'd be good to add the AudioWorklet backend to the CI to avoid build breaks.

This change adds that. The risk here is that the atomics feature requires nightly Rust to compile that backend. _Hopefully_ that eventually gets resolved, but it's been half a decade now and progress is just maybe picking back up: https://github.com/rust-lang/rust/issues/77839

A number of projects do use nightly in their CI and it works out OK, but if it proves to be an issue it could be disabled later.

### DIFF
--- a/.github/workflows/cpal.yml
+++ b/.github/workflows/cpal.yml
@@ -213,11 +213,13 @@ jobs:
             setup-emscripten: true
             args: --release
           - target: wasm32-unknown-unknown
-            setup-emscripten: false
             args: --features=wasm-bindgen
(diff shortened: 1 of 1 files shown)


kettle11:
Note that the failing check in this PR is because it's now correctly catching the build break addressed by https://github.com/RustAudio/cpal/pull/1048 .

PR #1052, opened 2025-11-21
fix(coreaudio): fix undefined behavior causing sanitizer crashes

sinkingsugar:
## Summary

Fixes undefined behavior in CoreAudio backend that causes crashes under AddressSanitizer and ThreadSanitizer. This was introduced in #943 when migrating from `coreaudio-rs` to `objc2-core-audio`.

## The Bug

In 10 locations across `macos/enumerate.rs` and `macos/device.rs`, `data_size` variables were declared immutable but passed to CoreAudio APIs that write to them:

```rust
// BROKEN - causes UB
let data_size = 0u32;
AudioObjectGetPropertyDataSize(..., NonNull::from(&data_size), ...)
//                                   ^^^^^^^^^^^^^^^^^^^^^^^^^ immutable ref!
```

CoreAudio's size parameter is `inout` - the API writes the actual data size back. Creating `NonNull` from an immutable reference and then writing through it is undefined behavior.

## Why It Wasn't Caught Earlier

The old raw pointer code hid the UB with unsafe casts:
```rust
&data_size as *const _ as *mut _  // Unsafe cast bypasses Rust's type system
```

The new `objc2-core-audio` API properly enforces mutability contracts via `NonNull`, exposing the latent bug. Normal builds work fine because at the machine level the stack memory is writable, but **sanitizers enforce Rust's safety contracts** and crash.

## The Fix

Changed all affected variables to `mut` and updated `NonNull` calls:
```rust
// FIXED
let mut data_size = 0u32;
AudioObjectGetPropertyDataSize(..., NonNull::from(&mut data_size), ...)
```

**Files changed:**
- `src/host/coreaudio/macos/enumerate.rs`: 2 `data_size` variables, 4 API calls
- `src/host/coreaudio/macos/device.rs`: 7 `data_size` variables, 14 API calls

## Reproducing the Bug

```bash
# With cpal 0.16 (before this fix):
cargo +nightly test -Zbuild-std --target aarch64-apple-darwin \
  --lib --verbose
RUSTFLAGS="-Zsanitizer=address"
# ☠️ CRASH: AddressSanitizer: SEGV on unknown address
```

The existing tests (`test_play`, `test_record`) trigger device enumeration, which hits the buggy code.

## Bonus: Sanitizer CI Workflow

This PR includes a new `.github/workflows/sanitizers.yml` that would have caught this bug before release. It runs ASAN and TSAN on both macOS (CoreAudio) and Linux (ALSA) since UB can be platform-specific.

Benefits:
- Catches memory safety bugs, use-after-free, buffer overflows (ASAN)
- Catches race conditions, data races (TSAN)
- Runs on existing tests with no code changes needed
- Would have prevented this UB from reaching 0.16 release

## Checklist

- [x] Fixed all 10 instances of the bug
- [x] Verified fix compiles (`cargo check`)
- [x] Tested with sanitizers locally (crashes fixed)
- [x] Added sanitizer CI to prevent future UB

## Related

- Root cause: #943 (objc2-core-audio migration)
- This affects cpal 0.16.0 only (0.15 used old API)

### DIFF
--- a/.github/workflows/sanitizers.yml
+++ b/.github/workflows/sanitizers.yml
@@ -1,0 +1,92 @@
+name: Sanitizers
+
+on:
+  push:
+    branches: [master]
+    paths-ignore:
+      - "**.md"
+      - "docs/**"
+      - "LICENSE*"
+      - ".gitignore"
+  pull_request:
+    branches: [master]
+    paths-ignore:
+      - "**.md"
+      - "docs/**"
+      - "LICENSE*"
+      - ".gitignore"
+  workflow_dispatch:
+
+jobs:
+  sanitizers:
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - os: macOS-latest
+            target: aarch64-apple-darwin
+            sanitizer: address
+            name: ASAN-macOS
+
+          - os: macOS-latest
+            target: aarch64-apple-darwin
+            sanitizer: thread
+            name: TSAN-macOS
+
+          - os: ubuntu-latest
+            target: x86_64-unknown-linux-gnu
+            sanitizer: address
+            name: ASAN-Linux
+
+          - os: ubuntu-latest
+            target: x86_64-unknown-linux-gnu
+            sanitizer: thread
+            name: TSAN-Linux
+
+    name: ${{ matrix.name }}
+    runs-on: ${{ matrix.os }}
+    steps:
+      - uses: actions/checkout@v5
+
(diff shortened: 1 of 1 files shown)


sinkingsugar:
> Very clear description and nice enhancement to ensure safety. I think the workflow can be cleaned up by using matrix strategy.

Thank you, I pushed a cleanup using matrix strategy.

roderickvd:
Looks great, thanks. Merging now.

Issue #1028, opened 2025-10-01
Segfault when listing devices macOS through CoreAudio host


alexjthomson:
Hey 👋

It would appear that when executing the following code on macOS you get a segfault (make sure you have `RUST_BACKTRACE=1`:
```rust
let host = cpal::default_host();
let _ host.output_devices(); // segfault occurs here
```

This is purely speculation, but I'd imagine this is because when you iterate over the output devices, really you're iterating over every device and filtering out by those that are output devices. This is a problem if the application doesn't have microphone permission and will cause the segfault in the CoreAudio host.

This is not a problem in older versions of cpal.

__Edit__: My system specs are as follows:
- MacBook Pro November 2024
- Apple M4 Pro
- 48GB RAM
- macOS 15.7

simlay:
Hmm. Curious. This failure does not reproduce for me. Here's a [cargo script](https://internals.rust-lang.org/t/pre-rfc-cargo-script-for-everyone/18639) that works on macOS 15.7., apple silicon M1, xcode 26.

```
#!/usr/bin/env cargo +nightly -Zscript
---
[dependencies]
cpal = "0.16.0"
---

fn main() {
    use cpal::traits::HostTrait;
    let host = cpal::default_host();
    let _ = host.output_devices(); // segfault occurs here
    println!("Hello world");
}
```

roderickvd:
> This is not a problem in older versions of cpal.

And in which version are you experiencing it now? Is this on master (which commit) or v0.16?

alexjthomson:
Apologies for not including such an important piece of information! I'm on 0.16.0

I'm unsure what exactly is causing the segfault but as mentioned I'm assuming it's a quirk with macOS. Unfortunately I'm unfamiliar with how macOS works so I'm not sure what is required to reproduce this.

roderickvd:
As additional data point would you check on current master? The CoreAudio code was refactored there so if there _is_ something to fix then I want to take that as baseline.

roderickvd:
At least with latest master, I can't reproduce this on my MacBook Air M3 on either macOS Sequoia 15.6.1 or 15.7.

yury:
MacBook Air M2, macOS Tahoe 26.1 beta 3 - no segfault

jcalderw:
[...]

PR #1055, opened 2025-12-02
Fix coreaudio/macos UB

jcalderw:
#1028

Fix UB from immutable pointers and misaligned buffers, correct the nominal sample rate setter, and use the correct type for buffer frame size.

- Replace immutable `data_size` pointers with mutable ones in CoreAudio backend (`device.rs` and `enumerate.rs`).
- Replace misaligned `Vec<u8>` temporary buffers with typed, aligned allocations for `AudioValueRange` queries; avoid forming references from possibly misaligned `AudioBufferList` by using unaligned reads.
- Set `kAudioDevicePropertyNominalSampleRate` using a single `f64` with the correct size (was incorrectly passing an `AudioValueRange`).
- Read `kAudioDevicePropertyBufferFrameSize` as `u32` and cast to `usize` (was read as `usize`).

These updates may also apply to coreaudio/ios, but I have not checked.


jcalderw:
Well, after submitting this, I noticed PR #1052. There is likely overlap here. I should have checked before submitting. Sorry for the churn!

roderickvd:
😄 no worries your contribution is valued nonetheless! I'm working as we speak on reviewing #1052. Let's get that in first, and then it'd be great if you could pick up after anything you offer here that's not offered there.

roderickvd:
I'm surprised this still merges cleanly after merging #1052. Just in case, would you be so kind to rebase your changes on `master` now?

jcalderw:
Done (I think)! The only things I missed from #1052 are some additional mutables (which I'm not sure are strictly necessary, but probably don't hurt).

src/host/coreaudio/macos/device.rs
```diff
@@ -104,18 +102,18 @@ fn set_sample_rate(
...
         // Now that we have the available ranges, pick the one matching the desired rate.
         let sample_rate = target_sample_rate.0;
         let maybe_index = ranges
             .iter()
             .position(|r| r.mMinimum as u32 == sample_rate && r.mMaximum as u32 == sample_rate);
-        let range_index = match maybe_index {
-            None => return Err(BuildStreamError::StreamConfigNotSupported),
-            Some(i) => i,
-        };
+        if maybe_index.is_none() {
```
roderickvd: Could this and the above now be simplified to:

```rust
if !ranges.iter().any(|r|
    r.mMinimum as u32 == sample_rate && r.mMaximum as u32 == sample_rate
) {
    return Err(BuildStreamError::StreamConfigNotSupported);
}
```

roderickvd:
> These updates may also apply to coreaudio/ios, but I have not checked.

Maybe so, using unaligned reads:

```rust
let buffer = if is_input {
    let nb_ptr = core::ptr::addr_of!((*args.data.data).mNumberBuffers);
    let len = core::ptr::read_unaligned(nb_ptr) as usize;    
    let first_buf_ptr = 
        core::ptr::addr_of!((*args.data.data).mBuffers) as *const AudioBuffer;
    core::ptr::read_unaligned(first_buf_ptr)
} else {
    let buf_ptr = core::ptr::addr_of!((*args.data.data).mBuffers[0]);
    core::ptr::read_unaligned(buf_ptr)
};
```

jcalderw:
Great suggestions! I've incorporated and rebased.

roderickvd:
Hey, that's quick 😄 

Commit log and number of changed files is kinda weird... did you merge master in or rebase? You'd expect rebasing to only show your changes.

jcalderw:
I thought that I rebased, but I agree that something was weird. I get confused by git most days. I don't know why the merge had to happen; hopefully it doesn't cause a problem. And then CI exposed several clippy issues that I didn't see locally; that's the explanation for the last few commits. There's probably a better way for me to set up my local environment. I'm using GitKraken and RustRover on Apple silicon; suggestions are welcome. :)

roderickvd:
[fe361be](https://github.com/RustAudio/cpal/pull/1055/commits/fe361bee2be6429625e0f2f7786eb68467851df5) was a merge instead of a rebase. Could you try rebasing?

jcalderw:
OK. It's much cleaner now.

roderickvd:
Thanks a lot, merged!

PR #1056, opened 2025-12-02
refactor: split CI into quality, platforms, and sanitizers workflows

roderickvd:
Now that sanitization has been added, it seems appropriate to clean up the various workflows.


roderickvd:
Also let's try and fix those clippy warnings now platform by platform.

PR #1058, opened 2025-12-06
ASIO: Handle kAsioResetRequest to fix unresponsive driver changes.

### LINKED ISSUE #1007: asio-sys: handle asio messages
Currently TODOs in code: https://github.com/RustAudio/cpal/blob/85fd82f5675e433e98920d7e693162f281137ba9/asio-sys/src/bindings/mod.rs#L934-L941

I'm not sure if the plan was to handle everything internally in cpal or to let the user listen for the messages for manual handling, but currently neither is supported.
The result is being unable to detect when a user changes audio settings in e.g. the ASIO4ALL GUI.

Sin-tel:
Fixes #1007.

This PR implements handling for the kAsioResetRequest message from the ASIO driver. This message is sent by drivers (e.g., ASIO4ALL, or when changing sample rates via a control panel) to signal that the stream must be destroyed and recreated. Currently, cpal ignores this, causing the UI/Driver to become unresponsive.

I *only* handle `kAsioResetRequest`. We could handle more, but this is the only one strictly necessary to fix the issue. It's the user's responsibility to handle it and rebuild the whole stream.

It sends a `BackendSpecificError` with a string description. It's a bit awkward for the user to match against a string, but I didn't want to touch the cross-platform `StreamError` enum and break the API for everyone else.

### DIFF
--- a/src/host/asio/stream.rs
+++ b/src/host/asio/stream.rs
@@ -652,7 +652,7 @@ impl Device {
             // Check specifically for ResetRequest
             if let sys::AsioMessageSelectors::kAsioResetRequest = msg {
                 if let Ok(mut cb) = error_callback_shared.lock() {
-                    cb(StreamError::AsioResetRequest);
+                    cb(StreamError::StreamInvalidated);
                 }
             }
         })
--- a/src/error.rs
+++ b/src/error.rs
@@ -289,14 +289,16 @@ impl From<BackendSpecificError> for PauseStreamError {
 
 /// Errors that might occur while a stream is running.
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
-#[non_exhaustive]
 pub enum StreamError {
     /// The device no longer exists. This can happen if the device is disconnected while the
     /// program is running.
     DeviceNotAvailable,
-    /// The ASIO driver requests a reset, in case of an unexpected event or a reconfiguration.
-    /// Rebuilding the stream must be deferred to the next safe time, as this code is called from the driver.
-    AsioResetRequest,
+    /// The stream configuration is no longer valid and must be rebuilt.
+    /// This occurs when the backend requires a full reset:
+    /// - ASIO: Driver reset request (sample rate/format change via control panel)
+    /// - JACK: Sample rate changed by server
+    /// - Other backends: Format/configuration invalidated
+    StreamInvalidated,
     /// See the [`BackendSpecificError`] docs for more information about this error variant.
     BackendSpecific { err: BackendSpecificError },
 }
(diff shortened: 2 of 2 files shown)


asio-sys/src/bindings/mod.rs
```diff
@@ -9,7 +9,7 @@ use std::ffi::{CStr, CString};
 use std::os::raw::{c_char, c_double, c_long, c_void};
 use std::ptr::null_mut;
 use std::sync::{
-    atomic::{AtomicBool, Ordering},
+    atomic::{AtomicBool, AtomicU32, Ordering},
```
roderickvd: Where is the `AtomicU32` used?

asio-sys/src/bindings/mod.rs
```diff
@@ -936,7 +970,17 @@ extern "C" fn asio_message(
...
-            // TODO: Handle this.
+
+            // Get the list of active message callbacks.
+            let callbacks: Vec<_> = {
+                let lock = MESSAGE_CALLBACKS.lock().unwrap();
+                lock.iter().map(|(_, cb)| cb.0.clone()).collect()
+            };
+            // Release lock and call them.
+            for cb in callbacks {
+                cb(AsioMessageSelectors::kAsioResetRequest);
```
roderickvd, line 981: The ASIO SDK documentation states: "You cannot reset the driver right now, as this code is called from the driver." - we might want to document that somewhere.

src/host/asio/stream.rs
```diff
@@ -60,6 +61,22 @@ impl Device {
...
+        // Register the message callback with the driver
+        let error_callback_shared = Arc::new(Mutex::new(error_callback));
+
+        let message_callback_id = self.driver.add_message_callback(move |msg| {
+            // Check specifically for ResetRequest
+            if let sys::AsioMessageSelectors::kAsioResetRequest = msg {
+                if let Ok(mut cb) = error_callback_shared.lock() {
+                    cb(StreamError::BackendSpecific {
+                        err: BackendSpecificError {
+                            description: "ASIO reset request.".to_string(),
```
roderickvd: I think it's OK to add a backend-specific `StreamError` variant. Next version is going to be v0.17 anyway, so breaking changes are acceptable.

Sin-tel: I did that, unsure if there should be a `cfg[feature = "asio"]` on that variant specifically, might be  more trouble than it's worth, since you can just handle the `_ => {}` case. I did mark it non-exhaustive, since it is likely there will be more variants added.

asio-sys/src/bindings/mod.rs
Sin-tel: Added in docs for the new `StreamError` variant.

roderickvd:
Having given it some more thought, I think we should call it `StreamInvalidated`: not specifically tied to ASIO and indeed more generally applicable:

```rust
/// The stream configuration is no longer valid and must be rebuilt.
/// This occurs when the backend requires a full reset:
/// - ASIO: Driver reset request (sample rate/format change via control panel)
/// - JACK: Sample rate changed by server
/// - Other backends: Format/configuration invalidated
StreamInvalidated,
```

PR #1059, opened 2025-12-06
WASAPI: Handle AUDCLNT_E_DEVICE_IN_USE 

Sin-tel:
Small fix: changes the backend-specific error from a cryptic "0x8889000A" to "The device is already in use.".

### DIFF
--- a/src/host/wasapi/mod.rs
+++ b/src/host/wasapi/mod.rs
@@ -96,12 +96,8 @@ fn windows_err_to_cpal_err_message<E: ErrDeviceNotAvailable>(
     message: &str,
 ) -> E {
     match e.code() {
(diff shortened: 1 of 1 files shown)


roderickvd:
Hmm. I'm not sure about the custom string. Why not have it also return `DeviceNotAvailable`?

Sin-tel:
Currently docs say that `DeviceNotAvailable` means e.g. the device is literally unplugged. In my use-case I want to display something helpful like "Another application is using this device." to the user.

roderickvd:
You could argue that it's another kind of "not available". If important to discern between the two then we could consider returning a variant, or heck, even consider some kind of `std::io::Error`.

roderickvd:
Thanks, merged!

PR #1062, opened 2025-12-07
refactor!: make SampleRate a u32 alias

roderickvd:
Since forever `SampleRate` has been a newtype `pub struct SampleRate(pub u32)` without offering any further type guarantees or functionality. On the contrary `FrameCount` and `ChannelCount` are just type aliases to some integer type.

This PR changes `SampleRate` to also be a type alias. While breaking, it'll be easier downstream.


roderickvd:
For yet another level in this discussion we could change these types to be `NonZero`.

PR #1074, opened 2025-12-13
ASIO: open control panel

Sin-tel:
Just putting this here for now so people know it is at least possible.

Will have to think about a proper API if we want to include this. (extension traits?)
I would also be fine having this only exposed in asio-sys for now, but you'd need to expose driver internals if you want to use it from cpal.

Also, I had to modify the build script to get rid of linker errors, not sure what's up with that.


roderickvd:
Hey! I think the asio-sys part is worth having regardless of the cpal API discussion. While I believe the ASIO control panels are blocking modals, it seems that PortAudio and JUCE also expose them.

For the build script changes: the linker errors make sense - `ASIOControlPanel()` uses Windows registry APIs (`Advapi32`) and UI functions (`User32`). Just noting that you have `user32` on line 66 and `User32` on line 68 - they're the same lib on Windows (case-insensitive) so it works, just redundant.

Re: the cpal API, I agree we should use extension traits rather than exposing this through `as_inner()` downcasting. Here's what I'm thinking:

```rust
// In src/platform/asio.rs or similar
#[cfg(all(target_os = "windows", feature = "asio"))]
pub trait AsioDeviceExt {
    fn asio_open_control_panel(&self) -> Result<(), BackendSpecificError>;
}

impl AsioDeviceExt for Device {
    fn asio_open_control_panel(&self) -> Result<(), BackendSpecificError> {
        if let DeviceInner::Asio(ref d) = self.as_inner() {
            d.open_control_panel()
        } else {
            // or maybe introduce a new error variant
            Err(BackendSpecificError {
                description: "Not an ASIO device".to_string(),
            })
        }
    }
}
```

Usage would be like:

```rust
use cpal::platform::asio::AsioDeviceExt;

let device = host.default_output_device()?;
device.asio_open_control_panel()?;
```

This keeps the main `Device` trait platform-agnostic while allowing opt-in access to host-specific features. We could do similar for WASAPI exclusive mode, ALSA period counts, etc.

roderickvd:
Had the above extension traits in mind for v0.18 by the way, not v0.17.

Sin-tel:
Cool, trying out the extension trait now. It seems pretty nice! How I did it right now still has the problem with docs not generating though.

examples/asio_panel.rs
roderickvd, line 1: What do you think about making this an example in `asio-sys/examples` instead?

src/platform/asio.rs
```diff
@@ -0,0 +1,24 @@
+use crate::platform::DeviceInner;
+use crate::BackendSpecificError;
+use crate::Device;
+
+pub trait AsioDeviceExt {
```
roderickvd, line 6: I'd like the public API to have good documentation. Proposal:

```rust
/// Extension trait for ASIO-specific device functionality.
```

src/platform/asio.rs
```diff
@@ -0,0 +1,24 @@
...
+    fn asio_open_control_panel(&self) -> Result<(), BackendSpecificError>;
```
roderickvd:
```rust
/// Opens the ASIO driver's control panel window.
///
/// This provides access to device-specific settings like buffer size,
/// sample rate, input/output routing, and hardware-specific features.
///
/// # Blocking Behavior
///
/// **WARNING**: This call blocks until the user closes the control panel.
/// Consider spawning a thread to avoid blocking the main thread.
///
/// # Errors
///
/// Returns an error if this device is not an ASIO device.
```

src/platform/mod.rs
```diff
@@ -10,6 +10,9 @@ pub use self::platform_impl::*;
 #[cfg(feature = "custom")]
 pub use crate::host::custom::{Device as CustomDevice, Host as CustomHost, Stream as CustomStream};
 
+#[cfg(all(target_os = "windows", feature = "asio"))]
```
roderickvd: For cross-platform compatibility in downstream code, the configuration guard should be moved from here into `platform/asio.rs` like so:

```rust
#[cfg(all(target_os = "windows", feature = "asio"))]
impl AsioDeviceExt for Device {
    // ...
}

#[cfg(not(all(target_os = "windows", feature = "asio")))]
impl AsioDeviceExt for Device {
    fn is_asio_device(&self) -> bool {
        false
    }
    
    fn asio_open_control_panel(&self) -> Result<(), BackendSpecificError> {
        Err(not_available())
    }

    fn asio_reset_device(&self) -> Result<(), BackendSpecificError> {
        Err(not_available())
    }
}

fn not_available() -> BackendSpecificError {
    BackendSpecificError {
        description: "ASIO is not available on this platform".to_string(),
    }
}
```

roderickvd (changes requested):
Some considerations inline. Let's pay particular attention to getting the extension traits right. When we start adding them, it'd be highly desirable to keep them consistent and stable for the other hosts and time to come.

Also don't forget about a changelog entry.

roderickvd:
@Sin-tel how's your extension traits experiment coming along? Not to haste - just checking in.
This should become the template for extension traits in other hosts as well.

examples/asio_panel.rs
Sin-tel: The example is a bit useless I think, was kind of intended to just show off that it works at all. With the extension trait it's not really necessary anymore.

Sin-tel:
Got distracted by other things. Mostly done now!

Any idea why windows-crate-v0.61.3 is failing? 

EDIT: Seems like windows runners are randomly failing, I'm blaming github since it's all working fine for me locally.

roderickvd:
@Sin-tel checking in - what's your status on this one?

Sin-tel:
Needs some fixes due to #1107

Sin-tel:
Well it works, but arguably pretty verbose to get the driver now that it is not a field on Device anymore.

asio-sys/build.rs
```diff
@@ -73,6 +73,7 @@ fn main() {
     println!("cargo:rustc-link-lib=dylib=advapi32");
     println!("cargo:rustc-link-lib=dylib=ole32");
     println!("cargo:rustc-link-lib=dylib=user32");
+    println!("cargo:rustc-link-lib=dylib=Advapi32");
```
roderickvd: `advapi32` is also present three lines up in lowercase... are they different?

CHANGELOG.md
```diff
@@ -44,6 +44,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **ALSA**: Example demonstrating ALSA error suppression during enumeration.
 - **ALSA**: Support for native DSD playback.
 - **WASAPI**: Enable as-necessary resampling in the WASAPI server process.
+- **ASIO**: Extension trait for ASIO devices, which allows opening the control panel.
```
roderickvd, line 235: Nitpick but would you keep this list sorted alphabetically, and move it up one line?

src/platform/asio.rs
```diff
@@ -0,0 +1,83 @@
...
+                    description: format!("Failed to load driver: {:?}", e),
+                })?;
+
+            driver
+                .open_control_panel()
+                .map_err(|e| BackendSpecificError {
+                    description: format!("Failed to open control panel: {:?}", e),
+                })
+        } else {
+            Err(BackendSpecificError {
```
roderickvd: Maybe we should use `unreachable` here instead.

src/platform/asio.rs
```diff
@@ -0,0 +1,83 @@
...
+    /// **WARNING**: This call may block until the user closes the control panel.
+    /// Consider spawning a thread to avoid blocking the main thread.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if this device is not an ASIO device.
+    fn asio_open_control_panel(&self) -> Result<(), BackendSpecificError>;
+}
+
+#[cfg(all(target_os = "windows", feature = "asio"))]
```
roderickvd: Let's make it discoverable on docs.rs:
```rust
#[cfg_attr(docsrs, doc(cfg(all(target_os = "windows", feature = "asio"))))]
```

src/platform/asio.rs
```diff
@@ -0,0 +1,83 @@
...
+        if let DeviceInner::Asio(ref asio_device) = self.as_inner() {
+            let description = asio_device
+                .description()
+                .map_err(|e| BackendSpecificError {
+                    description: format!("Could not get device name: {:?}", e),
+                })?;
+
+            let driver = GLOBAL_ASIO
+                .get()
+                .ok_or(BackendSpecificError {
```
roderickvd: Maybe `expect` here instead of returning silently - this should never be possible when we've got a `Device` constructed.

roderickvd (changes requested):
Sorry for my going back and forth as I'm thinking how the extension trait API should look. I think the current proposal isn't scalable when we know we'll also have `is_wasapi_host`, `is_alsa_host`, etc. in the future.

How about this:
```rust
pub trait AsioDeviceExt {
    fn as_asio(&self) -> Option<AsioDevice<'_>>;
}

pub struct AsioDevice<'a>(&'a crate::host::asio::Device);

impl AsioDevice<'_> {
    pub fn open_control_panel(&self) -> Result<(), BackendSpecificError> {
        use crate::host::asio::GLOBAL_ASIO;
        let driver_name = self.0.name();
        GLOBAL_ASIO
            .get()
            .expect("GLOBAL_ASIO is always set when an ASIO Device exists")
            .load_driver(driver_name)
            .map_err(|e| BackendSpecificError { description: format!("{e:?}") })?
            .open_control_panel()
            .map_err(|e| BackendSpecificError { description: format!("{e:?}") })
    }
}

impl AsioDeviceExt for crate::Device {
    fn as_asio(&self) -> Option<AsioDevice<'_>> {
        match self.as_inner() {
            #[cfg(all(target_os = "windows", feature = "asio"))]
            crate::platform::DeviceInner::Asio(d) => Some(AsioDevice(d)),
            _ => None,
        }
    }
}
```

Sin-tel:
Ok so, it took me a bit to figure out the trick here to get it to compile (using `AsioDevice<'a>(std::marker::PhantomData<&'a ()>);`). 

Besides that, I think the new design is clearly better in terms namespacing, error handling, etc.

Sin-tel:
Ok, I moved all cfg gates to be inside of the struct.

Pros: User code does not need any `cfg` at all. Very easy to write clean cross-platform code, and you can just match the option.
Cons: Maybe slightly more annoying to write on the cpal side, but there's really no way to avoid it, unless we gate the entire struct.

A minimal panel example now looks like this:
```rust
fn main() {
    use cpal::HostId;
    use cpal::platform::asio::AsioDeviceExt;
    use cpal::traits::HostTrait;

    #[cfg(all(windows, feature = "asio"))]
    let host = cpal::host_from_id(HostId::Asio).unwrap();
    #[cfg(not(all(windows, feature = "asio")))]

    if let Some(d) = device.as_asio() {
        d.open_control_panel().unwrap();
    } else {
        println!("Not an ASIO device.");
    }

    // Keep the thread alive so the window doesn't close immediately
    std::thread::sleep(std::time::Duration::from_secs(5));
}
```

src/platform/asio.rs
```diff
@@ -0,0 +1,72 @@
...
+
+            let description = self.inner.description().map_err(|e| BackendSpecificError {
+                description: format!("{e:?}"),
+            })?;
+            let driver_name = description.name();
+
+            GLOBAL_ASIO
+                .get()
+                .expect("GLOBAL_ASIO is always set when an ASIO device exists")
+                .load_driver(driver_name)
```
roderickvd: I think you mentioned it before, but why do we need to call `load_driver()` here again? If we've got a `Device` then its driver is loaded, right?

src/platform/asio.rs
```diff
@@ -0,0 +1,72 @@
...
+    /// sample rate, input/output routing, and hardware-specific features.
+    ///
+    /// # Blocking Behavior
+    ///
+    /// This call may block until the user closes the control panel.
+    /// Consider spawning a thread to avoid blocking the main thread.
+    pub fn open_control_panel(&self) -> Result<(), BackendSpecificError> {
+        #[cfg(all(target_os = "windows", feature = "asio"))]
+        {
+            use crate::host::asio::GLOBAL_ASIO;
```
roderickvd: Promoting `GLOBAL_ASIO` to `pub(crate)` just to serve the extension trait seems wrong. Since `stream.rs` already follows the `GLOBAL_ASIO` to `load_driver` pattern, I think `open_control_panel` belongs directly on `asio::Device`, where it can reach `GLOBAL_ASIO` via `super::` without any visibility change. The extension trait then just calls `self.inner.open_control_panel() and `GLOBAL_ASIO` can revert to a `static`.

src/platform/asio.rs
```diff
@@ -0,0 +1,72 @@
+//! Implementations for ASIO-specific device functionality.
+
+#[allow(unused_imports)]
```
roderickvd: For this one and the line below, it'd be more isolated to change it to `#[cfg(not(all(target_os = "windows", feature = "asio")))]`.

roderickvd (changes requested):

src/platform/asio.rs
Sin-tel: Since the ASIO device is just some memo-ized data, it doesn't guarantee the driver is actually loaded.

src/platform/asio.rs
Sin-tel: Good point, I'll just mirror the stream implementation.

Sin-tel:
Minor point but we can also use `.expect()` when getting `GLOBAL_ASIO` in stream creation, for the same reasons.

src/platform/asio.rs
roderickvd: OK, got it.

src/platform/asio.rs
roderickvd: Could you still take a look at this one?

src/host/asio/device.rs
```diff
@@ -152,6 +153,27 @@ impl Device {
...
         configs
     }
+
+    /// Opens the ASIO driver's control panel window.
+    pub fn open_control_panel(&self) -> Result<(), BackendSpecificError> {
+        com::com_initialized();
+        let description = self.description().map_err(|e| BackendSpecificError {
+            description: format!("{e:?}"),
+        })?;
+        let driver_name = description.name();
```
roderickvd: We can just do `&self.name` here.

Sin-tel:
There was some discussion on Discord on whether this needs an extension trait in the first place, because we could also just expose the inner device directly.

No conclusion really, but I'll just defer this decision since I don't really have a strong preference either way. If we decide this is the way forward I'll go ahead and fix conflicts with upstream.

src/platform/asio.rs
Sin-tel: As it stands, I don't think any specific doc attributes are necessary since everything is available on all platforms and features.

roderickvd:
Finally took the time to fully think through the extension API: #1334

PR #1077, opened 2025-12-14
ASIO: fix for FL Studio ASIO quirk

Sin-tel:
It seems like some drivers (notably "FL Studio ASIO") do not use the ASIO double buffering mechanism, and always pass a buffer_index of 0, instead of the usual 0 - 1 swapping.

This means we can not rely on buffer_index to check when to silence the buffer.

`src/host/asio/stream.rs:326`
```rust
let silence = current_buffer_index.load(Ordering::Acquire) != callback_info.buffer_index;
```
This check always fails, so the buffer never gets cleared. Output will only get added, stacking up forever.

I replaced it with a similar check, using a global counter instead that gets incremented once in the buffer_switch.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -68,6 +68,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - **ALSA**: Format selection to probe hardware endianness instead of assuming native byte order.
 - **ASIO**: Handling for `kAsioResetRequest` message to prevent driver UI becoming unresponsive.
+- **ASIO**: Buffer silencing logic to work with non-conformant drivers (e.g., FL Studio ASIO).
 - **CoreAudio**: Timestamp accuracy.
 - **CoreAudio**: Segfaults when enumerating devices.
 - **CoreAudio**: Undefined behavior related to null pointers and aligned reads.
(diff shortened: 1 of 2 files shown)


roderickvd:
Nice catch on this driver quirk!

I was thinking though - since ASIO only supports one driver at a time, could we simplify this to just toggle between 0 and 1 instead of counting up indefinitely?

Something like:

```rust
static CALLBACK_FLAG: AtomicU32 = AtomicU32::new(1);

// In buffer_switch_time_info:
let callback_flag = CALLBACK_FLAG.fetch_xor(1, Ordering::Relaxed);

let callback_info = CallbackInfo {
    buffer_index: double_buffer_index,
    callback_flag,  // Alternates: 1, 0, 1, 0, ...
};
```

Then initialize the device storage to the opposite value:

```rust
current_callback_flag: Arc::new(AtomicU32::new(0)),
```

This would give the same behavior but:
- No wraparound concerns after 4 billion callbacks
- Only two states to reason about
- More explicit that we're just detecting "is this a new buffer switch"

What do you think?

Sin-tel:
Yeah, a flag is simpler overall, but there's still some subtle issue with the initialization since it could be set to 0 or 1 by a previous stream. So I still set it to some other value like `u32::MAX`.

roderickvd:
Thank you! I just added changelog entries and will merge when the CI gives the green light.
Then I think we're good for an asio-sys 0.2.4 and cpal 0.17 release 🎊

PR #1078, opened 2025-12-17
add native dsd support

ljufa:


src/samples_formats.rs
```diff
@@ -286,3 +308,123 @@ impl SizedSample for f32 {
...
+    type Float = f32;
+    type Signed = i8; 
+    
+    const EQUILIBRIUM: Self = DsdU8(0x69); 
+}
+
+// Conflict with blanket impl in dasp_sample
+// impl FromSample<DsdU8> for DsdU8 { ... }
+
+impl FromSample<f32> for DsdU8 {
```
roderickvd: Yeah these are a bit of a pickle. Fully understood that `cpal` doesn't do PCM-to-DSD conversion, but should we then always return silence or an `unimplemented()`? What would it take to skip the implementations at all?

src/host/alsa/mod.rs
```diff
@@ -1429,10 +1447,18 @@ fn set_sw_params_from_format(
         }
         let start_threshold = match stream_type {
             alsa::Direction::Playback => {
-                // Always use 2-period double-buffering: one period playing from hardware, one
-                // period queued in the software buffer. This ensures consistent low latency
-                // regardless of the total buffer size.
-                2 * period
+                // For playback, we want to start only when the buffer is full for DSD content
+                // to avoid underruns. For PCM, we keep the default low-latency behavior.
+                let is_dsd = matches!(
```
roderickvd: This may warrant a separate `fn is_dsd()` helper function along the lines of `is_int`. That latter function may have to be updated to handle DSD types too.

src/host/alsa/mod.rs
```diff
@@ -1429,10 +1447,18 @@ fn set_sw_params_from_format(
         }
         let start_threshold = match stream_type {
             alsa::Direction::Playback => {
-                // Always use 2-period double-buffering: one period playing from hardware, one
-                // period queued in the software buffer. This ensures consistent low latency
-                // regardless of the total buffer size.
-                2 * period
+                // For playback, we want to start only when the buffer is full for DSD content
```
roderickvd: Why is that? Is that always the case for DSD playback, or should DSD playback have its default buffer size overridden? For DSD we should also aim for low latency - without under-runs, of course.

src/samples_formats.rs
```diff
@@ -153,6 +166,9 @@ impl SampleFormat {
             SampleFormat::U64 => u64::BITS,
             SampleFormat::F32 => 32,
             SampleFormat::F64 => 64,
+            SampleFormat::DsdU8 => 8,
```
roderickvd: Or "1" (for 1-bit modulation)?

src/host/alsa/mod.rs
```diff
@@ -1284,6 +1289,9 @@ fn fill_with_equilibrium(buffer: &mut [u8], sample_format: SampleFormat) {
         SampleFormat::U64 => fill_typed!(u64),
         SampleFormat::F32 => fill_typed!(f32),
         SampleFormat::F64 => fill_typed!(f64),
+        SampleFormat::DsdU8 => fill_typed!(u8),
```
roderickvd: This should probably be `fill_typed!(DsdU8)`

ljufa:
> Really nice add. I know it's still a draft, but I took the opportunity to a quick review anyway. Also don't forget to add a changelog entry - worth it 😄

Thank you for you review! I am not alsa expert but I've tested it in my rsplayer repo. 
I hope that change is not too specific for my use case and it will be generaly useful.

src/samples_formats.rs
ljufa: removed structs and from impl, and it seems to work without it :)

src/host/alsa/mod.rs
ljufa: Removed this change as unnecessary. I was playing with pop and noise on dsd stream start/stop...

src/host/alsa/mod.rs
ljufa: removed

src/samples_formats.rs
ljufa: good point, changed to 1

src/host/alsa/mod.rs
ljufa: now when I removed structs I am not sure what to use here. any suggestion?

roderickvd: Something like this?

```rust
const DSD_SILENCE_BYTE: u8 = 0x69;

SampleFormat::DsdU8 | SampleFormat::DsdU16 | SampleFormat::DsdU32 => buffer.fill(DSD_SILENCE_BYTE),
```

src/samples_formats.rs
```diff
@@ -153,6 +165,9 @@ impl SampleFormat {
             SampleFormat::U64 => u64::BITS,
             SampleFormat::F32 => 32,
             SampleFormat::F64 => 64,
+            SampleFormat::DsdU8 => 1,
```
roderickvd: As a matter of style we could reduce this to `SampleFormat::DsdU8 | SampleFormat::DsdU16 | SampleFormat::DsdU32 => 1`.

src/samples_formats.rs
```diff
@@ -208,6 +226,9 @@ impl Display for SampleFormat {
             SampleFormat::U64 => "u64",
             SampleFormat::F32 => "f32",
             SampleFormat::F64 => "f64",
+            SampleFormat::DsdU8 => "DsdU8",
```
roderickvd: For consistency, change into lowercase `dsdu8`.

src/samples_formats.rs
```diff
@@ -105,6 +105,15 @@ pub enum SampleFormat {
 
     /// `f64` with a valid range of `-1.0..=1.0` with `0.0` being the origin.
     F64,
+
+    /// DSD stream (U8)
```
roderickvd: For consistency I suggest `/// DSD 1-bit stream in u8 container (8 samples per byte) with `0x69` being the silence pattern.`

ljufa:
All suggestions applied. Thank you for your patience!

src/samples_formats.rs
```diff
@@ -181,6 +194,9 @@ impl SampleFormat {
                 | SampleFormat::U32
                 // | SampleFormat::U48
                 | SampleFormat::U64
+                | SampleFormat::DsdU8
```
roderickvd: DSD is not unsigned integer; these lines should be removed here.

src/samples_formats.rs
```diff
@@ -105,6 +105,15 @@ pub enum SampleFormat {
 
     /// `f64` with a valid range of `-1.0..=1.0` with `0.0` being the origin.
     F64,
+
+    /// DSD 1-bit stream in u8 container (8 samples per byte) with 0x69 being the silence pattern.
+    DsdU8,
+
+    /// DSD 1-bit stream in u16 container (8 samples per byte) with 0x69 being the silence pattern.
```
roderickvd: I know I suggested this before but how about:

"DSD 1-bit stream in u16 container (16 bits = 16 DSD samples) with 0x69 being the silence byte pattern."

CHANGELOG.md
```diff
@@ -36,6 +36,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **WASAPI**: `Send` and `Sync` implementations to `Stream`.
 - **WebAudio**: `Send` and `Sync` implementations to `Stream`.
 - **WebAudio**: `BufferSize::Fixed` validation against supported range.
+- **ALSA**: Add support native DSD playback
```
roderickvd: Add support **for** native DSD playback

roderickvd:
Thank you, merged!

Issue #1042, opened 2025-11-12
Allow querying the buffer size of a Stream


will3942:
On Android, amongst other platforms, when using the default buffer size the platform sets the buffer size, this would be useful to query at runtime.

Context on PR - https://github.com/RustAudio/cpal/pull/1039

roderickvd:
Agreed. #964 did some work on that. Maybe we could have both: buffer size _and_ latency, if supported.
I was thinking to work on this onto v0.18 (with v0.17 currently in release cycle). Happy to receive contributions.

edwloef:
I wonder whether the default buffer size could be put in `Device::default_{input, output}_config`, like the device's preferred sample rate already is? Currently it returns a (fairly useless, on `pipewire-alsa` at least) `SupportedStreamConfig`, in the sense that the buffer size it provides is just a range from 1 to 4194304.

roderickvd:
Most hosts that cross my mind actually don't have a way of returning their default buffer size. You may only query it after you've created a stream.

roderickvd:
> Most hosts that cross my mind actually don't have a way of returning their default buffer size. You may only query it after you've created a stream.

After some more investigation, I believe that CoreAudio, AAudio and JACK could return a default buffer size *before* trying to initialize a stream. ALSA & WASAPI can't. ASIO maybe. 3 or 4 out of 6... that's right on the edge of whether it's cross-platform yes/no.

I'll be pushing a PR soon to address _this_ issue of querying the buffer size of a stream. With a bit of a workaround that'd enable a user to try and recreate a stream, if the results of `BufferSize::Default` aren't to his or her liking. _If_ in the future the need for querying ahead of stream creation still comes up, we can consider.

roderickvd:
PR up at #1080

roderickvd:
#1080 merged.

Issue #942, opened 2025-01-25
Feature Request: read default buffer size from `Stream`, and/or query device's preferred buffer size


j-n-f:
With ASIO the driver will report a preferred buffer size, and I'm guessing this is what gets used when the default buffer size is chosen. It'd be nice to know what this value is before starting a stream.

Alternatively, one can use `BufferSize::Fixed`, but would need a way to query what that preferred buffer size is for some `Host` and `Device`.

Something like:

```rust
enum PreferredBufferSize {
    /// Device doesn't provide this information
    Unknown,
    /// Device can provide this information
    Fixed(usize),
}

let preferred_input_size = device.preferred_input_buffer_size();
```

j-n-f:
Small correction: it seems like the "default" buffer size is just whatever buffer size the device is already set to. This may be device/driver specific behaviour.

What I'm looking for is the buffer size the device reports as optimal, where the driver doesn't have to split/combine blocks (i.e. usually the same as the device's internal buffer size).

roderickvd:
Querying stream buffer sizes I'd like to consolidate to issue #1042.

> What I'm looking for is the buffer size the device reports as optimal, where the driver doesn't have to split/combine blocks (i.e. usually the same as the device's internal buffer size).

Best take would the device default buffer size. In v0.17 I overhauled the buffer size system to move from _cpal_ defaults to real _device_ defaults. So when you use `BufferSize::Default`, you get that "device's internal buffer size" you speak of.

roderickvd:
PR up at #1080 for reading default buffer size from `Stream`.

PR #1080, opened 2025-12-23
feat: add Stream::buffer_size() and improve AAudio buffer configuration

roderickvd:
Add `buffer_size()` method to `Stream` trait that returns the number of frames passed to each data callback invocation (actual size or upper limit depending on platform).

AAudio improvements:
- `BufferSize::Default` now explicitly configures using optimal burst size from AudioManager, following Android low-latency audio best practices
- `buffer_size()` query falls back to burst size if `frames_per_data_callback` was not explicitly set
- Refactored buffer configuration to eliminate code duplication

Addresses #1042
Relates to #964, #942

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `DeviceBusy` error variant to `SupportedStreamConfigsError`, `DefaultStreamConfigError`, and
   `BuildStreamError` for retryable device access errors (EBUSY, EAGAIN).
 - `StreamConfig` now implements `Copy`.
+- `StreamTrait::buffer_size()` to query the stream's current buffer size in frames per callback.
 - **PipeWire**: New host for Linux and some BSDs using the PipeWire API.
 - **PulseAudio**: New host for Linux and some BSDs using the PulseAudio API.
 
@@ -26,6 +27,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   direction-mismatched devices (e.g. querying input configs on an output-only device) instead of
   silently returning an empty list.
 - **AAudio**: Buffers with default sizes are now dynamically tuned.
+- **ALSA**: Device disconnection now stops the stream with `StreamError::DeviceNotAvailable` instead of looping.
+- **ALSA**: Polling errors trigger underrun recovery instead of looping.
+- **ALSA**: Try to resume from hardware after a system suspend.
 - **ASIO**: `Device::driver`, `asio_streams`, and `current_callback_flag` are no longer `pub`.
 
 ### Fixed
--- a/src/host/aaudio/mod.rs
+++ b/src/host/aaudio/mod.rs
@@ -291,8 +291,10 @@ fn configure_for_device(
     // - https://developer.android.com/ndk/reference/group/audio#aaudiostreambuilder_setframesperdatacallback
     // - https://developer.android.com/ndk/guides/audio/audio-latency#buffer-size
     if let BufferSize::Fixed(size) = config.buffer_size {
-        // Only for fixed sizes, the user explicitly wants control over the callback size.
-        builder = builder.frames_per_data_callback(size as i32);
+        // For fixed sizes, the user explicitly wants control over the callback size.
+        builder = builder
+            .frames_per_data_callback(size as i32)
+            .buffer_capacity_in_frames(2 * size as i32);
     }
 
     builder
@@ -440,7 +442,7 @@ where
     tuning.capacity.store(capacity, Ordering::Relaxed);
 
     let mixer_bursts = match AudioManager::get_mixer_bursts() {
-        Ok(bursts) => bursts,
+        Ok(bursts) => bursts.max(0),
         Err(_) => {
             let burst_size = stream.frames_per_burst();
             if burst_size > 0 {
--- a/src/traits.rs
+++ b/src/traits.rs
@@ -307,9 +307,9 @@ pub trait StreamTrait {
     ///
     /// Returns the platform's best estimate of the number of frames per callback.
     ///
-    /// - [`BufferSize::Fixed`]: the actual callback size after hardware negotiation, which may
+    /// - [`crate::BufferSize::Fixed`]: the actual callback size after hardware negotiation, which may
     ///   differ from the requested value due to hardware constraints.
-    /// - [`BufferSize::Default`]: the system-configured callback size (e.g. ALSA period,
+    /// - [`crate::BufferSize::Default`]: the system-configured callback size (e.g. ALSA period,
     ///   JACK buffer size, AAudio burst size). This reflects the typical callback size, not a
     ///   guaranteed upper bound.
     ///


roderickvd:
@will3942 @marcpabst @j-n-f as you've requested or contributed to similar in the past, would you give your view?

pgaskin:
Note that the burst size can be smaller than the actual buffer size, and on some devices (e.g., the Pixel 9), it can be low enough that using it will cause underruns in many cases.

c.f. [cmus/cmus#1386](https://github.com/cmus/cmus/pull/1386/changes/56f21a455bc611d879ad5384ca26c154bd59a34d)

roderickvd:
> Note that the burst size can be smaller than the actual buffer size, and on some devices (e.g., the Pixel 9), it can be low enough that using it will cause underruns in many cases.
> 
> c.f. [cmus/cmus#1386](https://github.com/cmus/cmus/pull/1386/changes/56f21a455bc611d879ad5384ca26c154bd59a34d)

Thanks for that nudge. With this default buffer thing it seems that no matter how you try do it right, you never do it right.

While I can see a 80 ms minimum working for many scenarios, setting minimums is precisely what we've been trying to steer away from since v0.17 because in the end, they are rather arbitrary and use-cases dependent.

It seems that the best thing to do would be to dynamically optimizing the buffer by monitoring underruns like here: https://developer.android.com/ndk/guides/audio/aaudio/aaudio#tuning-buffers.

I'll save that for another PR - I think that this PR won't make things on Pixel 9 better or worse from what we already had. Can you confirm?

will3942:
@roderickvd Will this set the same buffer size as AAudio would default to if we don't specify a buffer size (as in the latest master)? If so then I'm happy, but I can't test until 12 Jan.

Agreed that we need to adjust in response to underruns, or give the developer the ability to. At the moment we've measured an acceptable buffer size for our application in production and set that manually.

roderickvd:
> @roderickvd Will this set the same buffer size as AAudio would default to if we don't specify a buffer size (as in the latest master)? If so then I'm happy, but I can't test until 12 Jan.

I'd also need someone to confirm for me.

pgaskin:
> Will this set the same buffer size as AAudio would default to if we don't specify a buffer size

Not entirely sure. In cmus, I ended up taking a shortcut and just choosing a relative high buffer size based on a fixed period to reduce CPU usage and to avoid underruns (especially as newer devices are capable of lower latency streams) since I wasn't concerned about latency.

When testing AAudio with cmus, the default buffer size on the 9 was 3 times the burst size, but on the 8, it was 2 times the burst size.

I haven't fully followed how it's set in AOSP, but some random thoughts:

- In [AAudioServiceEndpointPlay](https://cs.android.com/android/platform/superproject/main/+/main:frameworks/av/services/oboeservice/AAudioServiceEndpointPlay.cpp;l=43-64;drc=1de7952ebda8865adc3d398b3dc2ffa5ad100593), it defaults to getAAudioMixerBurstCount or hardcoded of 2. This is multiplied by the burst size from the HAL to set the buffer size.
- In [AudioSystem](https://cs.android.com/android/platform/superproject/+/android-latest-release:frameworks/av/media/libaudioclient/AudioSystem.cpp;l=2933-2937;drc=d3193bf8465fdbdaef80e983a81e18d9df21b689), getAAudioMixerBurstCount ends up calling [getAAudioMixerBurstCountFromSystemProperty](https://cs.android.com/android/platform/superproject/main/+/main:frameworks/av/services/audioflinger/afutils/PropertyUtils.cpp;l=80-90;drc=1798b12a5202c77785d34be268ef8ab4bc1013cf), which uses `aaudio.mixer_bursts` (set in the vendor props on some devices) or defaults to 2.
- Although not directly relevant, the logic used in the OpenSLES [implementation](https://cs.android.com/android/platform/superproject/+/android-latest-release:external/oboe/src/opensles/AudioStreamOpenSLES.cpp;l=159-224;drc=0b20d04392c3a2c271df07cb25d5ad3c2db391ac) might be interesting to look at.

You get the value of `PROPERTY_OUTPUT_FRAMES_PER_BUFFER`, which comes from [getPrimaryOutputFrameCount](https://cs.android.com/android/platform/superproject/+/android-latest-release:frameworks/av/services/audioflinger/AudioFlinger.cpp;l=2632-2637;drc=f491cc6f1a6f9dc4fdf68bb2b424bb33e86f9eae). You can see the HAL frame count in `dumpsys media.audio_flinger` under `HAL frame count`. I think a default of 256 is a bit low, as my devices are all at least double.

roderickvd:
> > Will this set the same buffer size as AAudio would default to if we don't specify a buffer size
> 
> Not entirely sure.

You may be right. Your pushback makes me think that we probably cannot reliably make assumptions about what devices do when we set nothing at all.

> When testing AAudio with cmus, the default buffer size on the 9 was 3 times the burst size, but on the 8, it was 2 times the burst size.
>
> I haven't fully followed how it's set in AOSP, but some random thoughts: [...]

These are very valuable resources, thank you very much for pointing to these lines so specifically. Key pattern seems that nobody actually calls `setFramesPerDataCallback` at all - they only set the buffer capacity to `mixer_bursts * framesPerBurst` and let AAudio decide the optimal callback size.

Google seems to set a large capacity but start with minimal active size, then grow it on underruns. So our current implementation is likely over-constraining AAudio by explicitly setting the callback size when we should just be setting the capacity and staying out of the way.

This seems to be confirmed over at https://developer.android.com/ndk/reference/group/audio:

> For the lowest possible latency, do not call this function. AAudio will then call the dataProc callback function with whatever size is optimal. That size may vary from one callback to another.
>
> Only use this function if the application requires a specific number of frames for processing. The application might, for example, be using an FFT that requires a specific power-of-two sized buffer.

So I implemented this equivalently in [5656daa](https://github.com/RustAudio/cpal/pull/1080/commits/5656daa450105ad234d96195e163640e0f6fe96c), what do you think?

On top of this, we'd still want the dynamic buffer tuning.

> I think a default of 256 is a bit low, as my devices are all at least double.

Yeah, I know. I took it from https://developer.android.com/ndk/guides/audio/audio-latency#buffer-size. Oboe sets 16 as floor.

roderickvd:
Now that I'm working on a tuning implementation, I'm realizing that we should be setting buffer _size_ instead of _capacity_.

roderickvd:
> @roderickvd Will this set the same buffer size as AAudio would default to if we don't specify a buffer size (as in the latest master)? If so then I'm happy, but I can't test until 12 Jan.

@will3942 would you be able to report how it goes for you?

will3942:
@roderickvd Apologies for the delay - I haven't been able to get this setup on our stack. I will aim to do that this week. With that said having reviewed these changes it seems as though even if the developer specifies a fixed buffer size we still tune these dynamically? I would think we would want the capability to let a developer avoid that?

roderickvd:
@will3942 

> @roderickvd Apologies for the delay - I haven't been able to get this setup on our stack. I will aim to do that this week. 

No worries, let me know when you have.

> With that said having reviewed these changes it seems as though even if the developer specifies a fixed buffer size we still tune these dynamically? I would think we would want the capability to let a developer avoid that?

You are right. If you could let me know if it works at all, then I'll update accordingly after.

Issue #634, opened 2022-01-17
Enumerating ALSA devices leaves device open, preventing opening other devices


nyanpasu64:
As a result of https://github.com/RustAudio/cpal/pull/506/files, when iterating through `host.{devices,input_devices,output_devices}()`, each `Device` holds its corresponding `alsa::PCM`/`snd_pcm_t`s open until dropped. This causes problems in practice:

- ALSA often won't let you `snd_pcm_open` more than one hw device at once (or two if one is USB audio), returning error `EBUSY`. Plug devices are unaffected, and sysdefault (no clue, never used it, is it dmix?) is somewhere in between.
	- As a result, if you collect input devices into a Vec/etc. instead of iterating them one by one (dropping before fetching the next item), all but 1-2 hw device is missing entirely. Holding a GPU device open prevents you from opening motherboard audio.
- When you call `host.input_devices()`, every returned device is opened for both input and output. If you keep exactly one input device in-scope ("front:CARD=Generic,DEV=0" is motherboard audio) and call `host.output_devices()`, that device disappears from the list, since trying to open it for output returns `EBUSY` and causes cpal to skip over them. Additionally, some but not all devices disappear from the list (other motherboard channels vanish, GPU and USB audio can still be opened).
	- Should cpal special-case `EBUSY` and set an "already opened" flag (or raise an error when you actually create a stream) instead of making the device disappear?

I created a debug branch which modifies `examples/feedback.rs` to print the results of (iterating over all devices, collecting into a Vec, and collecting a second time), before and after opening "front:CARD=Generic,DEV=0" as input: https://github.com/RustAudio/cpal/compare/master...nyanpasu64:test-device-open

- Results at https://gist.github.com/nyanpasu64/308a912f6b81c00e72140ed375765f54

Alternatively, https://github.com/RustAudio/cpal/compare/master...nyanpasu64:log-device-open adds noisy verbose printing to `cpal::host::alsa::DeviceHandles::try_open()` (which is called both by `host.devices()` and by `host.input_devices()` calling `supported_input_configs` -> `supported_configs` -> `get_mut`). Unless I made a mistake scanning over the results, all failed opens which initially succeeded failed due to `EBUSY`.

- Results at https://gist.github.com/nyanpasu64/b1798d72d78a6f0705ba7c0241f1d01e

## Suggestions

[...]

Issue #615, opened 2021-11-15
alsa: Calling Host::input_devices() breaks Host::output_devices(), and vice versa


ssssam:
To reproduce: find a device name that supports capture and playback (you can use `aplay -L` and `arecord -L`), then pass it as poth input and output to `feedback` example:

```
> cargo run --example feedback sysdefault:CARD=sofhdadsp sysdefault:CARD=sofhdadsp
thread 'main' panicked at 'failed to find output device', cpal/examples/feedback.rs:119:6
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Note the device name is same input and output, and it exists when we call `input_devices()` but not when we call `output_devices()`. The cause: the second time we iterate Devices, the code calls `DeviceHandles::open()` on the device and returns `EBUSY`, so the iterator doesn't return the device.

The code in feedback.rs has not actually opened the device, so perhaps the iterator isn't correctly freeing the handle after opening it.

ssssam:
I am struggling to find how to work around or fix this problem. 

Options I've thought of:

  * Add duplex API, as in: https://github.com/RustAudio/cpal/pull/553
  * Add a 'stream_type' hint to [Devices](https://docs.rs/cpal/0.13.4/cpal/platform/struct.Devices.html), use it to avoid opening playback handle when we are querying capture devices

roderickvd:
This is due to #634, merging it with that one.

Issue #384, opened 2020-03-31
unexpected print when enumerate the devices


bitbegin:
```
ALSA lib setup.c:547:(add_elem) Cannot obtain info for CTL elem (MIXER,'AC97 2ch->4ch Copy Switch',0,0,0): No such file or directory
ALSA lib setup.c:547:(add_elem) Cannot obtain info for CTL elem (MIXER,'AC97 2ch->4ch Copy Switch',0,0,0): No such file or directory
ALSA lib setup.c:547:(add_elem) Cannot obtain info for CTL elem (PCM,'IEC958 Playback PCM Stream',0,0,0): No such file or directory
ALSA lib setup.c:547:(add_elem) Cannot obtain info for CTL elem (PCM,'IEC958 Playback PCM Stream',0,0,0): No such file or directory
ALSA lib pcm_dmix.c:990:(snd_pcm_dmix_open) The dmix plugin supports only playback stream
  5. "dmix:CARD=AudioPCI,DEV=0"
ALSA lib pcm_dmix.c:990:(snd_pcm_dmix_open) The dmix plugin supports only playback stream
ALSA lib pcm_dmix.c:990:(snd_pcm_dmix_open) The dmix plugin supports only playback stream
Error: InvalidArgument
ALSA lib pcm_dmix.c:990:(snd_pcm_dmix_open) The dmix plugin supports only playback stream
  6. "dmix:CARD=AudioPCI,DEV=1"
ALSA lib pcm_dmix.c:990:(snd_pcm_dmix_open) The dmix plugin supports only playback stream
ALSA lib pcm_dmix.c:990:(snd_pcm_dmix_open) The dmix plugin supports only playback stream
Error: InvalidArgument
ALSA lib pcm_dsnoop.c:556:(snd_pcm_dsnoop_open) The dsnoop plugin supports only capture stream
  7. "dsnoop:CARD=AudioPCI,DEV=0"
    Default input stream config:
      SupportedStreamConfig { channels: 2, sample_rate: SampleRate(48000), sample_format: I16 }
    All supported input stream configs:
      7.1. SupportedStreamConfigRange { channels: 2, min_sample_rate: SampleRate(48000), max_sample_rate: SampleRate(48000), sample_format: I16 }
ALSA lib pcm_dsnoop.c:556:(snd_pcm_dsnoop_open) The dsnoop plugin supports only capture stream
[...]

PR #1085, opened 2026-01-01
alsa: update to alsa-rs 0.11 and fix device access issues

roderickvd:
This PR updates `alsa-rs` to 0.11 and fixes several related issues around device enumeration, duplex device queries, and ALSA's `stderr` spam during enumeration. This also includes some quality-of-life improvements like `Debug` derives and better error handling.

## Reverts device handle caching

Previously, we cached device handles to avoid opening them repeatedly (#506). This seemed efficient, and fixed drivers that have issues with rapidly closing and opening, but it caused two real problems:
- When you tried to query both input and output configs on a duplex device, the second query would fail with `EBUSY` because we already had it open
- Other applications couldn't access devices while we were enumerating them

I've reverted the caching: devices are now opened fresh for each operation. For broken drivers where rapid open/close causes issues (like some NVIDIA HDMI cards), applications can use the new `DeviceBusy` error variant to implement retry logic.

Fixes:
- #615
- #634

## Suppresses ALSA's `stderr` spam

ALSA has the habit of printing error messages directly to `stderr` during normal device enumeration. Users were understandably confused by this. The `enumerate.rs` example now shows how to use the new `alsa::Output::local_error_handler()` to suppress these messages during enumeration.

Fixes:
- #384

## Manages ALSA's global configuration

ALSA maintains a process-global configuration cache that we were never properly initializing or cleaning up. This led to memory leaks that showed up in Valgrind - somewhat false positives, but still. 

Now we call `alsa::config::update()` when creating the first `Host`, and `alsa::config::update_free_global()` when dropping the last one, after the last `Device` and `Stream` are dropped as well. The former reduces Valgrind reports, the latter correctly throws an error when ALSA is configured incorrectly.

## Handling of retriable errors with `DeviceBusy`

I've added a new `DeviceBusy` error variant to several error types. This lets applications distinguish between retriable errors (like `EBUSY` or `EAGAIN`) and permanent failures (like `ENOENT` or `EPERM`). This allows users to implement smart retry logic when it makes sense, and counter the rapid open/close issues described above.

## `Debug` implementations

I've added `Debug` derives to `Host`, `Device`, `Stream`, and internal types.

## Breaking changes

This will require a semver minor update to v0.18:

- MSRV has been updated to 1.78 globally, and to 1.82 for ALSA specifically.

- The new `DeviceBusy` error variant is technically breaking because the error enums weren't marked as `#[non_exhaustive]`. Users with match statements that don't have a catch-all pattern will get compilation errors.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,31 +5,42 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [0.17.1] - 2026-01-04
+## [Unreleased]
 
 ### Added
 
 - `DeviceBusy` error variant for retriable device access errors (EBUSY, EAGAIN).
-- **ALSA**: `Default` implementation for `Device` (returns the ALSA "default" device).
 - **ALSA**: `Debug` implementations for `Host`, `Device`, `Stream`, and internal types.
 - **ALSA**: Example demonstrating ALSA error suppression during enumeration.
-- **CI**: Checks default/no-default/all feature sets with platform-dependent MSRV for JACK.
-
-### Fixed
-
-- **ALSA**: Device enumeration now includes both hints and physical cards.
-- **ALSA**: Enumerating input and output devices no longer interferes with each other.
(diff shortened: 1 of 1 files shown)


roderickvd:
Merging this to give it greater testing exposure. Targeted for v0.18.

PR #1086, opened 2026-01-03
Fix wasm-bindgen break

kettle11:
Without this line when even just importing cpal types while using wasm-bindgen (for either Wasm backend) wasm-bindgen would fail with this cryptic error:

`thread 'main' (37303789) panicked at /Users/ian/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/wasm-bindgen-cli-support-0.2.106/src/descriptor.rs:200:15:
index out of bounds: the len is 0 but the index is 0`

This change fixes that break and both Wasm backends should work correctly again.

The above break went in before the latest Cpal release, so theoretically anyone using Cpal on Wasm who updates to the latest release will hit the above error. It may be worthwhile to yank the previous release because it's a subtle and difficult error to track down.

It seems like `wasm-bindgen` could also do better here to surface a more meaningful error, and it may be worthwhile for cpal to consider running wasm-bindgen in the CI just to make sure it works.

### DIFF
--- a/.github/workflows/platforms.yml
+++ b/.github/workflows/platforms.yml
@@ -417,6 +417,8 @@ jobs:
       - name: Build wasm-beep example
         working-directory: ./examples/wasm-beep
         run: trunk build
+        env:
+          RUSTUP_TOOLCHAIN: ${{ env.MSRV_WASM }}
 
   # WebAssembly - AudioWorklet
   wasm-audioworklet:
@@ -449,6 +451,8 @@ jobs:
       - name: Build audioworklet-beep example
         working-directory: ./examples/audioworklet-beep
         run: trunk build
+        env:
+          RUSTUP_TOOLCHAIN: nightly
 
   # WebAssembly - WASI Preview 1
   wasm-wasip1:


roderickvd:
Thank you. The CI breakage on Linux and macOS is due to an unrelated error that I'll fix now.
Would you also add a change entry, and the CI check that you're suggesting?

Yanking v0.17 may be too strong when the other platforms are fine, but we can release a new version shortly.

kettle11:
I updated the change log and made it so the CI builds both Wasm examples with trunk, which correctly finds the right version of wasm-bindgen to run against the examples.

I also think it's still worth considering yanking 0.17.0 after 0.17.1 is released, because if someone is unlucky enough to hit this issue it is hard to track down and cpal does have a lot of Wasm users.

roderickvd:
> I updated the change log and made it so the CI builds both Wasm examples with trunk, which correctly finds the right version of wasm-bindgen to run against the examples.

Sounds good. Does `trunk` also use the right Rust toolchain? I don't know it and just notice that it doesn't have `+${{ env.MSRV_WASM }}` anymore.

kettle11:
That's a good catch on the MSRV. I think trunk might have selected a different version of Rust from MSRV_WASM without that, so I've reintroduced it.

PR #1088, opened 2026-01-03
Fix AudioWorklet example

kettle11:
This fixes the audioworklet example for the breaking change recently introduced by: https://github.com/rust-lang/rust/pull/147225

Even though many projects are now using workers and threading on web it's still dependent on nightly Rust and so breaking changes can occur. This fixes the example so it doesn't crash when run.


roderickvd:
Thanks for your continued maintenance of this!

PR #1094, opened 2026-01-07
edition 2024

Decodetalkers:
because pipewire need it, so I want to upgrade it

### DIFF
--- a/.github/workflows/platforms.yml
+++ b/.github/workflows/platforms.yml
@@ -24,13 +24,13 @@ on:
 
 env:
   # MSRV varies by backend due to platform-specific dependencies
-  MSRV_AAUDIO: "1.82"
-  MSRV_ALSA: "1.82"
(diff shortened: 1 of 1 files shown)


roderickvd:
I'd like to keep cpal down to Rust 2021 for now, so we can keep the MSRV reasonably low.
For the `pipewire` crate, it should suffice to introduce a `MSRV_PIPEWIRE` and set it to 1.85.

Decodetalkers:
> I'd like to keep cpal down to Rust 2021 for now, so we can keep the MSRV reasonably low.
> For the `pipewire` crate, it should suffice to introduce a `MSRV_PIPEWIRE` and set it to 1.85.

But we cannot set the edition into two versions according to the platform....

The min requirement for 2024 is rust 1.85

Ohh.. I see! Not edition 2024 is needed. That report is just the version of the compiler was too low! Thank you!

Issue #628, opened 2022-01-09
Plan for implementing an ALSA duplex API


nyanpasu64:
This is a draft.

## Overview and speculation

I'm told that JACK clients are fed input and output buffers synchronously, by jackd (the audio server), and that JACK's application-facing API abstracts away buffer size management from the app, and instead jackd (the server) handles routing and hardware IO/buffering. ALSA clients are not like that. You open independent input and output streams, and you have to align their block sizes, sampling rates, open them at the same time, read and write the same amount from both streams, etc. I hear that Apple's Core Audio exposes a JACK-like synchronous duplex API that communicates with physical hardware. (On Linux, you can have JACK's synchronous buffers, ALSA's direct hardware access, or neither when an ALSA app talks to pulseaudio-alsa or pipewire-alsa.)

On Linux, I get the impression that the only apps designed to be routable in a graph are JACK apps. PipeWire lets you route the inputs and outputs of Pulse/ALSA apps arbitrarily as well (in a patchbay app), but the apps were not written with this in mind. Worse, in ALSA's case, the application-facing API was written around timing being determined by hardware in real time, and the app managing data buffering itself. As a result, I *think* ALSA duplex can achieve the same round-trip latency on physical hardware (from hardware line in to speaker out) as a JACK client, but I'd be surprised if you can chain 3 ALSA duplex apps in a PipeWire graph and not get 1-2 periods of added latency per app, whereas 3 JACK duplex apps on pipewire-jack (see below) add zero latency compared to 1 app.

![Helvum showing three cpal duplex clients chained in a row, in the PipeWire graph](https://user-images.githubusercontent.com/913957/148678611-7a5c67c2-afde-4c80-b756-2b2de3bc3ce5.png)

jackd never changes buffer/period sizes. pipewire changes buffer/period sizes when you open and close apps. I'm not sure if/how it changes the period size of an ALSA device, but it seems buggy. Canberra notification sounds set the period to 8192 samples (absurdly high latency) after they start playing, there is/was audio glitches when periods get longer (https://gitlab.freedesktop.org/pipewire/pipewire/-/issues/1436 ?), and the round-trip latency of `jack_iodelay` fed through speaker output, a physical aux cable, and line-in input can change (when input/output period sizes diverge? upon xrun?).

[...]

PR #1096, opened 2026-01-17
Duplex support

### LINKED ISSUE #349: Duplex Stream Support
This is a tracking issue for supporting duplex streams. Duplex streams are streams that have device-synchronised input and output, an essential requirement for many real-time and pro-audio applications.

This is a follow-up to #116 but focused specifically on duplex support.

gulbrand:
**UPDATE on AI Usage** 

The [Rust Audio AI policy](https://rust.audio/community/ai/) has been updated and this PR is compliant with the policy.

  # Add synchronized duplex stream support

  ## Summary

  This PR introduces synchronized duplex streams to cpal, starting with CoreAudio support on macOS.

 ## Development Note

  Developed with assistance from Claude Code (Anthropic's AI coding assistant).

  ## Motivation

  Currently, applications requiring synchronized input/output (like DAWs, real-time effects, or audio analysis tools) must use separate input and output streams with ring buffers for synchronization. This approach:
  - Adds latency due to buffering
  - Requires manual synchronization logic
  - Can experience drift between input and output clocks
  - Is more complex to implement correctly

  Duplex streams solve this by using a single device context for both input and output, guaranteeing sample-accurate alignment at the hardware level.

  ## API Overview

  ```rust
  use cpal::duplex::DuplexStreamConfig;

  let device = host.default_output_device()?;


  let config = DuplexStreamConfig {
      input_channels: opt.input_channels,
      output_channels: opt.output_channels,
      sample_rate: opt.sample_rate,
      buffer_size: BufferSize::Fixed(opt.buffer_size),
  };
  let stream = device.build_duplex_stream::<f32, _, _>(
      &config,
      |input, output, info| {
          // Process audio with guaranteed synchronization
          output.copy_from_slice(input);
      },
      |err| eprintln!("Stream error: {}", err),
      None,
  )?;

  stream.play()?;
```
 
## Potentially Breaking Changes

build_duplex_stream and build_duplex_stream_raw have been added to `DeviceTrait` with a default impl. This shouldn't break any existing implementations, but just calling this out.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **ALSA**: `device_by_id` now accepts PCM shorthand names such as `hw:0,0` and `plughw:foo`.
 - **PipeWire**: New host for Linux and some BSDs using the PipeWire API.
 - **PulseAudio**: New host for Linux and some BSDs using the PulseAudio API.
+- `DeviceTrait::build_duplex_stream` and `build_duplex_stream_raw` for synchronized input/output.
+- `duplex` module with `DuplexStreamConfig` and `DuplexCallbackInfo` types.
+- **CoreAudio**: Duplex stream support with hardware-synchronized input/output.
+- Example `duplex_feedback` demonstrating duplex stream usage.
 
 ### Changed
 
@@ -91,6 +95,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- **POTENTIALLY BREAKING**: `DeviceTrait` now includes `build_duplex_stream()` and `build_duplex_stream_raw()` methods. The default implementation returns `StreamConfigNotSupported`, so external implementations are compatible without changes.
 - Bump overall MSRV to 1.78.
 - **ALSA**: Update `alsa` dependency to 0.11.
 - **ALSA**: Bump MSRV to 1.82.
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ This library currently supports the following:
 - Enumerate known supported input and output stream formats for a device.
 - Get the current default input and output stream formats for a device.
 - Build and run input and output PCM streams on a chosen device with a given stream format.
-- Build and run duplex (simultaneous input/output) streams with hardware clock synchronization (macOS only, more platforms coming soon).
+- Build and run duplex (simultaneous input/output) streams with hardware clock synchronization.
 
 Currently, supported platforms include:
 
--- a/README.md
+++ b/README.md
@@ -30,6 +30,7 @@ This library currently supports the following:
 - Enumerate known supported input and output stream formats for a device.
 - Get the current default input and output stream formats for a device.
 - Build and run input and output PCM streams on a chosen device with a given stream format.
+- Build and run duplex (simultaneous input/output) streams with hardware clock synchronization (macOS only, more platforms coming soon).
 
 Currently, supported platforms include:
 
@@ -174,7 +175,7 @@ If you are unable to build the library:
 
 ## Examples
 
-CPAL comes with several examples in `examples/`.
+CPAL comes with several examples in `examples/`, including `duplex_feedback` for hardware-synchronized duplex streams.
 
 Run an example with:
 ```bash
--- a/examples/duplex_feedback.rs
+++ b/examples/duplex_feedback.rs
@@ -1,0 +1,104 @@
+//! Feeds back the input stream directly into the output stream using a duplex stream.
+//!
+//! Unlike the `feedback.rs` example which uses separate input/output streams with a ring buffer,
+//! duplex streams provide hardware-synchronized input/output without additional buffering.
+//!
+//! Note: Currently only supported on macOS (CoreAudio). Windows (WASAPI) and Linux (ALSA)
+//! implementations are planned.
+
+#[cfg(target_os = "macos")]
+mod imp {
+    use clap::Parser;
+    use cpal::duplex::DuplexStreamConfig;
+    use cpal::traits::{DeviceTrait, HostTrait, StreamTrait};
+    use cpal::BufferSize;
+
+    #[derive(Parser, Debug)]
+    #[command(version, about = "CPAL duplex feedback example", long_about = None)]
+    struct Opt {
+        /// The audio device to use (must support duplex operation)
+        #[arg(short, long, value_name = "DEVICE")]
+        device: Option<String>,
+
+        /// Number of input channels
+        #[arg(long, value_name = "CHANNELS", default_value_t = 2)]
+        input_channels: u16,
+
+        /// Number of output channels
+        #[arg(long, value_name = "CHANNELS", default_value_t = 2)]
+        output_channels: u16,
+
+        /// Sample rate in Hz
+        #[arg(short, long, value_name = "RATE", default_value_t = 48000)]
+        sample_rate: u32,
+
+        /// Buffer size in frames
+        #[arg(short, long, value_name = "FRAMES", default_value_t = 512)]
+        buffer_size: u32,
+    }
+
+    pub fn run() -> anyhow::Result<()> {
+        let opt = Opt::parse();
+        let host = cpal::default_host();
+
+        // Find the device by device ID or use default
+        let device = if let Some(device_id_str) = opt.device {
+            let device_id = device_id_str.parse().expect("failed to parse device id");
+            host.device_by_id(&device_id)
+                .unwrap_or_else(|| panic!("failed to find device with id: {}", device_id_str))
+        } else {
+            host.default_output_device()
+                .expect("no default output device")
+        };
+
+        println!("Using device: \"{}\"", device.description()?.name());
+
+        // Create duplex stream configuration.
+        let config = DuplexStreamConfig {
+            input_channels: opt.input_channels,
+            output_channels: opt.output_channels,
+            sample_rate: opt.sample_rate,
+            buffer_size: BufferSize::Fixed(opt.buffer_size),
+        };
+
+        println!("Building duplex stream with config: {config:?}");
+
+        let stream = device.build_duplex_stream::<f32, _, _>(
+            &config,
+            move |input, output, _info| {
+                output.fill(0.0);
+                let copy_len = input.len().min(output.len());
+                output[..copy_len].copy_from_slice(&input[..copy_len]);
+            },
+            |err| eprintln!("Stream error: {err}"),
+            None,
+        )?;
+
+        println!("Successfully built duplex stream.");
+        println!(
+            "Input: {} channels, Output: {} channels, Sample rate: {} Hz, Buffer size: {} frames",
+            opt.input_channels, opt.output_channels, opt.sample_rate, opt.buffer_size
+        );
+
+        println!("Starting duplex stream...");
+        stream.play()?;
+
+        println!("Playing for 10 seconds... (speak into your microphone)");
+        std::thread::sleep(std::time::Duration::from_secs(10));
+
+        drop(stream);
+        println!("Done!");
+        Ok(())
+    }
+}
+
+fn main() {
+    #[cfg(target_os = "macos")]
+    imp::run().unwrap();
+
+    #[cfg(not(target_os = "macos"))]
+    {
+        eprintln!("Duplex streams are currently only supported on macOS.");
+        eprintln!("Windows (WASAPI) and Linux (ALSA) support is planned.");
+    }
+}
--- a/examples/duplex_feedback.rs
+++ b/examples/duplex_feedback.rs
@@ -11,7 +11,7 @@ mod imp {
     use clap::Parser;
     use cpal::duplex::DuplexStreamConfig;
     use cpal::traits::{DeviceTrait, HostTrait, StreamTrait};
-    use cpal::BufferSize;
+    use cpal::{BufferSize, ChannelCount, FrameCount, Sample, SampleRate};
 
     #[derive(Parser, Debug)]
     #[command(version, about = "CPAL duplex feedback example", long_about = None)]
@@ -22,19 +22,19 @@ mod imp {
 
         /// Number of input channels
         #[arg(long, value_name = "CHANNELS", default_value_t = 2)]
-        input_channels: u16,
+        input_channels: ChannelCount,
 
         /// Number of output channels
         #[arg(long, value_name = "CHANNELS", default_value_t = 2)]
-        output_channels: u16,
+        output_channels: ChannelCount,
 
         /// Sample rate in Hz
         #[arg(short, long, value_name = "RATE", default_value_t = 48000)]
-        sample_rate: u32,
+        sample_rate: SampleRate,
 
-        /// Buffer size in frames
-        #[arg(short, long, value_name = "FRAMES", default_value_t = 512)]
-        buffer_size: u32,
+        /// Buffer size in frames (omit for device default)
+        #[arg(short, long, value_name = "FRAMES")]
+        buffer_size: Option<FrameCount>,
     }
 
     pub fn run() -> anyhow::Result<()> {
@@ -42,13 +42,15 @@ mod imp {
         let host = cpal::default_host();
 
         // Find the device by device ID or use default
-        let device = if let Some(device_id_str) = opt.device {
-            let device_id = device_id_str.parse().expect("failed to parse device id");
-            host.device_by_id(&device_id)
-                .unwrap_or_else(|| panic!("failed to find device with id: {}", device_id_str))
-        } else {
-            host.default_output_device()
-                .expect("no default output device")
+        let device = match opt.device {
+            Some(device_id_str) => {
+                let device_id = device_id_str.parse().expect("failed to parse device id");
+                host.device_by_id(&device_id)
+                    .expect(&format!("failed to find device with id: {}", device_id_str))
+            }
+            None => host
+                .default_output_device()
+                .expect("no default output device"),
         };
 
         println!("Using device: \"{}\"", device.description()?.name());
@@ -58,7 +60,10 @@ mod imp {
             input_channels: opt.input_channels,
             output_channels: opt.output_channels,
             sample_rate: opt.sample_rate,
-            buffer_size: BufferSize::Fixed(opt.buffer_size),
+            buffer_size: opt
+                .buffer_size
+                .map(|s| BufferSize::Fixed(s))
+                .unwrap_or(BufferSize::Default),
         };
 
         println!("Building duplex stream with config: {config:?}");
@@ -66,7 +71,7 @@ mod imp {
         let stream = device.build_duplex_stream::<f32, _, _>(
             &config,
             move |input, output, _info| {
-                output.fill(0.0);
+                output.fill(Sample::EQUILIBRIUM);
                 let copy_len = input.len().min(output.len());
                 output[..copy_len].copy_from_slice(&input[..copy_len]);
             },
@@ -76,7 +81,7 @@ mod imp {
 
         println!("Successfully built duplex stream.");
         println!(
-            "Input: {} channels, Output: {} channels, Sample rate: {} Hz, Buffer size: {} frames",
+            "Input: {} channels, Output: {} channels, Sample rate: {} Hz, Buffer size: {:?} frames",
             opt.input_channels, opt.output_channels, opt.sample_rate, opt.buffer_size
         );
 
@@ -86,7 +91,6 @@ mod imp {
         println!("Playing for 10 seconds... (speak into your microphone)");
         std::thread::sleep(std::time::Duration::from_secs(10));
 
-        drop(stream);
         println!("Done!");
         Ok(())
     }
@@ -98,7 +102,6 @@ fn main() {
 
     #[cfg(not(target_os = "macos"))]
     {
-        eprintln!("Duplex streams are currently only supported on macOS.");
-        eprintln!("Windows (WASAPI) and Linux (ALSA) support is planned.");
+        eprintln!("Duplex streams are not supported on this platform.");
     }
 }
--- a/examples/duplex_feedback.rs
+++ b/examples/duplex_feedback.rs
@@ -1,10 +1,4 @@
-//! Feeds back the input stream directly into the output stream using a duplex stream.
-//!
-//! Unlike the `feedback.rs` example which uses separate input/output streams with a ring buffer,
-//! duplex streams provide hardware-synchronized input/output without additional buffering.
-//!
-//! Note: Currently only supported on macOS (CoreAudio). Windows (WASAPI) and Linux (ALSA)
-//! implementations are planned.
+// Duplex feedback example: feeds the input stream directly into the output.
 
 #[cfg(target_os = "macos")]
 mod imp {
@@ -41,7 +35,6 @@ mod imp {
         let opt = Opt::parse();
         let host = cpal::default_host();
 
-        // Find the device by device ID or use default
         let device = match opt.device {
             Some(device_id_str) => {
                 let device_id = device_id_str.parse().expect("failed to parse device id");
@@ -55,7 +48,6 @@ mod imp {
 
         println!("Using device: \"{}\"", device.description()?.name());
 
-        // Create duplex stream configuration.
         let config = DuplexStreamConfig {
             input_channels: opt.input_channels,
             output_channels: opt.output_channels,
--- a/src/duplex.rs
+++ b/src/duplex.rs
@@ -1,0 +1,79 @@
+//! Duplex audio stream support with synchronized input/output.
+//!
+//! This module provides types for building duplex (simultaneous input/output) audio streams
+//! with hardware clock synchronization.
+//!
+//! # Overview
+//!
+//! Unlike separate input and output streams which may have independent clocks, a duplex stream
+//! uses a single device context for both input and output, ensuring they share the same
+//! hardware clock. This is essential for applications like:
+//!
+//! - DAWs (Digital Audio Workstations)
+//! - Real-time audio effects processing
+//! - Audio measurement and analysis
+//! - Any application requiring sample-accurate I/O synchronization
+//!
+//! See `examples/duplex_feedback.rs` for a working example.
+
+use crate::{InputStreamTimestamp, OutputStreamTimestamp, SampleRate};
+
+/// Information passed to duplex callbacks.
+///
+/// This contains timing information for the current audio buffer, combining
+/// both input and output timing. A duplex stream has a single callback invocation
+/// that provides synchronized input and output data.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+pub struct DuplexCallbackInfo {
+    input_timestamp: InputStreamTimestamp,
+    output_timestamp: OutputStreamTimestamp,
+}
+
+impl DuplexCallbackInfo {
+    /// Create a new DuplexCallbackInfo.
+    ///
+    /// Note: Both timestamps will share the same `callback` instant since there is
+    /// only one callback invocation for a duplex stream.
+    pub fn new(
+        input_timestamp: InputStreamTimestamp,
+        output_timestamp: OutputStreamTimestamp,
+    ) -> Self {
+        Self {
+            input_timestamp,
+            output_timestamp,
+        }
+    }
+
+    /// The timestamp for the input portion of the duplex stream.
+    ///
+    /// Contains the callback instant and when the input data was captured.
+    pub fn input_timestamp(&self) -> InputStreamTimestamp {
+        self.input_timestamp
+    }
+
+    /// The timestamp for the output portion of the duplex stream.
+    ///
+    /// Contains the callback instant and when the output data will be played.
+    pub fn output_timestamp(&self) -> OutputStreamTimestamp {
+        self.output_timestamp
+    }
+}
+
+/// Configuration for a duplex audio stream.
+///
+/// Unlike separate input/output streams, duplex streams require matching
+/// configuration for both directions since they share a single device context.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct DuplexStreamConfig {
+    /// Number of input channels.
+    pub input_channels: u16,
+
+    /// Number of output channels.
+    pub output_channels: u16,
+
+    /// Sample rate in Hz.
+    pub sample_rate: SampleRate,
+
+    /// Requested buffer size in frames.
+    pub buffer_size: crate::BufferSize,
+}
--- a/src/duplex.rs
+++ b/src/duplex.rs
@@ -3,20 +3,9 @@
 //! This module provides types for building duplex (simultaneous input/output) audio streams
 //! with hardware clock synchronization.
 //!
-//! # Overview
-//!
-//! Unlike separate input and output streams which may have independent clocks, a duplex stream
-//! uses a single device context for both input and output, ensuring they share the same
-//! hardware clock. This is essential for applications like:
-//!
-//! - DAWs (Digital Audio Workstations)
-//! - Real-time audio effects processing
-//! - Audio measurement and analysis
-//! - Any application requiring sample-accurate I/O synchronization
-//!
 //! See `examples/duplex_feedback.rs` for a working example.
 
-use crate::{InputStreamTimestamp, OutputStreamTimestamp, SampleRate};
+use crate::{ChannelCount, InputStreamTimestamp, OutputStreamTimestamp, SampleRate};
 
 /// Information passed to duplex callbacks.
 ///
@@ -66,10 +55,10 @@ impl DuplexCallbackInfo {
 #[derive(Clone, Debug, Eq, PartialEq)]
 pub struct DuplexStreamConfig {
     /// Number of input channels.
-    pub input_channels: u16,
+    pub input_channels: ChannelCount,
 
     /// Number of output channels.
-    pub output_channels: u16,
+    pub output_channels: ChannelCount,
 
     /// Sample rate in Hz.
     pub sample_rate: SampleRate,
--- a/src/duplex.rs
+++ b/src/duplex.rs
@@ -1,17 +1,6 @@
-//! Duplex audio stream support with synchronized input/output.
-//!
-//! This module provides types for building duplex (simultaneous input/output) audio streams
-//! with hardware clock synchronization.
-//!
-//! See `examples/duplex_feedback.rs` for a working example.
-
 use crate::{ChannelCount, InputStreamTimestamp, OutputStreamTimestamp, SampleRate};
 
-/// Information passed to duplex callbacks.
-///
-/// This contains timing information for the current audio buffer, combining
-/// both input and output timing. A duplex stream has a single callback invocation
-/// that provides synchronized input and output data.
+// Timing information for a duplex callback, combining input and output timestamps.
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
 pub struct DuplexCallbackInfo {
     input_timestamp: InputStreamTimestamp,
@@ -19,10 +8,6 @@ pub struct DuplexCallbackInfo {
 }
 
 impl DuplexCallbackInfo {
-    /// Create a new DuplexCallbackInfo.
-    ///
-    /// Note: Both timestamps will share the same `callback` instant since there is
-    /// only one callback invocation for a duplex stream.
     pub fn new(
         input_timestamp: InputStreamTimestamp,
         output_timestamp: OutputStreamTimestamp,
@@ -33,36 +18,19 @@ impl DuplexCallbackInfo {
         }
     }
 
-    /// The timestamp for the input portion of the duplex stream.
-    ///
-    /// Contains the callback instant and when the input data was captured.
     pub fn input_timestamp(&self) -> InputStreamTimestamp {
         self.input_timestamp
     }
 
-    /// The timestamp for the output portion of the duplex stream.
-    ///
-    /// Contains the callback instant and when the output data will be played.
     pub fn output_timestamp(&self) -> OutputStreamTimestamp {
         self.output_timestamp
     }
 }
 
-/// Configuration for a duplex audio stream.
-///
-/// Unlike separate input/output streams, duplex streams require matching
-/// configuration for both directions since they share a single device context.
 #[derive(Clone, Debug, Eq, PartialEq)]
 pub struct DuplexStreamConfig {
-    /// Number of input channels.
     pub input_channels: ChannelCount,
-
-    /// Number of output channels.
     pub output_channels: ChannelCount,
-
-    /// Sample rate in Hz.
     pub sample_rate: SampleRate,
-
-    /// Requested buffer size in frames.
     pub buffer_size: crate::BufferSize,
 }
--- a/src/duplex.rs
+++ b/src/duplex.rs
@@ -27,7 +27,7 @@ impl DuplexCallbackInfo {
     }
 }
 
-#[derive(Clone, Debug, Eq, PartialEq)]
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
 pub struct DuplexStreamConfig {
     pub input_channels: ChannelCount,
     pub output_channels: ChannelCount,
--- a/src/host/coreaudio/macos/device.rs
+++ b/src/host/coreaudio/macos/device.rs
@@ -1,9 +1,6 @@
 use super::OSStatus;
 use super::Stream;
-use super::{
-    asbd_from_config, check_os_status, frames_to_duration, host_time_to_stream_instant,
-    DuplexCallbackPtr,
-};
+use super::{asbd_from_config, check_os_status, frames_to_duration, host_time_to_stream_instant};
 use crate::duplex::DuplexCallbackInfo;
 use crate::host::coreaudio::macos::loopback::LoopbackDevice;
 use crate::host::coreaudio::macos::StreamInner;
@@ -19,9 +16,7 @@ use coreaudio::audio_unit::render_callback::{self, data};
 use coreaudio::audio_unit::{AudioUnit, Element, Scope};
 use objc2_audio_toolbox::{
     kAudioOutputUnitProperty_CurrentDevice, kAudioOutputUnitProperty_EnableIO,
-    kAudioUnitErr_TooManyFramesToProcess, kAudioUnitProperty_SetRenderCallback,
-    kAudioUnitProperty_StreamFormat, AURenderCallbackStruct, AudioUnitRender,
-    AudioUnitRenderActionFlags,
+    kAudioUnitProperty_StreamFormat,
 };
 use objc2_core_audio::kAudioDevicePropertyDeviceUID;
 use objc2_core_audio::kAudioObjectPropertyElementMain;
@@ -37,11 +32,11 @@ use objc2_core_audio::{
     AudioObjectPropertyScope, AudioObjectSetPropertyData,
 };
 use objc2_core_audio_types::{
-    kAudio_ParamError, AudioBuffer, AudioBufferList, AudioStreamBasicDescription, AudioTimeStamp,
-    AudioValueRange,
+    AudioBuffer, AudioBufferList, AudioStreamBasicDescription, AudioValueRange,
 };
 use objc2_core_foundation::CFString;
 use objc2_core_foundation::Type;
+use std::mem::ManuallyDrop;
 
 pub use super::enumerate::{
     default_input_device, default_output_device, SupportedInputConfigs, SupportedOutputConfigs,
@@ -57,14 +52,13 @@ use super::invoke_error_callback;
 use super::property_listener::AudioObjectPropertyListener;
 use coreaudio::audio_unit::macos_helpers::get_device_name;
 
-/// Value for `kAudioOutputUnitProperty_EnableIO` to enable I/O on an AudioUnit element.
-const AUDIO_UNIT_IO_ENABLED: u32 = 1;
+pub(super) const AUDIO_UNIT_IO_ENABLED: u32 = 1;
 /// Value for `kAudioOutputUnitProperty_EnableIO` to disable I/O on an AudioUnit element.
 const AUDIO_UNIT_IO_DISABLED: u32 = 0;
 
 /// Attempt to set the device sample rate to the provided rate.
 /// Return an error if the requested sample rate is not supported by the device.
-fn set_sample_rate(
+pub(super) fn set_sample_rate(
     audio_device_id: AudioObjectID,
     target_sample_rate: SampleRate,
 ) -> Result<(), BuildStreamError> {
@@ -271,7 +265,7 @@ fn get_io_buffer_frame_size_range(
     })
 }
 
-fn estimate_capture_instant<E>(
+pub(super) fn estimate_capture_instant<E>(
     callback_instant: StreamInstant,
     delay: Duration,
     error_callback: &Mutex<E>,
@@ -292,7 +286,7 @@ where
     })
 }
 
-fn estimate_playback_instant<E>(
+pub(super) fn estimate_playback_instant<E>(
     callback_instant: StreamInstant,
     delay: Duration,
     error_callback: &Mutex<E>,
@@ -402,7 +396,14 @@ impl DeviceTrait for Device {
         D: FnMut(&Data, &mut Data, &DuplexCallbackInfo) + Send + 'static,
         E: FnMut(StreamError) + Send + 'static,
     {
-        Device::build_duplex_stream_raw(self, config, sample_format, data_callback, error_callback)
+        Device::build_duplex_stream_raw(
+            self,
+            config,
+            sample_format,
+            data_callback,
+            error_callback,
+            _timeout,
+        )
     }
 }
 
@@ -885,12 +886,13 @@ impl Device {
         let stream = Stream::new(
             StreamInner {
                 playing: true,
-                audio_unit,
+                audio_unit: ManuallyDrop::new(audio_unit),
                 device_id: self.audio_device_id,
                 _loopback_device: loopback_aggregate,
                 duplex_callback_ptr: None,
             },
             error_callback_for_stream,
+            false,
         )?;
 
         stream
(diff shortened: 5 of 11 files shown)

### CODE REVIEW

File: src/duplex.rs [hunk 1]
```diff
@@ -0,0 +1,457 @@
...
+//!         output[..input.len()].copy_from_slice(input);
+//!     },
+//!     |err| eprintln!("Stream error: {}", err),
+//!     None,
+//! ).expect("failed to build duplex stream");
+//! ```
+
+use crate::{PauseStreamError, PlayStreamError, SampleRate, StreamInstant};
+
+/// Hardware timestamp information from the audio device.
```
damyanp: Is there something inherent about supporting duplex streams that makes AudioTimestamp required?  Or is this something that would also be useful for normal streams?

gulbrand: No I don't think the duplex stream case has a unique or special need for AudioTimestamp (one caveat below). I think it would be helpful for input and output streams as well. But, adding support to input and output streams might be a breaking change? (I'm unsure).

I added it to support sample accurate operations in the application. For example, a DAW with automation events on nodes in a graph needs to know with high degrees of accuracy the sample position as a reference point for position and offset calculations for ramp ups, ramp downs, parameter automation, etc.

I believe it is technically possible to implement those without the hardware clock derived sample offset, but I think the edge case count increases when not doing so.

The only caveat I can think of for duplex is during simultaneous playback and recording in a DAW where its far more convenient to the user for the software to calculate latencies and offsets and adjust the final position of the recorded audio on the timeline. The math is precise (nearly perfectly so) with hardware clock sample offset rather than application-based timing calculations. Not critical though, just far simpler and more precise.

If removing this would be more consistent with the existing API would help, I can do that and then perhaps this concept could be introduced across all APIs in a future PR.

What do you think?

damyanp: My opinion here shouldn't have a huge amount of weight since I'm pretty new to this community.  But I'll give it anyway :)

For something like this that seems to have had several attempts over the years, I think it might be worth figuring out what the minimum surface area would be for the feature and try and get that supported for all the target platforms. 

I haven't looked closely enough to figure out if AudioTimestamp is something that can only be obtained from host-specific data? If it's something that could be entirely derived from information that's already in OutputCallbackInfo then it seems like a no-brainer to remove it for now.

If it is an abstraction that can be provided by all hosts, and would be useful for simplex as well as duplex, then IMO it'd be good to add it to both in one go. I'm not sure what the bar is for introducing breaking changes.

Similar to the concern with breaking changes, I'd suggest having proofs of concept, if not full implementations, for all platforms so we can be confident that the proposed API can work everywhere and won't need to be revised as soon as, say, Windows support is added.

gulbrand: Thanks for the feedback. This is helpful.

If I'm reading the feedback correctly, then if I remove the AudioTimestamp change from this, it greatly simplifies the scope since we don't need to worry about testing all the different platforms and host implementations for compatibility with the AudioTimestamp interface. Doing all of that work would be time consuming and is arguably orthogonal to adding duplex support to one platform / host.

Am I understanding the feedback correctly? (If so, I agree and I will remove AudioTimestamp)

damyanp: Yes, that's an accurate summary of my feedback.

src/host/coreaudio/macos/mod.rs
```diff
@@ -263,6 +263,203 @@ impl StreamTrait for Stream {
...
+// 2. AudioDeviceID is Copy
+// 3. duplex_callback_ptr points to a Send type (DuplexProcWrapper)
+//    and is only accessed during Drop after stopping the audio unit
+unsafe impl Send for DuplexStreamInner {}
+
+impl DuplexStreamInner {
+    fn play(&mut self) -> Result<(), PlayStreamError> {
+        if !self.playing {
+            if let Err(e) = self.audio_unit.start() {
+                let description = format!("{e}");
```
Decodetalkers: I think this line can just be e.to_string()

src/host/coreaudio/macos/mod.rs
```diff
@@ -263,6 +263,203 @@ impl StreamTrait for Stream {
...
+        device_id: AudioDeviceID,
+        stream_weak: Weak<Mutex<DuplexStreamInner>>,
+        error_callback: Arc<Mutex<ErrorCallback>>,
+    ) -> Result<Self, crate::BuildStreamError> {
+        let (shutdown_tx, shutdown_rx) = mpsc::channel();
+        let (disconnect_tx, disconnect_rx) = mpsc::channel();
+        let (ready_tx, ready_rx) = mpsc::channel();
+
+        let disconnect_tx_clone = disconnect_tx.clone();
+        std::thread::spawn(move || {
```
Decodetalkers: maybe we also need to collect the thread handler?

src/host/coreaudio/macos/mod.rs
```diff
@@ -263,6 +263,203 @@ impl StreamTrait for Stream {
...
+
+/// Duplex stream disconnect manager - handles device disconnection.
+struct DuplexDisconnectManager {
+    _shutdown_tx: mpsc::Sender<()>,
+}
+
+impl DuplexDisconnectManager {
+    fn new(
+        device_id: AudioDeviceID,
+        stream_weak: Weak<Mutex<DuplexStreamInner>>,
```
Decodetalkers: what is the meaning of `Weak<Mutex>`? I cannot understand

src/duplex.rs
```diff
@@ -0,0 +1,457 @@
...
+    }
+}
+
+/// A placeholder duplex stream type for backends that don't yet support duplex.
+///
+/// This type implements `StreamTrait` but all operations return errors.
+/// Backend implementations should replace this with their own type once
+/// duplex support is implemented.
+pub struct UnsupportedDuplexStream {
+    _private: (),
```
Decodetalkers: I do not know why we need a _private here.. And maybe here just use #[derive(Default)] will be ok

src/host/coreaudio/macos/mod.rs
gulbrand: Yep, I changed the rest of them as well. Thanks!

src/duplex.rs
gulbrand: Great, thanks, I'll remove AudioTimestamp

src/duplex.rs
gulbrand: I think your suggestion is correct. I am not sure but I think this was added in anticipation of this maybe, possibly, needing an additional field for one of the backends. But now that I've implemented the necessary traits for each backend, it is clear this is not needed. I'll remove this and make this a unit struct.

src/host/coreaudio/macos/mod.rs
gulbrand: Ah, good catch. I need to think about this one a bit more, but I think your suggestion is correct because the thread should be joined during drop and without the handle, drop can't join.

I need to make sure I understand how this works. I'll update this comment when I'm more confident on this.

Thanks for pointing this out.

roderickvd:
@gulbrand thanks for your contribution. I don't mind the AI usage as long as the output is of high quality and what I'd expect from a seasoned developer.

Let me know when you've addressed @Decodetalkers review points, and are ready for my detailed review.

Very quickly, I like having `build_duplex_stream` analogous to the input and output stream build functions, but wonder if we need a separate `DuplexStream` or whether we could unify it with the existing `Stream`. I admit I've given the latter very little thought, so tell me if I'm missing something entirely obvious.

src/host/coreaudio/macos/mod.rs
gulbrand: Thank you for your feedback!

I dug deeper into this and I have a weakly held opinion that this is fine as-is.

I reviewed the threads and their terminating conditions and to me it looks like the shutdown_tx value does a good enough job of making sure the thread terminates after the StreamInner is dropped.

This thread's sole job is to make sure the property_address value is dropped in the same thread that it is created in ... The CoreAudio docs are not clear to me that this is indeed 100% required, but it is safer to keep the allocation and the drop on the same thread just in case.

That said, I'm working on getting rid of DuplexStream entirely so that the duplex code paths can work with Stream and StreamInner directly. If that works, then this extra code for DuplexDisconnectManager will be deleted and the implementations will be unified to the DisconnectManager.

gulbrand: In terms of the join, I think a case could be made to change this code to allow for a join, but I'm not convinced it is necessary for this to be correct. The lifetimes do appear to be correct and the only benefit of joining would be to catch an error.

I agree there is a case to be made for that, but I'd like to try to keep that change and discussion separate from this PR. Is that a reasonable approach?

src/host/coreaudio/macos/mod.rs
gulbrand: I believe there the purposes here are:

1. Allow the StreamInner to be dropped regardless of the DisconnectManager's state. Using an Arc means the DisconnectManager will block dropping until it is ready to drop. 
2. Additionally this is also the key indication to the thread that is handling disconnect events to exit its while loop. 
 
See the else block [here](https://github.com/RustAudio/cpal/blob/master/src/host/coreaudio/macos/mod.rs#L142):

If the upgrade from Weak to Arc fails, it means the stream has been dropped and this thread should exit immediately.

```rust
        std::thread::spawn(move || {
            while disconnect_rx.recv().is_ok() {
                // Check if stream still exists
                if let Some(stream_arc) = stream_weak_clone.upgrade() {
                    // First, try to pause the stream to stop playback
                    if let Ok(mut stream_inner) = stream_arc.try_lock() {
                        let _ = stream_inner.pause();
                    }

                    // Always try to notify about device disconnection
                    invoke_error_callback(
                        &error_callback_clone,
                        crate::StreamError::DeviceNotAvailable,
                    );
                } else {
                    // Stream is gone, exit the handler thread
                    break;
                }
            }
        });

```
All of that said, and like I mentioned in the other comment, I am working on removing DuplexStream and DuplexStreamInner which would unify this change onto the existing Stream types and also reuse the existing DisconnectManager. At that point, this area would seem to me to be out of scope for this PR.

Any thoughts on this I should consider?

Thank you!

src/duplex.rs
gulbrand: PR is updated to remove AudioTimestamp. Any concerns with resolving this conversation?

src/host/coreaudio/macos/mod.rs
gulbrand: PR is updated to change these to to_string. Thank you! Any concerns if I resolve this?

src/host/coreaudio/macos/mod.rs
gulbrand: I've updated the PR to reuse the existing DisconnectManager. You're concern is still valid, but I think it should be addressed outside of this PR as it is an existing issue. I think this is ok though. Any concerns if I resolve this?

src/duplex.rs
gulbrand: I've updated to PR to not even have this type anymore and to just use the StreamConfigNotSupported error instead. 

Any concerns if I resolve this? Thank you!

gulbrand:
> @gulbrand thanks for your contribution. I don't mind the AI usage as long as the output is of high quality and what I'd expect from a seasoned developer.
> 
> Let me know when you've addressed @Decodetalkers review points, and are ready for my detailed review.
> 
> Very quickly, I like having `build_duplex_stream` analogous to the input and output stream build functions, but wonder if we need a separate `DuplexStream` or whether we could unify it with the existing `Stream`. I admit I've given the latter very little thought, so tell me if I'm missing something entirely obvious.

Agreed. I've updated the PR but need a bit of time to test in separate projects again and to double check the safety claims. I'll ping again once I'm done with testing but I think this PR is in a much better state now.

src/duplex.rs
Decodetalkers: I see

src/duplex.rs
damyanp: Thanks! Happy for this to be resolved. (It's odd I don't seem to be able to do that myself!)

src/host/coreaudio/macos/mod.rs
gulbrand: This has been removed from this PR as I was able to unify DuplexInnerStream and InnerStream, making the DisconnectManager reusable.

src/host/coreaudio/macos/mod.rs
gulbrand: I've removed the custom DisconnectManager so this comment no longer applies to this PR, however, this is still potentially relevant for the existing DisconnectManager.

src/host/coreaudio/macos/device.rs
```diff
@@ -945,6 +969,449 @@ impl Device {
...
+                                    "AudioUnitRender failed for input: OSStatus {}",
+                                    status
+                                ),
+                            },
+                        },
+                    );
+                    input_buffer[..input_bytes].fill(0);
+                }
+
+                let input_data = unsafe {
```
gulbrand: I missed a safety comment, I will add this.

src/host/coreaudio/macos/device.rs
```diff
@@ -945,6 +969,464 @@ impl Device {
...
+            audio_unit,
+            device_id: self.audio_device_id,
+            _loopback_device: None,
+            duplex_callback_ptr: Some(DuplexCallbackPtr(wrapper_ptr)),
+        };
+
+        // Create error callback for stream - either dummy or real based on device type.
+        // For duplex, only swallow disconnect if the device is the default for both
+        // roles — otherwise Core Audio won't re-route both directions.
+        let error_callback_for_stream: super::ErrorCallback =
```
gulbrand: I've changed my mind on this. I'm not sure though. I think if a disconnect happens at all in duplex mode, the stream needs to be torn down and recreated.
@roderickvd what do you think?

examples/duplex_feedback.rs
```diff
@@ -0,0 +1,104 @@
...
+    #[derive(Parser, Debug)]
+    #[command(version, about = "CPAL duplex feedback example", long_about = None)]
+    struct Opt {
+        /// The audio device to use (must support duplex operation)
+        #[arg(short, long, value_name = "DEVICE")]
+        device: Option<String>,
+
+        /// Number of input channels
+        #[arg(long, value_name = "CHANNELS", default_value_t = 2)]
+        input_channels: u16,
```
roderickvd: Please use type aliases like `ChannelCount` and `SampleRate`.

examples/duplex_feedback.rs
```diff
@@ -0,0 +1,104 @@
...
+        #[arg(long, value_name = "CHANNELS", default_value_t = 2)]
+        output_channels: u16,
+
+        /// Sample rate in Hz
+        #[arg(short, long, value_name = "RATE", default_value_t = 48000)]
+        sample_rate: u32,
+
+        /// Buffer size in frames
+        #[arg(short, long, value_name = "FRAMES", default_value_t = 512)]
+        buffer_size: u32,
```
roderickvd: Idea: make it `Option<usize>` and then have it use either `BufferSize::Default` or `Fixed`.

examples/duplex_feedback.rs
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    pub fn run() -> anyhow::Result<()> {
+        let opt = Opt::parse();
+        let host = cpal::default_host();
+
+        // Find the device by device ID or use default
+        let device = if let Some(device_id_str) = opt.device {
```
roderickvd: `match` seems more idiomatic here.

examples/duplex_feedback.rs
```diff
@@ -0,0 +1,104 @@
...
+            let device_id = device_id_str.parse().expect("failed to parse device id");
+            host.device_by_id(&device_id)
+                .unwrap_or_else(|| panic!("failed to find device with id: {}", device_id_str))
```
roderickvd: Why not use `expect` here as well?

examples/duplex_feedback.rs
```diff
@@ -0,0 +1,104 @@
...
+            sample_rate: opt.sample_rate,
+            buffer_size: BufferSize::Fixed(opt.buffer_size),
+        };
+
+        println!("Building duplex stream with config: {config:?}");
+
+        let stream = device.build_duplex_stream::<f32, _, _>(
+            &config,
+            move |input, output, _info| {
+                output.fill(0.0);
```
roderickvd: Better to use `Sample::EQUILIBRIUM` instead of `0.0`.

examples/duplex_feedback.rs
```diff
@@ -0,0 +1,104 @@
...
+            opt.input_channels, opt.output_channels, opt.sample_rate, opt.buffer_size
+        );
+
+        println!("Starting duplex stream...");
+        stream.play()?;
+
+        println!("Playing for 10 seconds... (speak into your microphone)");
+        std::thread::sleep(std::time::Duration::from_secs(10));
+
+        drop(stream);
```
roderickvd: Should not be necessary?

examples/duplex_feedback.rs
```diff
@@ -0,0 +1,104 @@
...
+}
+
+fn main() {
+    #[cfg(target_os = "macos")]
+    imp::run().unwrap();
+
+    #[cfg(not(target_os = "macos"))]
+    {
+        eprintln!("Duplex streams are currently only supported on macOS.");
+        eprintln!("Windows (WASAPI) and Linux (ALSA) support is planned.");
```
roderickvd: Planned by who? 😄 Maybe just remove that second line for now.

src/host/coreaudio/macos/device.rs
```diff
@@ -945,6 +969,464 @@ impl Device {
 
         Ok(stream)
     }
+
+    /// Calculate latency-adjusted capture and playback timestamps with graceful error handling.
+    ///
+    /// Note: input/output streams use .expect() here and will panic - while it is
```
roderickvd: 1. The code comments explaining your thoughts are appreciated, but keep them `//` even though they aren't exposed publicly. I also think it could be shorter.

2. Let's go for consistent behavior. If you handle errors gracefully here, then let's make it so in simplex streams also.

src/host/coreaudio/macos/device.rs
```diff
@@ -945,6 +969,464 @@ impl Device {
...
+    /// While theoretically and probably practically impossible, this is a critical
+    /// user-experience issue -- the app shouldn't crash and potentially lose end-user data.
+    /// The only reason `sub` would overflow is if delay is massively too large, which would
+    /// indicate some kind of major bug or failure in the OS since callback_instant is derived
+    /// from host time. Still, I think the best practice is NOT to panic but tell the app
+    /// and fallback to a degraded latency estimate of no latency.
+    fn calculate_duplex_timestamps<E>(
+        callback_instant: crate::StreamInstant,
+        delay: std::time::Duration,
+        error_callback: &Arc<Mutex<E>>,
```
roderickvd: Passing `&Arc<Mutex<E>>` by reference seems weird. `Arc` is already a shared pointer. Shouldn't this simplify to `&Mutex<E>`?

src/host/coreaudio/macos/device.rs
```diff
@@ -945,6 +969,464 @@ impl Device {
...
+    ) -> (crate::StreamInstant, crate::StreamInstant)
```
roderickvd: Maybe import `StreamInstant` into scope instead of repeating the path.

src/host/coreaudio/macos/device.rs
```diff
@@ -945,6 +969,464 @@ impl Device {
...
+        config: &crate::duplex::DuplexStreamConfig,
+        sample_format: SampleFormat,
+        data_callback: D,
+        error_callback: E,
+    ) -> Result<Stream, BuildStreamError>
+    where
+        D: FnMut(&Data, &mut Data, &DuplexCallbackInfo) + Send + 'static,
+        E: FnMut(StreamError) + Send + 'static,
+    {
+        // Validate that device supports duplex
```
roderickvd: This code comment is rather superfluous - please check other code comments too if they can be removed because the code is self-explanatory.

src/host/coreaudio/macos/device.rs
```diff
@@ -945,6 +969,464 @@ impl Device {
...
+        if !self.supports_input() || !self.supports_output() {
```
roderickvd: Can be simplified to `self.supports_duplex`.

src/host/coreaudio/macos/device.rs
```diff
@@ -945,6 +969,464 @@ impl Device {
...
+        }
+
+        // Potentially change the device sample rate to match the config.
+        set_sample_rate(self.audio_device_id, config.sample_rate)?;
+
+        // Create HAL AudioUnit - always use HalOutput for duplex
+        let mut audio_unit = AudioUnit::new(coreaudio::audio_unit::IOType::HalOutput)?;
+
+        // Enable BOTH input and output on the AudioUnit
+        let enable: u32 = 1;
```
roderickvd: If this is a magic number then should it be defined as a constant for clarity?

src/host/coreaudio/macos/device.rs
```diff
@@ -945,6 +969,464 @@ impl Device {
...
+        }
+
+        // Query the buffer size range to determine the maximum possible buffer size.
+        //
+        // BUFFER ALLOCATION STRATEGY:
+        // Duplex streams must pre-allocate an input buffer because AudioUnitRender() requires
+        // us to provide the buffer. Unlike input/output streams where CoreAudio provides the
+        // buffer, we cannot dynamically reallocate in the callback without:
+        //   1. Violating real-time constraints (allocation is not RT-safe)
+        //   2. Implementing complex double/triple buffering (not worth it for rare edge case)
```
roderickvd: Not sure what you're hinting at, but double-buffering doesn't need to be complex. CoreAudio probably also does it. Do you mean double-buffering _on top of_ CoreAudio?

src/host/coreaudio/macos/device.rs
```diff
@@ -945,6 +969,464 @@ impl Device {
...
+        //
+        // Therefore, we allocate for the MAXIMUM buffer size reported by the device. This
```
roderickvd: From ALSA / PipeWire experience I know that this can be quite costly. In practice it means that the `cpal` app will hog RAM. What are ways to circumvent this?

src/host/coreaudio/macos/device.rs
```diff
@@ -945,6 +969,464 @@ impl Device {
...
+        //
+        // LIMITATION:
+        // If CoreAudio requests more frames than the device's reported maximum buffer size,
+        // the stream will fail with kAudioUnitErr_TooManyFramesToProcess. This can happen if:
+        //   - The device's buffer size is changed after stream creation (via System Settings
+        //     or another audio application)
+        //   - The device reconfigures itself (e.g., joins/leaves an aggregate device)
+        //   - The device's supported buffer size range changes due to sample rate or format
+        //     changes
+        // Recovery requires the user to destroy and recreate the stream, which will query
```
roderickvd: And `cpal` throws an error, right? Please have it throw a `DeviceNotAvailable` and in a future PR we will split that variant into another variant that makes it clear it needs to be torn down and reconfigured.

src/host/coreaudio/macos/device.rs
```diff
@@ -945,6 +969,464 @@ impl Device {
...
+        // the new buffer size range and allocate appropriately.
+        let buffer_size_range = get_io_buffer_frame_size_range(&audio_unit)?;
+        let max_buffer_size = match buffer_size_range {
+            crate::SupportedBufferSize::Range { max, .. } => max,
+            crate::SupportedBufferSize::Unknown => {
+                // Fallback: query current size and add headroom for safety
+                let current: u32 = audio_unit
```
roderickvd: This seems redundant with `get_io_buffer_frame_size_range`?

src/host/coreaudio/macos/device.rs
```diff
@@ -945,6 +969,464 @@ impl Device {
...
+        // Configuration for callback
+        let input_channels = config.input_channels as usize;
+        let sample_bytes = sample_format.sample_size();
+
+        // Pre-allocate input buffer for the maximum buffer size (in bytes).
+        // This ensures we can handle dynamic buffer size changes without reallocation.
+        // See the BUFFER ALLOCATION STRATEGY comment above for detailed rationale.
+        let input_buffer_samples = max_buffer_size as usize * input_channels;
+        let input_buffer_bytes = input_buffer_samples * sample_bytes;
+        let mut input_buffer: Vec<u8> = vec![0u8; input_buffer_bytes];
```
roderickvd: When the size is fixed, you can make it a `Box<u8>` for a tiny optimization and clearer intent.

src/host/coreaudio/macos/device.rs
```diff
@@ -945,6 +969,464 @@ impl Device {
...
+        let raw_audio_unit = *audio_unit.as_ref();
+
+        // Configuration for callback
+        let input_channels = config.input_channels as usize;
+        let sample_bytes = sample_format.sample_size();
+
+        // Pre-allocate input buffer for the maximum buffer size (in bytes).
+        // This ensures we can handle dynamic buffer size changes without reallocation.
+        // See the BUFFER ALLOCATION STRATEGY comment above for detailed rationale.
+        let input_buffer_samples = max_buffer_size as usize * input_channels;
```
roderickvd: I believe the buffer size, and so the callback size, is for both channels at once.

src/host/coreaudio/macos/device.rs
```diff
@@ -1017,3 +1499,76 @@ fn get_device_buffer_frame_size(audio_unit: &AudioUnit) -> Result<usize, coreaud
     )?;
     Ok(frames as usize)
 }
+
+// ============================================================================
+// Duplex callback infrastructure
```
roderickvd: This section could live in its own file.

src/host/coreaudio/macos/mod.rs
```diff
@@ -182,13 +196,22 @@ struct StreamInner {
     /// Manage the lifetime of the aggregate device used
     /// for loopback recording
     _loopback_device: Option<LoopbackDevice>,
+    /// Pointer to the duplex callback wrapper, manually managed for duplex streams.
+    ///
+    /// coreaudio-rs doesn't support duplex streams (enabling both input and output
```
roderickvd, line 253: How much of what you've done in this PR should live in `coreaudio-rs`, you think?

src/duplex.rs
```diff
@@ -0,0 +1,79 @@
+//! Duplex audio stream support with synchronized input/output.
+//!
+//! This module provides types for building duplex (simultaneous input/output) audio streams
+//! with hardware clock synchronization.
+//!
+//! # Overview
+//!
+//! Unlike separate input and output streams which may have independent clocks, a duplex stream
+//! uses a single device context for both input and output, ensuring they share the same
+//! hardware clock. This is essential for applications like:
```
roderickvd: No need to mention the use cases.

src/duplex.rs
```diff
@@ -0,0 +1,79 @@
...
+}
+
+/// Configuration for a duplex audio stream.
+///
+/// Unlike separate input/output streams, duplex streams require matching
+/// configuration for both directions since they share a single device context.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct DuplexStreamConfig {
+    /// Number of input channels.
+    pub input_channels: u16,
```
roderickvd: `ChannelCount`

src/traits.rs
```diff
@@ -286,6 +298,71 @@ pub trait DeviceTrait {
     where
         D: FnMut(&mut Data, &OutputCallbackInfo) + Send + 'static,
         E: FnMut(StreamError) + Send + 'static;
+
+    /// Create a duplex stream with synchronized input and output sharing the same hardware clock.
+    ///
+    /// Essential for DAWs, real-time effects, and any application requiring sample-accurate I/O sync.
```
roderickvd: Can be removed.

README.md
```diff
@@ -209,6 +210,7 @@ CPAL comes with several examples demonstrating various features:
 - `beep` - Generate a simple sine wave tone
 - `enumerate` - List all available audio devices and their capabilities
 - `feedback` - Pass input audio directly to output (microphone loopback)
+- `duplex_feedback` - Hardware-synchronized duplex stream loopback (macOS only)
```
roderickvd: Let's drop the "macOS only" part: we'll forget when we add duplex support for other hosts.

src/host/coreaudio/macos/device.rs
```diff
@@ -1017,3 +1499,76 @@ fn get_device_buffer_frame_size(audio_unit: &AudioUnit) -> Result<usize, coreaud
...
+/// move semantics, so no Mutex is needed.
+///
+/// Note: `extern "C-unwind"` is required here because `AURenderCallbackStruct`
+/// from coreaudio-sys types the `inputProc` field as `extern "C-unwind"`.
+/// Ideally this would be `extern "C"` so that a panic aborts rather than
+/// unwinding through CoreAudio's C frames (this callback runs on CoreAudio's
+/// audio thread with no Rust frames above to catch an unwind). Per RFC 2945
+/// (https://github.com/rust-lang/rust/issues/115285), `extern "C"` aborts on
+/// panic, which would be the correct behavior here.
+extern "C-unwind" fn duplex_input_proc(
```
roderickvd: A standard workaround is `catch_unwind`:

```rust
extern "C-unwind" fn duplex_input_proc(...) -> i32 {
    let wrapper = unsafe { in_ref_con.cast::<DuplexProcWrapper>().as_mut() };
    let result = std::panic::catch_unwind(std::panic::AssertUnwindSafe(|| {
        (wrapper.callback)(...)
    }));
    match result {
        Ok(ret) => ret,
        Err(_) => {
            // Log or invoke error callback if possible; the panic was caught
            kAudio_ParamError
        }
    }
}
```

src/host/coreaudio/macos/mod.rs
```diff
@@ -210,6 +233,31 @@ impl StreamInner {
...
 
+impl Drop for StreamInner {
+    fn drop(&mut self) {
+        // Clean up duplex callback if present.
+        if let Some(DuplexCallbackPtr(ptr)) = self.duplex_callback_ptr {
+            if !ptr.is_null() {
+                // Stop the audio unit to ensure the callback is no longer being called
+                // before reclaiming duplex_callback_ptr below. We must stop here regardless
+                // of AudioUnit::drop's behavior.
+                // Note: AudioUnit::drop will also call stop() — likely safe, but we stop here anyway.
```
roderickvd: Following your points over at Discord, you are right to be careful about assuming `audio_unit.stop()` being idempotent. `ManuallyDrop<AudioUnit>` would be cleaner and more explicit.

src/host/coreaudio/macos/device.rs
```diff
@@ -945,6 +969,464 @@ impl Device {
...
+            _loopback_device: None,
+            duplex_callback_ptr: Some(DuplexCallbackPtr(wrapper_ptr)),
+        };
+
+        // Create error callback for stream - either dummy or real based on device type.
+        // For duplex, only swallow disconnect if the device is the default for both
+        // roles — otherwise Core Audio won't re-route both directions.
+        let error_callback_for_stream: super::ErrorCallback =
+            if is_default_input_device(self) && is_default_output_device(self) {
+                Box::new(|_: StreamError| {})
```
roderickvd: I think we shouldn't silently swallow disconnects, and propagate the error instead. A duplex stream is broken when either direction changes device.

src/host/coreaudio/macos/device.rs
```diff
@@ -945,6 +969,464 @@ impl Device {
...
+    /// audio thread. Errors here should be extremely rare (timestamp arithmetic
+    /// overflow/underflow).
+    ///
+    /// While theoretically and probably practically impossible, this is a critical
+    /// user-experience issue -- the app shouldn't crash and potentially lose end-user data.
+    /// The only reason `sub` would overflow is if delay is massively too large, which would
+    /// indicate some kind of major bug or failure in the OS since callback_instant is derived
+    /// from host time. Still, I think the best practice is NOT to panic but tell the app
+    /// and fallback to a degraded latency estimate of no latency.
+    fn calculate_duplex_timestamps<E>(
```
roderickvd: This utility method could be a module-level function instead of on the struct.

roderickvd (changes requested):
Took a while but here's my first code review.

examples/duplex_feedback.rs
edwloef: `expect` can't take a format string. I've done this `.unwrap_or_else(|| panic!(...))` pattern myself a few times, It's the best way I know of to custom-format the panic message.

src/host/coreaudio/macos/device.rs
roderickvd: I agree.

roderickvd:
@gulbrand checking in - what's your planning on this one? No pressure, just asking.

gulbrand:
> @gulbrand checking in - what's your planning on this one? No pressure, just asking.

I'll be able to focus on this PR this weekend.

src/host/coreaudio/macos/mod.rs
gulbrand: Creating a record here of what I pointed out in Discord:

The issue is that StreamInner now holds a duplex_callback_ptr: Option<*mut ...> (b/c coreaudio-rs doesn't have a builder for duplex streams, we have to implement it here unless we also want to change coreaudio-rs).

```
struct StreamInner {
    playing: bool,
    audio_unit: AudioUnit,
    
    duplex_callback_ptr: Option<*mut device::DuplexProcWrapper>,
}
```

The problem is the StreamInner.audio_unit needs to be stopped before duplex_callback_ptr is dropped--something cpal hasn't had to worry about due to coreaudio-rs doing the work.

It looks like I'll need to implement Drop for StreamInner, but this needs to call audio_unit.stop first which leaves self.audio_unit for full drop later, and that also calls audio_unit.stop. I believe this works and is safe, but this feels awkard and I don't know this is safe to rely on audio_unit.stop being idempotent now and in the future.

or
```
struct StreamInner {
    playing: bool,
    audio_unit: ManuallyDrop<AudioUnit>,
}
```

This feels much better. Now in StreamInner::drop we can manually drop the audio_unit and then drop the duplex_callback_ptr.

I'm leaning toward ManuallyDrop, but I'm looking for feedback from folks that know better than I do (probably everyone here 🙂 )

gulbrand: Per the above, I will try the ManuallyDrop approach and test. I think that's the best path forward.

README.md
gulbrand: Will be fixed in next commit

examples/duplex_feedback.rs
gulbrand: I like expect better. They both seem idiomatic, but to me personally, expect feels a little more "gooder" here :) . This is just the example app so I'm not sure it matters too much.

examples/duplex_feedback.rs
gulbrand: You, of course! 🤣 Just kidding. Actually, me, I am planning on doing it this winter/spring, but yes you are correct we should remove this.

gulbrand: I changed the first line as well to "not supported on this platform" so that this is true while other platforms are added.

src/host/coreaudio/macos/device.rs
gulbrand: To be transparent, this is a bit above my understanding and I'm going off of what I'm learning about this aspect of Rust and C FFI.

IIUC, catch_unwind would prevent UB by returning an error to CoreAudio which would kill the audio stream and keep the process away from UB.

That makes sense to me so I'll make this change.

src/host/coreaudio/macos/device.rs
gulbrand: I added a two constants and updated the entire file to use them, which creeped slightly outside the scope of duplex, but I think that's fine.

src/host/coreaudio/macos/device.rs
gulbrand: We could do that, but there's multiple call sites that *should* also be updated that are not related to the duplex change.

I changed the signature in mod.rs and auto-deref means we don't *have* to update the non-duplex call sites.

Do you prefer I update all the call sites, even outside the scope of the duplex call site?

src/host/coreaudio/macos/device.rs
gulbrand: Yeah, hrm. I didn't catch this earlier. I think we're better off still handling all the variants but returning an error for Unknown to guard against get_io_buffer_frame_size_range changing.

I'll keep thinking about this, but for now, I'll change the code to do this. For now it would be dead code practically, but when someone or if someone changes get_io_buffer_frame_size_range it won't be a panic but an error instead.

Thoughts?

gulbrand: Also, I'm noodling on how to handle buffer pre-allocation and re-allocation so this might be moot anyway.

src/host/coreaudio/macos/device.rs
gulbrand: I broke the function into two, usable by input, output and the composition for duplex, so now all three Streams report an error instead of a panic.

gulbrand:
@roderickvd Thank you for your feedback!

FYI: I squashed my branch. I had 40+ commits and rebasing was a pain.

I finished a [first pass at addressing your feedback](https://github.com/RustAudio/cpal/pull/1096/changes/74331e6cb569bfb9ab61e35f81cae2b0ca54f86f)--mostly focussing on the easy fixes 🤣 .

I want to follow-up on:

1. refactor the buffer pre-alloc/re-alloc so we don't have to allocate the max buffer size;
2. refactor the code generally to try to put anything we might want in coreaudio-rs in its own file / set of files.

I think those two steps would answer the remaining feedback comments and answer the question about what might move to coreaudio-rs.

examples/duplex_feedback.rs
```diff
@@ -0,0 +1,107 @@
...
+            buffer_size: opt
+                .buffer_size
+                .map(|s| BufferSize::Fixed(s))
+                .unwrap_or(BufferSize::Default),
+        };
+
+        println!("Building duplex stream with config: {config:?}");
+
+        let stream = device.build_duplex_stream::<f32, _, _>(
+            &config,
```
Decodetalkers, line 64: Since it is copy able. So we do not need & I think

gulbrand:
I made some progress today answering the question of what would belong in coreaudio-rs vs cpal. I tried a couple of prototypes and there's one more I want to look at. I'll need another round on this one before I can form a helpful opinion.

On the allocation issue -- I will try to rework that so that the stream is invalidated with an error if coreaudio asks for a different buffer size during streaming. I looked at what a few pro audio apps do and they detect this and stop audio processing and some of them even ask the user what to do about it. This seems like the right approach here as well. I think that's going to be better than just blindly maxing out the allocation buffer.

@roderickvd thoughts on this approach?

The other change I want to make is I think this should be feature flagged behind os target macos. I haven't been able to test this anywhere else, like iOS etc. I won't be able to test those anytime soon either.

src/host/coreaudio/macos/device.rs
roderickvd: Next version will bump SemVer so that's not a constraint. I can't think of a reason why we'd need a `&Arc<Mutex<E>>` so I'm for some opportunistic refactoring.

src/host/coreaudio/macos/device.rs
roderickvd: Could you clarify what you are proposing, to be sure I'm understanding correctly?

roderickvd:
> On the allocation issue -- I will try to rework that so that the stream is invalidated with an error if coreaudio asks for a different buffer size during streaming. I looked at what a few pro audio apps do and they detect this and stop audio processing and some of them even ask the user what to do about it. This seems like the right approach here as well. I think that's going to be better than just blindly maxing out the allocation buffer.

Yes, I think that's better.

With regard to "asking what to do about it" - not sure if and how we could do that in cpal beyond firing the error callback with a specific error type. I'm planning on refactoring cpal errors into one `std::io::Error` kind of error type with a non-exhaustive list of error kinds, if that helps. I hope to put up that PR shortly.

src/host/coreaudio/macos/device.rs
gulbrand: I actually did go ahead with this change and I discovered that we can change the signature without needing to change any call sites, so I'll move forward with this suggestion.

roderickvd:
@gulbrand I linked #349 but not yet #628 as the latter is specific to ALSA.

gulbrand:
@roderickvd 

~~Should we put this behind a feature 'duplex-macos' for now? I haven't tested this on iOS or anything else and I have no plans to do so right away.~~ Nevermind, this is already behind macos

examples/duplex_feedback.rs
gulbrand: True, it probably should be Copy. but the current interface for input/output streams are not, even though they can be Copy today. I would imagine that this would always be Copy able, but I'm inclined to keep the interface the same for now.

TBH, I'm also trying to keep scope as limited as possible for this change. I'm happy to take on more work around these kinds of things, but this being not Copy isn't a correctness issue and changing it to Copy opens scope to changing the other Streams and their builder functions. I'd rather stick to a smaller scope to get this wrapped up.

gulbrand:
@roderickvd I plan on an update ready for your review today/tomorrow. I'm really close, I just cleaning up comments now. The files have been separated and most of the comments are addressed either with a comment here or a code change.

As for what could move over to coreaudio-rs, that remains unclear to me how to best do that. The first challenge I ran into was the error handling. Keeping this code in cpal for now means we can control what to do on errors better. Moving that to coreaudio-rs may mean updating its error model.

With some work, I think it would be possible, but that would be more work over in coreaudio-rs. Looking at the changes that *would* go over to coreaudio-rs it isn't as much. Probably just the AudioUnit setup code to create the AudioUnits for the duplex callback.

I could be wrong, but that's my best read for now and my recommendation is to keep here. I'm biased toward minimal changes, and so my recommendation is based on minimizing what would need to change in coreaudio-rs--with the error model changing, that might be too much.

Do you think we should try to do it anyway?

src/host/coreaudio/macos/device.rs
gulbrand: Addressed — instead of pre-allocating the max buffer size, the stream now allocates exactly the current buffer size and fires StreamError::StreamInvalidated if CoreAudio changes it at runtime.

src/host/coreaudio/macos/device.rs
gulbrand: Disconnect propagates `DeviceNotAvailable` via `DisconnectManager`; the stream is paused and the error callback is invoked.

examples/duplex_feedback.rs
gulbrand: Now uses `ChannelCount`, `SampleRate`, and `FrameCount` type aliases.

examples/duplex_feedback.rs
gulbrand: Buffer size is now `Option<FrameCount>`, mapped to `BufferSize::Fixed` or `BufferSize::Default` as the default.

examples/duplex_feedback.rs
gulbrand: using `match`.

examples/duplex_feedback.rs
gulbrand: Changed to `expect` this is demo code so its ok either way.

examples/duplex_feedback.rs
gulbrand: yep, using `Sample::EQUILIBRIUM` and I also updated other locations as well.

examples/duplex_feedback.rs
gulbrand: yep, correct, not necessary. removed.

src/host/coreaudio/macos/device.rs
gulbrand: I also cleaned up many many comments throughout.

src/host/coreaudio/macos/device.rs
gulbrand: done!

src/host/coreaudio/macos/device.rs
gulbrand: done!

src/host/coreaudio/macos/device.rs
gulbrand: yeah I agree. I tried to clean all of them up. The AudioUnit setup code is hard for me to remember so the comments there were for me to keep track of that, but they are not needed, folks can lookup the documentation from apple instead.

src/host/coreaudio/macos/device.rs
gulbrand: yep, done. thanks!

src/host/coreaudio/macos/device.rs
gulbrand: Yes, by not pre-allocating the max size. The code now errors when it detects a buffer size change so the preallocated buffer is now exactly what is required at setup time.

src/host/coreaudio/macos/device.rs
gulbrand: StreamInvalidated seems more appropriate? I went with StreamInvalidted for now. Thoughts?

src/host/coreaudio/macos/device.rs
gulbrand: The code now errors on buffer size changes so pre-allocation is precise at setup time and re-allocation is put into the hands of the host app when it detects this error. That should be all cpal does for now IMHO.

src/host/coreaudio/macos/device.rs
gulbrand: using Box<[u8]> now. thanks!

src/host/coreaudio/macos/device.rs
gulbrand: fixed. I think that was working by accident of using max buffer size, but now it is correct.

src/host/coreaudio/macos/device.rs
gulbrand: yeah, I did some moving around. does this look better?

gulbrand: yeah, I did some moving around. does this look better?

src/host/coreaudio/macos/mod.rs
gulbrand: I answered this on the conversation side of gh ui. At the moment, my best theory is most of the AudioUnit setup etc, but the problem is in the error plumbing. cpal currently benefits from having this here to control the errors, moving that to coreaudio-rs means plumbing errors through. maybe that is ok. for now my recommendation is to leave in cpal until work can be done in coreaudio-rs.

src/duplex.rs
gulbrand: cleaned up

src/duplex.rs
gulbrand: done

src/traits.rs
gulbrand: done

README.md
gulbrand: done

src/host/coreaudio/macos/device.rs
gulbrand: done. I understand this better, but to be totally transparent, I do not have any experience with this part of rust and FFI so I need to lean on your advice here. that said, this does make sense.

src/host/coreaudio/macos/mod.rs
gulbrand: ManuallyDrop worked. I recommend we go with this. One thing I noticed: this still relies on the audio_unit self stopping. That could change in the future, however likely or unlikely that may be, and if it does, boom. That will be a panic or UB I think. I do not see a clear win here, we either rely on stop being idempotent or rely on Drop stopping the audio unit.

src/host/coreaudio/macos/device.rs
gulbrand: I think I lost track of this one, but I blieve this was about sending an error when the device disconnect. The DeviceManager is updated to include buffer change detection. Now that I am looking at all the early returns from the callback I wonder if we need to stop the callback and invoke the error callback. This means adding another channel to pass eror flags to the DeviceManager. Should we do that?

src/host/coreaudio/macos/device.rs
gulbrand: added two module-level utility methods : estimate_capture_isntant and estimate_playback_instant . all streams use these.

gulbrand:
@roderickvd this is ready for another review. A few things to still highlight:

1. The callback has a handful of early returns. As I looked at that again, I realized we might want to signal to the DeviceManager to close the stream as well as invoked_error_callback from the callback. This would mean plumbing another channel to the DeviceManager (I already added one for the buffer size change. Part of me thinks we should do this now, the other part thinks these are rare. I'm inclined to add this in now, but wanted to get your opinion first.
2. Let me know if you think we should try to build this into coreaudio-rs first. The big hang-up there is going to be the error plumbing, that is in DeviceManager. My theory is that this could move to coreaudio-rs but the change would be larger than this to make sure error cases are properly handled.

Thoughts on how to proceed?

Decodetalkers:
I think maybe you need to do git rebase first, to resolve the conflicts

gulbrand:
Conflicts resolved.

src/traits.rs
```diff
@@ -286,6 +298,67 @@ pub trait DeviceTrait {
...
+    ///
+    /// * `config` - The duplex stream configuration specifying channels, sample rate, and buffer size.
+    /// * `sample_format` - The sample format of the audio data.
+    /// * `data_callback` - Called periodically with synchronized input and output buffers as [`Data`].
+    /// * `error_callback` - Called when a stream error occurs (e.g., device disconnected).
+    /// * `timeout` - Optional timeout for backend operations. `None` indicates blocking behavior,
+    ///   `Some(duration)` sets a maximum wait time. Not all backends support timeouts.
+    fn build_duplex_stream_raw<D, E>(
+        &self,
+        _config: &DuplexStreamConfig,
```
Decodetalkers, line 350: no, we do not need & any more, it is copyable

examples/duplex_feedback.rs
roderickvd: I was wrong in my initial comment. `expect(&format!("{x}"))` allocates eagerly. `unwrap_or_else(|| panic!("{x}"))` is better.

examples/duplex_feedback.rs
roderickvd: Since the example is macOS-only at the trait, this is essentially dead code, right? In which case a `compile_error!` may be cleaner.

src/host/coreaudio/macos/device.rs
roderickvd: Thinking about it, `supports_duplex` is interesting in that it should report whether both directions share the same hardware clock. But a USB headset with a microphone and headphones would return `true` for `supports_input() && supports_output()` while not sharing the same clock.

I'm not sure if and how we should do anything about that, other than letting people override the default trait implementation if they can do better in the host.

src/host/coreaudio/macos/device.rs
roderickvd: Yes, better, thanks.

src/host/coreaudio/macos/device.rs
roderickvd: Yes, thanks.

src/host/coreaudio/macos/mod.rs
roderickvd: I'm fine with that. It can be extracted later, if we or others feel that's valuable.

src/host/coreaudio/macos/device.rs
roderickvd: This is now done as per https://github.com/RustAudio/cpal/pull/1096#pullrequestreview-3777170700 right?

src/duplex.rs
roderickvd: All module-level docs are gone now though? 😉

CHANGELOG.md
```diff
@@ -91,6 +95,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Changed
 
+- **POTENTIALLY BREAKING**: `DeviceTrait` now includes `build_duplex_stream()` and `build_duplex_stream_raw()` methods. The default implementation returns `StreamConfigNotSupported`, so external implementations are compatible without changes.
```
roderickvd, line 98: This landed in the wrong section, probably due to rebasing.

You don't need to include "POTENTIALLY BREAKING" - SemVer will be bumped and this is listed under "Changed" so that's good enough for me.

src/duplex.rs
```diff
@@ -0,0 +1,36 @@
+use crate::{ChannelCount, InputStreamTimestamp, OutputStreamTimestamp, SampleRate};
+
+// Timing information for a duplex callback, combining input and output timestamps.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+pub struct DuplexCallbackInfo {
```
roderickvd, line 5: For these and other public types, please ensure that they have public Rustdoc.

src/duplex.rs
roderickvd, line 1: The module should probably have some short Rustdoc, as it's private. Compare with `InputCallbackInfo`/`OutputCallbackInfo` for tone and verbosity.

src/duplex.rs
```diff
@@ -0,0 +1,36 @@
...
+        self.output_timestamp
+    }
+}
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub struct DuplexStreamConfig {
+    pub input_channels: ChannelCount,
+    pub output_channels: ChannelCount,
+    pub sample_rate: SampleRate,
+    pub buffer_size: crate::BufferSize,
```
roderickvd, line 35: Nitpick: may be imported for consistency with the other types.

src/host/coreaudio/macos/device.rs
roderickvd: Ah, I see you already put that in the public docs.

src/host/coreaudio/macos/mod.rs
```diff
@@ -164,15 +195,50 @@ impl DisconnectManager {
             }
         });
 
+        if listen_buffer_size {
+            let stream_weak_clone = stream_weak.clone();
+            let error_callback_clone = error_callback.clone();
+            std::thread::spawn(move || {
```
roderickvd, line 201: Perhaps capture the relevant discussion between @Decodetalkers and you earlier:

```rust
// thread handle intentionally dropped; same pattern as disconnect handler 
```

src/host/coreaudio/macos/device.rs
```diff
@@ -260,6 +265,48 @@ fn get_io_buffer_frame_size_range(
...
 
+pub(super) fn estimate_capture_instant<E>(
+    callback_instant: StreamInstant,
+    delay: Duration,
+    error_callback: &Mutex<E>,
+) -> StreamInstant
+where
+    E: FnMut(StreamError) + Send,
+{
+    callback_instant.sub(delay).unwrap_or_else(|| {
```
roderickvd, line 276: I'm soon going to merge #1139 which will change the `StreamInstant` API somewhat. Here's what you might then replace it with on CoreAudio: https://github.com/RustAudio/cpal/pull/1139/changes#diff-8e684b369b47b2adec5fdeb79b478a2533366a37c684eb9a1e686abed4939bc6

src/host/coreaudio/macos/duplex.rs
```diff
@@ -0,0 +1,352 @@
...
+        });
+
+        let stream = Stream::new(inner, error_callback_for_stream, true)?;
+
+        stream
+            .inner
+            .lock()
+            .map_err(|_| BuildStreamError::BackendSpecific {
+                err: BackendSpecificError {
+                    description: "A cpal stream operation panicked while holding the lock - this is a bug, please report it".to_string(),
```
roderickvd, line 344: I know it's been like this in other places, but I think that we should propagate poisoned mutexes. I'll probably submit a PR to do so in other places too.

README.md
```diff
@@ -174,7 +175,7 @@ If you are unable to build the library:
 
 ## Examples
 
-CPAL comes with several examples in `examples/`.
+CPAL comes with several examples in `examples/`, including `duplex_feedback` for hardware-synchronized duplex streams.
```
roderickvd, line 178: No need to be so specific.

roderickvd (changes requested):
This is starting to look great. Almost ready to merge.

roderickvd:
> 1. The callback has a handful of early returns. As I looked at that again, I realized we might want to signal to the DeviceManager to close the stream as well as invoked_error_callback from the callback. This would mean plumbing another channel to the DeviceManager (I already added one for the buffer size change. Part of me thinks we should do this now, the other part thinks these are rare. I'm inclined to add this in now, but wanted to get your opinion first.

That sounds nice. Would it need _another_ channel or could we use a single channel and send an `enum` over it with various payload?

> 2. Let me know if you think we should try to build this into coreaudio-rs first. The big hang-up there is going to be the error plumbing, that is in DeviceManager. My theory is that this could move to coreaudio-rs but the change would be larger than this to make sure error cases are properly handled.

As you wish, though I'd be fine by doing it in `cpal` now and extracting it into `coreaudio-rs` later, should you or others feel so inclined.

roderickvd:
Don't want to cause you too many rebasing headaches, so want so swing this by you first: #1147 refactored the listener somewhat. From a glance that should be compatible with yours - but don't hesitate to let me know.

gulbrand:
@roderickvd 

I made progress on this, but I'd like to propose a multi PR process. I opened a new PR https://github.com/RustAudio/cpal/pull/1229 just for the API. 

I plan to close https://github.com/RustAudio/cpal/pull/1096 in favor of the above PR and a new PR I'm working on--the merge conflicts were a bit too much for me :) . My own fault for sitting on this for so long.

I'm hoping this API PR can merge quickly so we can at least lock that in.

I also have a proof of concept with ASIO on windows just to increase confidence that this API is in the right direction: https://github.com/gulbrand/cpal/pull/5 . Keep in mind I am only looking to make sure the API works with ASIO, not that the code is correct. Huge difference :). I am getting a wnidows laptop up and running to do testing.

gulbrand:
Closing in favor of: https://github.com/RustAudio/cpal/pull/1229

Boscop:
@gulbrand Thanks so much for working on this, your branch `duplex-support` works well for me on macos, I'm planning to use it for music live performance.

Would you say it's rock-solid / suitable for production usage? :)

---

Btw, since you started splitting up the work into different branches/PRs, which branch do you recommend using for now (until all your duplex branches get merged), in other words, which branch will contain all your duplex related changes going forward? :)

gulbrand:
> @gulbrand Thanks so much for working on this, your branch `duplex-support` works well for me on macos, I'm planning to use it for music live performance.
> 
> Would you say it's rock-solid / suitable for production usage? :)
> 
> Btw, since you started splitting up the work into different branches/PRs, which branch do you recommend using for now (until all your duplex branches get merged), in other words, which branch will contain all your duplex related changes going forward? :)

@Boscop here is the branch I am currently using https://github.com/gulbrand/cpal/tree/macos . 

This hasn't been proven through usage that it is production ready. I think the code is correct and the RustAudio group have helped me identify improvements I've made. I would love to hear if it isn't working for you--maybe try it in a low-risk situation first? And have a back-up in a high stakes situation -- e.g. a live performance.

I can say that the low latency is quite nice. I've been using this in my DAW and making music and I haven't had any problems so far. However, your mileage may vary.

Issue #593, opened 2021-07-12
`build_output_stream` fails on Windows 10 if the specified sample rate does not match the output device's default sample rate


Herohtar:
For context, the Windows audio device properties allow you to configure the default sample rate and bit depth that should be used when an audio device is running in shared mode:

![image](https://user-images.githubusercontent.com/10427164/125234256-f30b0900-e2a5-11eb-8f83-f8072e455baa.png)

If you try to create an output stream in `cpal` using the `build_output_stream` method and the specified sample rate does not match the sample rate defined in the device properties, the method fails.

Example code:
```rust
use cpal::traits::{DeviceTrait, HostTrait};

fn main() {
    let host = cpal::default_host();

    let device = host.default_output_device().unwrap();

    device.build_output_stream(
        &cpal::StreamConfig {
            channels: cpal::ChannelCount::from(2u16),
            sample_rate: cpal::SampleRate(48000), // Audio device default sample rate is set to 192000
            buffer_size: cpal::BufferSize::Default,
        },
        move |_: &mut [f32], _| {},
        move |_| {},
    ).unwrap();
}
```

Result:
```text
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: StreamConfigNotSupported', src\main.rs:16:7
```

The error can be resolved in one of two ways:
1. Change the sample rate in `StreamConfig` to 192000 to match the device setting
2. Change the device setting to 48000 to match the sample rate in the code

Am I correct in believing that this is a bug, and you should be able to create an output stream using any sample rate, or is there some limitation that prevents that from being possible?

leecbaker:
I'm seeing a very similar situation on Mac OS- the only supported formats are high sample rate floating point formats. My use case needs a much lower sample rate, and it doesn't appear to be available through this API.

Ralith:
> you should be able to create an output stream using any (supported) sample rate

In the case of Windows, it appears that the device's configured sample rate is the only supported one. In general, it is the application's responsibility to query [supported_output_configs](https://docs.rs/cpal/latest/cpal/platform/struct.Device.html#method.supported_output_configs) and select a valid configuration.

dragos-vlad:
[...]

PR #1097, opened 2026-01-23
wasapi: Enable resampling and rate adjustment

yeah-its-gloria:
this should allow resampling sample rates that are too quirky for native hardware by the WASAPI server :+1:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `DeviceBusy` error variant for retriable device access errors (EBUSY, EAGAIN).
 - **ALSA**: `Debug` implementations for `Host`, `Device`, `Stream`, and internal types.
 - **ALSA**: Example demonstrating ALSA error suppression during enumeration.
+- **WASAPI**: Allow non-native sample rates to be used via as-necessary resampling in the WASAPI server process.
 
 ### Changed
 


yeah-its-gloria:
~~Currently this still fails because of format checks, I will look into this and see how I can fix this~~ solved this

kaezrr:
I have tested this and this works, it allows me to request custom sampling rates in windows that was otherwise locked to only one single sampling rate conversion. 

This solves issue #593 that has been open for almost 5 years now.

@roderickvd you might remember me from the Rust Audio discord, i was ranting about the resampling issue on Windows and you recommended me to use rodio instead.

I can vouch that this PR works, you might be interested to take a look

src/host/wasapi/device.rs
```diff
@@ -61,6 +61,11 @@ const PKEY_AUDIOENDPOINT_JACKSUBTYPE: PROPERTYKEY = PROPERTYKEY {
     pid: 8,
 };
 
+const DEFAULT_FLAGS: u32 = Audio::AUDCLNT_STREAMFLAGS_EVENTCALLBACK
+    | Audio::AUDCLNT_STREAMFLAGS_RATEADJUST
```
roderickvd: `AUDCLNT_STREAMFLAGS_RATEADJUST` shouldn't be necessary, right? That's only when we call `IAudioClockAdjustment::SetSampleRate()` for clock drift correction, and I'm not sure the `wasapi` crate even exposes that.

src/host/wasapi/device.rs
```diff
@@ -696,7 +705,7 @@ impl Device {
             // will return `AUDCLNT_E_BUFFER_SIZE_ERROR` if the buffer size is not supported.
             let buffer_duration = buffer_size_to_duration(&config.buffer_size, config.sample_rate);
 
-            let mut stream_flags = Audio::AUDCLNT_STREAMFLAGS_EVENTCALLBACK;
+            let mut stream_flags = DEFAULT_FLAGS;
```
roderickvd, line 685: Instead of always using `DEFAULT_FLAGS`, it'd be nice if we could check if the requested sample rate is the default sample rate. If it is, then we only need `AUDCLNT_STREAMFLAGS_EVENTCALLBACK` without inserting any resampling.

src/host/wasapi/device.rs
```diff
@@ -187,10 +192,12 @@ unsafe fn data_flow_from_immendpoint(endpoint: &Audio::IMMEndpoint) -> Audio::ED
 
 // Given the audio client and format, returns whether or not the format is supported.
 pub unsafe fn is_format_supported(
-    client: &Audio::IAudioClient,
-    waveformatex_ptr: *const Audio::WAVEFORMATEX,
+    _client: &Audio::IAudioClient,
+    _waveformatex_ptr: *const Audio::WAVEFORMATEX,
 ) -> Result<bool, SupportedStreamConfigsError> {
-    // Check if the given format is supported.
+    // Checking formats is not needed for shared mode with auto-conversion, therefore this check has been removed until someone implements WASAPI exclusive mode support
```
roderickvd, line 196: Let's indeed document that we choose to auto-resample because shared mode requires it. At the same time let's remove code that you commented out.

yeah-its-gloria:
Apparently [PulseAudio](https://freedesktop.org/software/pulseaudio/doxygen/sample.html) transparently converts formats already, and PipeWire requires you to attach SPAs to your stream for conversion regardless, so I don't see the particular issue with automatic resampling here.
The main reason `AUDCLNT_STREAMFLAGS_AUTOCONVERTPCM` isn't standard is that it wasn't a thing in WASAPI until Windows 10 - even though WASAPI (and shared mode being default!) was part of Windows since Vista.

> catching an open stream error and falling back to the default stream configuration instead, then doing resampling themselves

Resampling can be annoying to implement, and APOs might do a better job at it (by, for example, using the sound hardware).

> using Rodio which is intended as a batteries-included audio crate, unlike cpal which is more low-level

I don't think low-level APIs should just ignore features implemented by the APIs beneath them (look right below).

> using extension traits in the future (see https://github.com/RustAudio/cpal/pull/1074 for an example), that could allow users to explicitly opt-in to resampling behavior

Once implemented, sure thing.

src/host/wasapi/device.rs
yeah-its-gloria: No resampling is done when the rates match, the flags become a no-op in that case.

src/host/wasapi/device.rs
yeah-its-gloria: I saw the flags used in SDL, which, weirdly enough, also doesn't use `IAudioClockAdjustment` anywhere. I copied them without looking deeper into it, sorry.

roderickvd:
Merged following our Discord discussion, thanks!

PR #1098, opened 2026-01-27
Safe join on Drop

Roms1383:
This PR refactor thread `join` on `Drop` to avoid `unwrap`.

While it's probably unnecessary in most contexts, I depend on `cpal` through [kira](https://github.com/tesselode/kira) in [audioware](https://github.com/cyb3rpsych0s1s/audioware) which runs in a video game plugin context.

In this particular context, the game can reclaim plugin's resource anytime (e.g. when player kill game in Windows taskbar), which, at times, causes race condition leading to CTD like this:
Screenshot 2026-01-27 174944

### DIFF
--- a/src/host/wasapi/device.rs
+++ b/src/host/wasapi/device.rs
@@ -511,7 +511,7 @@ impl Device {
         // Retrieve the `IAudioClient`.
         let lock = match self.ensure_future_audio_client() {
             Ok(lock) => lock,
-            Err(ref e) if e.code() == Audio::AUDCLNT_E_DEVICE_INVALIDATED.into() => {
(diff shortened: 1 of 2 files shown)


Roms1383:
By the way errors in CI is because `rustc` `1.93` requires the changes from [4bc4494](https://github.com/RustAudio/cpal/pull/1098/commits/4bc449427792d62449fbd3c131e2d512bf158681), but I can revert the commit if you want.

Roms1383:
I took the liberty to revert the changes from [4bc4494](https://github.com/RustAudio/cpal/pull/1098/commits/4bc449427792d62449fbd3c131e2d512bf158681) since I'm concerned a failing CI won't attract your attention 😸

roderickvd:
Good one, thanks!

Issue #969, opened 2025-04-15
`Oboe` returns duplicate devices


setoelkahfi:
<details>
  <summary>Click me</summary>

![Image](https://github.com/user-attachments/assets/7fb5345d-edd5-4cd1-bb17-5f685faac640)

</details>

This `available_hosts()` works correctly in other platforms, but Android returns duplicate entries. I investigated it and it turns out it calls the [getDevices](https://developer.android.com/reference/android/media/AudioManager#getDevices(int)) in the `AudioManager` API. The cpal implementation `utils.rs` line 168:
```

pub fn get_devices<'j>(
    env: &mut JNIEnv<'j>,
    subject: &JObject<'j>,
    flags: i32,
) -> JResult<JObjectArray<'j>> {
    env.call_method(
        subject,
        "getDevices",
        "(I)[Landroid/media/AudioDeviceInfo;",
        &[flags.into()],
    )?
    .l()
    .map(From::from)
}

```
Which is called from:

```

fn try_request_devices_info<'j>(
    env: &mut JNIEnv<'j>,
    context: &JObject<'j>,
    direction: AudioDeviceDirection,
) -> JResult<Vec<AudioDeviceInfo>> {
    let audio_manager = get_system_service(env, context, Context::AUDIO_SERVICE)?;

    let devices = get_devices(env, &audio_manager, direction as i32)?;

    let length = env.get_array_length(&devices)?;

    (0..length)
        .map(|index| {
            let device = env.get_object_array_element(&devices, index)?;
            let id = call_method_no_args_ret_int(env, &device, "getId")?;
            let address = call_method_no_args_ret_string(env, &device, "getAddress")?;
            let address = String::from(env.get_string(&address)?);
            let product_name =
                call_method_no_args_ret_char_sequence(env, &device, "getProductName")?;
            let product_name = String::from(env.get_string(&product_name)?);
            let device_type =
                FromPrimitive::from_i32(call_method_no_args_ret_int(env, &device, "getType")?)
                    .unwrap_or(AudioDeviceType::Unsupported);
            let direction = AudioDeviceDirection::new(
                call_method_no_args_ret_bool(env, &device, "isSource")?,
                call_method_no_args_ret_bool(env, &device, "isSink")?,
            );
            let channel_counts =
                call_method_no_args_ret_int_array(env, &device, "getChannelCounts")?;
            let sample_rates = call_method_no_args_ret_int_array(env, &device, "getSampleRates")?;
[...]

PR #1103, opened 2026-02-15
Fix aaudio device name in `device.description().name()`

SuperKenVery:
#965 fixed this for `device.name()`, which is now a deprecated method. I let `.description().name()` return the same name.

### DIFF
--- a/src/host/aaudio/mod.rs
+++ b/src/host/aaudio/mod.rs
@@ -405,7 +405,7 @@ impl DeviceTrait for Device {
         match &self.0 {
             None => Ok(DeviceDescriptionBuilder::new("Default Device".to_string()).build()),
             Some(info) => {
-                let mut builder = DeviceDescriptionBuilder::new(info.product_name.clone())
+                let mut builder = DeviceDescriptionBuilder::new(self.name()?)
                     .device_type(info.device_type.into())
                     .interface_type(info.device_type.into())
                     .direction(info.direction);
--- a/src/host/aaudio/mod.rs
+++ b/src/host/aaudio/mod.rs
@@ -405,7 +405,15 @@ impl DeviceTrait for Device {
         match &self.0 {
             None => Ok(DeviceDescriptionBuilder::new("Default Device".to_string()).build()),
             Some(info) => {
-                let mut builder = DeviceDescriptionBuilder::new(self.name()?)
+                let name = if info.address.is_empty() {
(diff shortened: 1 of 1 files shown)


roderickvd:
I'm not super-clear on the intent, can you elaborate? It seems that currently returning `info.product_name` is the same that #965 did?

SuperKenVery:
In #965, the name is changed to product name and device type, on android. Product name is always the device model.

You may see the code diff, it's a one line change, and the implementation of name is a few lines above.

roderickvd:
If I understand correctly, you are saying that you prefer the behavior of #965? Because I found it weird that a "name" includes a device type and an address.

SuperKenVery:
Yes on android. The product name on android is same for every device; it's always the phone product name. 

Screenshot_2026-02-19-12-14-31-66_b77199ee68e84a36b2a4aee65dc48fbd

Without this pr, all of them will be PJX110

setoelkahfi:
I didn't see this before working on fixing https://github.com/RustAudio/cpal/issues/969 I have my own patch locally. What's the status here?

SuperKenVery:
@setoelkahfi seems we are doing the same thing, with a little bit different implementation details 😉

setoelkahfi:
> @setoelkahfi seems we are doing the same thing, with a little bit different implementation details 😉

@SuperKenVery would you mind using my changes instead?

SuperKenVery:
@setoelkahfi My version ensures that `device.name()` returns the same value as `device.description().name()`, would this be preferred?

 cc @roderickvd

setoelkahfi:
> @setoelkahfi My version ensures that `device.name()` returns the same value as `device.description().name()`, would this be preferred?
> 
> cc @roderickvd

My changes are [here](https://github.com/RustAudio/cpal/pull/1109). I also filter out the devices based on the direction (output, input). 

The CI are failing but don't seem to be related to my changes.

setoelkahfi:
@SuperKenVery do we agree to use my changes instead? I'd like to see it merged. 

@roderickvd do you need more information about the issue/changes?

roderickvd:
Let's continue at #1109. @SuperKenVery please provide any reviews or comments there - thanks to both of you!

SuperKenVery:
Thanks, it seems I missed the email notifications in a flood of dependbot messages. Definitely happy to see your pr merged!

PR #1108, opened 2026-02-19
chore: make StreamConfig Copyable

Decodetalkers:


Decodetalkers:
maybe now instead of &StreamConfig, it should be changed to StreamConfig

roderickvd:
> maybe now instead of &StreamConfig, it should be changed to StreamConfig

I agree, let's do that and add appropriate changelog entries.
Next version will be v0.18 anyway, so API breakage is acceptable.

roderickvd:
Do you also want to change the parameters from `&StreamConfig` to `StreamConfig` (and document so in the changelog)?

Decodetalkers:
> Do you also want to change the parameters from `&StreamConfig` to `StreamConfig` (and document so in the changelog)?

Yes, but I want to wait for pipewire being finished first.

roderickvd:
OK, let me know when you're ready for review!

Decodetalkers:
@roderickvd done! I think it is ready

PR #1111, opened 2026-02-22
fix: emscripten build failed after new wasm-bindgen is published

Decodetalkers:


roderickvd:
I see I've got write access on your repo - thanks - I'll update and merge it.

PR #1115, opened 2026-03-01
fix(pulseaudio): several correctness and accuracy fixes

roderickvd:
A few issues I spotted in the PulseAudio backend after it landed. @colinmarc, would you mind casting an eye over these since you know the code best?

First commit contains hot-path and correctness fixes:
- `U8` silence is `0x80`, not `0x00`
- `store_latency` had a sign bug: casting a negative `write_offset - read_offset` to `u64` wrapped around
- Record streams were reading `sink_usec` instead of `source_usec`
- Worker thread errors from `play_all()` were silently dropped
- Latency poll interval reduced from 100 ms to 5 ms

Second commit improves channel maps, buffer configuration, and timestamps:
- Channel map was mapping everything except stereo to `Mono`; now implements standard PA layouts for 1–8 channels, 7.1.2 and 7.1.4 Dolby Atmos for 10 and 12 channels, and sequential `Aux` positions as a fallback
- `make_buffer_attr` split into separate playback and record functions: playback uses a double-buffer (`target_length` = 2N, `minimum_request_length` = N) to guarantee N-frame callbacks; record uses `fragment_size` instead of the playback-only `target_length` field
- `adjust_latency` flag now set when `BufferSize::Fixed` is requested, so PulseAudio configures the hardware end-to-end to hit the target
- `SystemTime` replaced with `Instant` (monotonic, not affected by NTP); playback timestamps now interpolate latency between polls rather than using a stale value for up to 5 ms

### DIFF
--- a/src/host/pulseaudio/mod.rs
+++ b/src/host/pulseaudio/mod.rs
@@ -280,7 +280,8 @@ impl DeviceTrait for Device {
 
         let sample_spec = make_sample_spec(config, format);
         let channel_map = make_channel_map(config);
-        let buffer_attr = make_buffer_attr(config, format);
+        let buffer_attr = make_record_buffer_attr(config, format);
+        let adjust_latency = matches!(config.buffer_size, crate::BufferSize::Fixed(_));
 
         let params = protocol::RecordStreamParams {
             sample_spec,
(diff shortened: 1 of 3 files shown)


roderickvd:
I'm on a merge streak and would like to get this in so it gets more testing.
@colinmarc if you see any issues with this, then do not hesitate to open an issue.

PR #1118, opened 2026-03-02
ASIO: Add COM initialization

### LINKED ISSUE #798: [ASIO] bug/strange behaviour with threads - only one thread can access devices
Here's an interesting example (assuming Windows 11 + ASIO4ALLv2 as a platform):

```
use std::thread;
use anyhow;
use cpal::{
    traits::{DeviceTrait, HostTrait},
};

fn main() -> anyhow::Result<()> {

    // the problem doesn't occur with Wasapi, neither with other
    // hosts on other platforms ...
    // let host = cpal::host_from_id(cpal::HostId::Wasapi).unwrap();

    // if this loop is commented out, the spawned thread finds the "ASIO4ALL v2" device,
    // otherwise the list is empty ...
    for dev in host.output_devices()? {
	println!("MAIN THREAD - found device {}", dev.name().unwrap());
    }

    let handler = thread::spawn(move || {
	if let Ok(devices) = host.output_devices() {
	    let mut dev_count = 0;
	    for dev in devices {
		dev_count += 1;
		println!("SPAWNED THREAD - found device {}", dev.name().unwrap());
	    }
	    println!("SPAWNED THREAD - found {dev_count} devices");
	} else {
	    println!("SPAWNED THREAD - can't find devices");
	}	
    });

    handler.join().unwrap();

    Ok(())
}
```

I'd expect (as this doesn't seem to be a problem at all with other hosts) that both the main thread and the spawned thread can access the list of devices. 

Turns out, this isn't the case. A list of devices is returned, but it's empty. If I comment out the statements in the main thread, the spawned thread can find all devices without problems.

Within the same thread, I can access the list of devices as often as I want.

I can't really explain this ... does anyone have a hint? 

Best,
N

Sin-tel:
Add COM initialization code to ASIO, which was already available in WASAPI.

TODO: figure out where exactly we need to sprinkle `com::com_initialized()` to make everything work correctly.

Fixes #798

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Reintroduce `audio_thread_priority` feature.
 - **ASIO**: Fix enumeration returning only the first device when using `collect`.
+- **ASIO**: Fix device enumeration and stream creation failing when called from spawned threads.
 - **Emscripten**: Fix build failure introduced by newer `wasm-bindgen` versions.
 
 ## [0.17.3] - 2026-02-18


roderickvd:
Thanks for sorting this out. Just give me a sign when you've completed that todo.

Sin-tel:
Adding them to stream should cover all cases. Not necessary in Device since all data is cached anyway.

PR #1120, opened 2026-03-02
refactor(pipewire): post-merge follow-ups

roderickvd:
- Stop capture stream on format mismatch in param_changed; the input closure was discarding the stream reference (`_`) so `set_active(false)` was never called, unlike the output handler
- Scope `fill_with_equilibrium` to the active buffer region (frames*stride) rather than the full mapped buffer, avoiding writes to unused memory
- Simplify `pw_stream_time` return from `Option<(StreamInstant, PwTime)>` to `Option<PwTime>`; callers construct both `StreamInstant`s from `now_ns`
- Change `assert_eq!` to `debug_assert_eq!` in `fill_with_equilibrium`, which runs in the RT process callback on every output cycle


roderickvd:
cc @Decodetalkers 

I'll merge this when the CI passes.

Issue #956, opened 2025-02-23
SupportedStreamConfig(Range) API inconsistencies - can we have a `SupportedStreamConfig.try_with_fixed_buffer_size`?


0xDA41:
The `SupportedStreamConfigRange`, `SuportedStreamConfig` and `StreamConfig` types make configuring streams in cpal far more complicated and unpleasant than it would need to be.

- `SupportedStreamConfigRange` - range of sampling rates and buffer sizes
- `SupportedStreamConfig` - fixed sampling rate, range of buffer sizes
- `StreamConfig` - fixed sampling rate, fixed or 'default' buffer size

I'm not sure why there's even a need for `SupportedStreamConfig`, but the main point why I opened this issue is that the naively expected chain of `Iterator<SupportedStreamConfigRange>` -> `SupportedStreamConfigRange` -> `SupportedStreamConfig` -> `StreamConfig` is incomplete.

- I get the iterator, and filter it to get one or more candidates.
- Then on those, I call `try_with_sample_rate` to get a `SupportedStreamConfig`. This is where the first problem is: bad documentation: `/// Retrieve a [SupportedStreamConfig] with the given sample rate and buffer size.` The signature of the method is `(self, sample_rate: SampleRate)`, no buffer size I could specify.
- Getting a `SupportedStreamConfig`, I'd then like to narrow down the buffer size, as at least on Linux, most devices return with a buffer size range of between 1 and MAXINT32 or so. However, there is no method for doing this. As mentioned in #401, the `config` method on this type just returns a `StreamConfig` with a `Default` buffersize setting.

So in the end, I'm still hand-assembling a `StreamConfig`.

I guess the first step would be to have a `try_with_fixed_buffer_size` method on `SupportedStreamConfig`. Then, if there's no good reason for it, it might be meaningful to roll SupportedStreamConfigRange and SupportedStreamConfig into a single type?

roderickvd:
You are right. And when we're on it, I'm thinking if we'd take that first step with `try_with_fixed_buffer_size` or immediately the next and introduce a builder? Claude came up with this, nice and additive:

```rust
/// Builder for creating a validated StreamConfig
pub struct StreamConfigBuilder {
    channels: ChannelCount,
    sample_rate: SampleRate,
    sample_format: SampleFormat,
    buffer_size: BufferSize,

    // Constraints from the supported config
    sample_rate_range: (SampleRate, SampleRate),
    buffer_size_range: SupportedBufferSize,
}

[...]

PR #1121, opened 2026-03-03
feat: add build_pattern for StreamConfig

Decodetalkers:
We always use StreamConfig by `into` from SupportedStreamConfig, and if we want to set the field in StreamConfig, we need firstly set the StreamConfig to mutable, then give the field a new value to it.

This pr induced three useful function to solve this problem, now we can set these properties in the way like
```rust
let config = supported_config
								.into()
								.with_fixed_buffer(1024);
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -400,6 +400,24 @@ pub struct StreamConfig {
     pub buffer_size: BufferSize,
 }
 
+impl StreamConfig {
+    /// This function set the channels in [StreamConfig]
+    pub fn with_channels(mut self, channels: ChannelCount) -> Self {
+        self.channels = channels;
+        self
+    }
+    /// This function set the buffer_size in [StreamConfig] to a fixed size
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -400,6 +400,24 @@ pub struct StreamConfig {
     pub buffer_size: BufferSize,
 }
 
+impl StreamConfig {
+    /// This function set the channels in [StreamConfig]
+    pub fn with_channels(mut self, channels: ChannelCount) -> Self {
```
roderickvd: All these functions should have a `#[must_use]` tag.

roderickvd:
The ergonomics with `SupportedStreamConfig` indeed are clumsy and discussed over at #956.

Do we then need this PR? Because we can already do the following today:
```rust
let config = StreamConfig {
    buffer_size: BufferSize::Fixed(256),
    ..supported_config.into()
};
```

I propose that we continue over at #956 to find the right API.

Issue #876, opened 2024-04-16
Suppport ScreenCaptureKit loopback


thewh1teagle:
In recent versions of `macOS` there's new API called [screen capture kit](https://developer.apple.com/documentation/screencapturekit/) which provides a way to capture the screen content (Audio / Video),
That way we can receive audio from the speakers, similar how `cpal` does that with WSAPI on Windows (Added in #251)
As starting point we can follow the approach used in the project [scap](https://github.com/helmerapp/scap) in [engine/mac/mod.rs](https://github.com/helmerapp/scap/blob/main/scap/src/capturer/engine/mac/mod.rs) they capture audio / video using `screen capture kit`

LatentDream:
+1

rustdesk:
+1

rustdesk:
Maybe, we can add a virtual device called loopback, which use screencapturekit, the others keep no change

louis030195:
+1

SuperKenVery:
ScreenCaptureKit has a relatively high latency, in my previous experience. Maybe we could have a look at [CoreAudio Tap API](https://developer.apple.com/documentation/coreaudio/audiohardwarecreateprocesstap(_:_:)?language=objc)

[Example usage](https://gist.github.com/sudara/34f00efad69a7e8ceafa078ea0f76f6f)

SuperKenVery:
Hey guys I implemented loopback recording on macOS with CoreAudio Tap API. 

Here's my pr: #1003

thewh1teagle:
Hey @SuperKenVery! I tested it on macOS 26.3 (apple chip) but unfortunately it doesn't work (it's recording silence)

thewh1teagle:
I will just use this fork https://github.com/rustdesk-org/cpal/tree/osx-screencapturekit

SuperKenVery:
@thewh1teagle Probably permission issue: https://github.com/RustAudio/cpal/pull/894#issuecomment-3823323664 (see last section)

thewh1teagle:
Thank you @SuperKenVery! you were right it recorded silence just because of missing permission. 
I made this small crate https://github.com/thewh1teagle/vibe/tree/main/crates/macos-permissions so I can ask/check permission and it works great now

thewh1teagle:
actually on prod build I get this error first time I allow it

```console
The requested device is no longer available. For example, it has been unplugged.

Location:
    desktop/src-tauri/src/cmd/audio.rs:215:18
```

[...]

PR #1123, opened 2026-03-05
fix(coreaudio): declare aggregate_device_id as mut and check AudioHardwareCreateProcessTap status

thewh1teagle:
I was using cpal 0.17.3 for loopback recording on macOS and getting `The requested device is no longer available` in release builds.

  Two bugs in `loopback.rs`:

  1. `aggregate_device_id` was not declared `mut` — UB when the compiler optimizes away the write in release mode, leaving the device ID as 0.
  2. `AudioHardwareCreateProcessTap` status was ignored — silent failure leads to invalid tap ID.

  Fixed by adding `mut` to `aggregate_device_id`, using `NonNull::from(&mut aggregate_device_id)`, and propagating the tap creation status via `check_os_status`.

Related

- https://github.com/RustAudio/cpal/issues/876#issuecomment-4002140096


roderickvd:
Thank you! Merging now.

PR #1124, opened 2026-03-05
feat(coreaudio): add macOS system audio permission helpers

thewh1teagle:
I was using cpal 0.18.0 (master) for loopback recording on macOS and needed to check and request the `System Audio Recording` permission before recording. Without it, recording succeeds but
  captures only silence.

  Added three public functions on macOS:
  - `cpal::check_system_audio_permission()` — silent check, no UI
  - `cpal::request_system_audio_permission()` — shows system prompt, blocking
  - `cpal::open_system_audio_settings()` — opens Privacy & Security > System Audio Recording

  Uses the private TCC framework (`kTCCServiceAudioCapture`) via `libloading`. `block2` and `objc2-core-foundation` were already transitive dependencies so no significant dep overhead.

Related

- https://github.com/RustAudio/cpal/issues/876#issuecomment-4002027224

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -139,6 +139,9 @@ objc2 = { version = "0.6" }
 
 [target.'cfg(target_os = "macos")'.dependencies]
 jack = { version = "0.13", optional = true }
+libloading = "0.8"
+block2 = "0.6"
+objc2-core-foundation = "0.3"
 
 [target.'cfg(target_os = "ios")'.dependencies]
 objc2-avf-audio = { version = "0.3", default-features = false, features = [
--- a/src/host/coreaudio/macos/mod.rs
+++ b/src/host/coreaudio/macos/mod.rs
@@ -20,6 +20,7 @@ use property_listener::AudioObjectPropertyListener;
 mod device;
 pub mod enumerate;
 mod loopback;
+pub mod permissions;
 mod property_listener;
 pub use device::Device;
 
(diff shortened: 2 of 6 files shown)


SuperKenVery:
I think it might be better if we check permission when recording from an output device, and if we don't have it, request it. 

Do you think it's better? cc @roderickvd 

If you agree, I can file a PR agains your branch!

thewh1teagle:
Feel free to create a PR against my branch or open another PR with me as a co-author :)

roderickvd:
Thinking out loud:

Since requesting permissions is a blocking operation showing a system-modal dialog, how about taking a middle road? In `LoopbackDevice::from_device()` we only `check_system_audio_permission()` which is non-blocking. If we don't have permissions, we return a new error variant. This catches the silent-silence bug without any blocking UI side-effects.

Then, we keep `request_system_audio_permission()` and `open_system_audio_settings()` as public API, so that a user can explicitly decide to request a blocking modal.

Going forward I would like to expose host-specific functionality with extension traits. Please refer to #1074 for an ongoing discussion how to do that for ASIO (getting the API right for _all_ hosts) as well as an outlook on how to expose new error variants.

thewh1teagle:
Hey, what's the status of this feature? sorry I accidentally deleted my fork

roderickvd:
I was awaiting a reply from @SuperKenVery or you after my last suggestions.
Meanwhile, I'm cooking up a proposal for host-specific functionality to be posted as a Gist for discussion soon.

SuperKenVery:
Oh sorry for the delay, I wasn't aware that you were waiting for my reply!

As for the design, I think @roderickvd 's suggestion is perfect. Currently, this PR's implementation is quite macOS-specific and didn't take cross-platform into account much. A non-blocking check + raise error and a blocking request permission api seems very good shape for future platforms e.g. pipewire!

PR #1125, opened 2026-03-05
fix(aaudio): Wait for AAudioStream status change

osoftware:
This PR fixes two bugs at once.

After starting the stream and before the status changes to `Started`, dropping it causes the main thread to lock permanently. The underlying `AAudioStream` never closes while the `drop` method waits for it to close. By using `wait_for_state_change` method we ensure the `play` method won't return befor it's safe to drop the stream.

If `wait_for_state_change` is never called then `AAudioStream_getTimestamp` always returns `InvalidState` error leading to `info.timestamp().capture` and `info.timestamp().playback` in stream callbacks always returning zero.

### DIFF
--- a/src/host/aaudio/mod.rs
+++ b/src/host/aaudio/mod.rs
@@ -700,12 +700,18 @@ impl StreamTrait for Stream {
 
     fn pause(&self) -> Result<(), PauseStreamError> {
         match self.direction {
-            DeviceDirection::Output => self
-                .inner
-                .lock()
-                .unwrap()
-                .request_pause()
-                .map_err(PauseStreamError::from),
+            DeviceDirection::Output => {
+                let stream = self.inner.lock().unwrap();
+
+                stream.request_pause().map_err(PauseStreamError::from)?;
+                stream
+                    .wait_for_state_change(
+                        ndk::audio::AudioStreamState::Pausing,
+                        DEFAULT_TIMEOUT_NANOS,
+                    )
+                    .map(|_| ())
+                    .map_err(PauseStreamError::from)
+            }
             _ => Err(BackendSpecificError {
                 description: "Pause only supported on output streams.".to_owned(),
             }


src/host/aaudio/mod.rs
```diff
@@ -598,18 +598,18 @@ impl DeviceTrait for Device {
...
+        let stream = match self {
+            Self::Input(stream) => stream,
+            Self::Output(stream) => stream,
         }
+        .lock()
+        .unwrap();
+
+        stream.request_start().map_err(PlayStreamError::from)?;
+        stream
+            .wait_for_state_change(ndk::audio::AudioStreamState::Starting, 300_000_000)
```
roderickvd: Not an expert in AAudio but isn't Oboe's `kDefaultTimeoutNanos` equal to `2_000_000_000`? Would that be worth using and documenting? If there is a reason why `300_000_000` would be more appropriate, that magic number would also be deserving of a comment.

src/host/aaudio/mod.rs
```diff
@@ -598,18 +598,18 @@ impl DeviceTrait for Device {
 
 impl StreamTrait for Stream {
     fn play(&self) -> Result<(), PlayStreamError> {
```
roderickvd, line 685: Wouldn't same also apply to `pause()`?

src/host/aaudio/mod.rs
osoftware: I wasn't aware of this const, I'll try it out.

src/host/aaudio/mod.rs
osoftware: I might makes sense to use `wait_for_state_change` consistently in both but not using it on pause doesn't yield any serious issues and the method returns faster which might be preferable. IDK.

roderickvd: Unless somebody knows better, let's do it. I think it could lead to the same class of problems where the stream is still pausing (transitioning to another state) and drop is called.

roderickvd:
Thanks! Merging.

roderickvd:
@osoftware By the way and off-topic to this PR but since you're an AAudio user, can you confirm the dynamic tuning has been working well for you? (#1080 with AAudio defaulting to `BufferSize::Default`)

osoftware:
I've just tried it on https://github.com/osoftware/rush_audio and it works fine 👌

PR #1140, opened 2026-03-30
avoid building asio on Linux

wangjia184:
It is causing error to rust-analyzer

### DIFF
--- a/asio-sys/CHANGELOG.md
+++ b/asio-sys/CHANGELOG.md
@@ -23,18 +23,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `sampleRateDidChange` notifications
 - `CallbackId` renamed to `BufferCallbackId`
 - Public-facing `c_long` fields and return types replaced with `i32`
-- Public-facing `c_double` parameters and return types replaced with `f64`.
-- `Driver::latencies()` now returns `Latencies { input, output }`.
-- `Driver::buffersize_range()` now returns `BufferSizeRange { min, max }`.
-- `CallbackInfo::system_time` is now `u64` nanoseconds.
-- `AsioError::ASE_NoMemory` renamed to `AsioError::NoMemory`.
+- Public-facing `c_double` parameters and return types replaced with `f64`
+- `Driver::latencies()` now returns `Latencies { input, output }`
+- `Driver::buffersize_range()` now returns `BufferSizeRange { min, max }`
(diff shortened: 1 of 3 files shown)


roderickvd:
I'm not experiencing this issue on either my MacBook or Linux system. May you have accidentally set something like `rust-analyzer.cargo.features = "all"` in your editor settings, in doing so also enabling the "asio" feature?

wangjia184:
Hi @roderickvd 

I am developing in docker container.  The following line is in my Cargo.toml

```
#PipeWire media server backend. Requires libpipewire-0.3-dev (Debian/Ubuntu) or pipewire-devel (Fedora).
cpal = { path = "../cpal/", default-features = false, features = ["pipewire"] }
```

The project can be built in docker container successfully.
The issue is from rust analyzer, it always tries to built asio. I made several attempts, clean all cache,  it does not work.

The following part is from my `devcontainer.json` .  no `rust-analyzer.cargo.features = "all"`
```
"customizations": {
    "vscode": {
      "extensions": [
        "rust-lang.rust-analyzer",
        "tamasfe.even-better-toml",
        "kangping.protobuf",
        "rooveterinaryinc.roo-cline",
        "svelte.svelte-vscode"
      ],
      "settings": {
        "terminal.integrated.defaultProfile.linux": "bash",
        "files.watcherExclude": {
          "**/target/**": true,
          "**/.git/objects/**": true
        },
        "rust-analyzer.cargo.targetDir": true
      }
    }
  },
  ```

**Comment by roderickvd** [comment]
I see: it matters whether you're working on cpal itself (like I am) or pulling it in as a dependency.

So your proposed fix is the right approach. A few small things before we merge:

- Use `CARGO_CFG_TARGET_OS` instead of string-matching `TARGET`.
- hat also lets you drop the redundant `contains("msvc")` clause (every MSVC triple already contains "windows").

Then moving it before the `DOCS_RS` guard, something like this:

```rust
fn main() {
  match std::env::var("CARGO_CFG_TARGET_OS") {
      Ok(os) if os != "windows" => {
          let out_dir = PathBuf::from(env::var("OUT_DIR").expect("bad path"));
          create_stub_bindings(&out_dir);
          return;
      }
      _ => {}
  }

    if std::env::var("DOCS_RS").is_ok() {
        // ...
    }
```

wangjia184:
Thanks @roderickvd ,  code is updated now

roderickvd:
Thanks. I pushed some refactoring to your branch so I don't need to bother you with it. Will merge when the CI passes.

PR #1153, opened 2026-04-10
WASAPI: Prefer "friendly name" in device description

Sin-tel:
Fixes common issue where users have multiple devices named "Speakers". The friendly name is the same one as displayed in the OS output selection.

I opted to remove the extended line since it doesn't really add anything now.

### DIFF
--- a/src/host/wasapi/device.rs
+++ b/src/host/wasapi/device.rs
@@ -372,14 +372,14 @@ impl Device {
             );
 
             // Prefer FriendlyName for name (e.g., "Speakers (XYZ Audio Adapter)"), fall back to DeviceDesc
-            let name = friendly_name
-                .clone()
-                .or(device_desc.clone())
-                .ok_or_else(|| DeviceNameError::BackendSpecific {
-                    err: BackendSpecificError {
-                        description: "failed to retrieve device name".to_string(),
-                    },
-                })?;
+            let name =
+                friendly_name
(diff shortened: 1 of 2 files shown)


src/host/wasapi/device.rs
```diff
@@ -371,10 +371,10 @@ impl Device {
                 &PKEY_AUDIOENDPOINT_JACKSUBTYPE as *const _ as *const _,
             );
 
-            // Prefer DeviceDesc for name, fall back to FriendlyName
-            let name = device_desc
+            // Prefer FriendlyName for name (e.g., "Speakers (XYZ Audio Adapter)"), fall back to DeviceDesc
+            let name = friendly_name
                 .clone()
-                .or(friendly_name.clone())
+                .or(device_desc.clone())
```
roderickvd: Is that `clone` still necessary, with the "Add FriendlyName to extended if different from the name we used" block a bit down removed?

roderickvd (commented):
Yes, you're right. If you could check out this one tiny thing? I may be wrong when clippy hasn't triggered on it before.

src/host/wasapi/device.rs
Sin-tel: You're right. Apparently `clippy::redundant_clone` is allow by default.

roderickvd:
Merged, thanks!

Issue #889, opened 2024-06-09
Adding visionOS support


eugenehp:
## Previous work

- https://github.com/RustAudio/coreaudio-sys/pull/102
- https://github.com/RustAudio/coreaudio-rs/pull/119
- https://github.com/RustAudio/coreaudio-sys/pull/103
- https://github.com/RustAudio/coreaudio-rs/pull/117

## Current state

- this repo reuses `examples/ios-feedback` from `coreaudio-rs`
- `src/host/coreaudio` reuses both `ios` and `macos` logic from `coreaudio-rs`

## Current work

- @simlay is working on adding tvOS and watchOS support to `coreaudio-rs`
- I've a version of `cpal` that works on `visionOS`
- Released [avfaudio](https://crates.io/crates/avfaudio)

## Questions

1. What's a preferred way for reusing iOS example from `coreaudio-rs`?
2. AVFAudio and AVAudioSession is required to set the playback/record mode on iOS and visionOS devices. Do we abstract it under `coreaudio-rs` or upstream the changes and include them in `cpal` too?
3. CI pipeline has to be rebuilt, but not everything can be tested. @simlay suggested to merge all CI builds under `apple-tier3` job under one of coreaudio repos. Do we do the same here?
4. Anything I missed?

roderickvd:
I think this is done with #1155 ❤️

PR #1155, opened 2026-04-12
Add tvOS support for CoreAudio backend

csnewman:
### Changes
- Use `target_vendor = "apple"` for platform gates that apply to all Apple platforms.
- Use `not(target_os = "macos")` within the coreaudio module to select the iOS/tvOS code path.
- Bump mach2 to 0.6 (uses `core::ffi` instead of `::libc`, fixing tvOS builds).

### DIFF
--- a/.github/workflows/platforms.yml
+++ b/.github/workflows/platforms.yml
@@ -413,9 +413,6 @@ jobs:
     steps:
       - uses: actions/checkout@v5
 
-      - name: Install dependencies
-        run: brew install llvm
-
       - name: Install Rust nightly (for build-std)
         uses: dtolnay/rust-toolchain@nightly
         with:
--- a/.github/workflows/platforms.yml
+++ b/.github/workflows/platforms.yml
@@ -26,7 +26,7 @@ env:
   # MSRV varies by backend due to platform-specific dependencies
   MSRV_AAUDIO: "1.85"
   MSRV_ALSA: "1.82"
-  MSRV_COREAUDIO: "1.80"
+  MSRV_COREAUDIO: "1.85"
   MSRV_JACK: "1.82"
   MSRV_PIPEWIRE: "1.85"
   MSRV_PULSEAUDIO: "1.88"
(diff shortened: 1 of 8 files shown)


simlay:
I think it might be a very close set of changes to also add visionOS. It was added with coreaudio-rs 0.140.0 https://github.com/RustAudio/coreaudio-rs/pull/145 (I somehow forgot to write it in the release notes).

.github/workflows/platforms.yml
```diff
@@ -26,7 +26,7 @@ env:
   # MSRV varies by backend due to platform-specific dependencies
   MSRV_AAUDIO: "1.85"
   MSRV_ALSA: "1.82"
-  MSRV_COREAUDIO: "1.80"
+  MSRV_COREAUDIO: "1.85"
```
roderickvd, line 29: Please document this MSRV and `mach2` bump in the changelog.

.github/workflows/platforms.yml
```diff
@@ -405,6 +405,36 @@ jobs:
...
+  # tvOS (Tier 3 target, requires nightly + build-std)
+  tvos:
+    runs-on: macOS-latest
+    env:
+      TARGET: aarch64-apple-tvos
+    steps:
+      - uses: actions/checkout@v5
+
+      - name: Install dependencies
+        run: brew install llvm
```
roderickvd: I guess not but just to check that this isn't a carry over: `llvm` isn't installed by default on that build target?

src/host/coreaudio/mod.rs
```diff
@@ -12,12 +12,12 @@ use crate::{BuildStreamError, SupportedStreamConfigsError};
 
 use crate::{BackendSpecificError, SampleFormat, StreamConfig};
 
-#[cfg(target_os = "ios")]
+#[cfg(not(target_os = "macos"))]
```
roderickvd, line 17: How does this work on tvOS? Does that use the same API surface as iOS?

src/host/coreaudio/mod.rs
roderickvd: ...and tvOS! 👍

roderickvd (commented):
Nice refactoring. Just a few points and questions.

roderickvd:
> I think it might be a very close set of changes to also add visionOS.

Cool. That'd be #889 right?

roderickvd:
Thank you for taking care of those points and rebasing. Merging now!

PR #1161, opened 2026-04-13
fix(pipewire): Early pw_deinit() called cause Segfault

### LINKED ISSUE #1158: Segfault on pipewire 1.6.3
Caused by commit https://github.com/PipeWire/pipewire/commit/6bc07dfe0e18043aaf99f866b2c37f20e4a5e065
I will provide more information later.

### LINKED ISSUE #1160: Pipewire backend: Early pw_deinit() called cause Segfault
Reproduce:
```
use rodio::{DeviceSinkBuilder};

#[tokio::test]
async fn test_output() {
    let out = DeviceSinkBuilder::open_default_sink().unwrap();
    let out2 = DeviceSinkBuilder::open_default_sink().unwrap();
    drop(out);
    drop(out2);
}
```
with pipewire feature enabled.

Decodetalkers:
issue: https://github.com/RustAudio/cpal/issues/1160

The problem is that once we called deinit, we cannot use the safe pw::init() again, because in fact, the pw::init() is a OnceLock https://gitlab.freedesktop.org/pipewire/pipewire-rs/-/blob/main/pipewire/src/lib.rs?ref_type=heads#L180-184, so it will never be init again. The only way to init it the second time is to use the unsafe, since we have already managed the lifetime of pipewire in cpal, so maybe it is the right way here. Though I think we can do pw::init free when we want, and never deinit it, because it is quit useless, and because pipewire-rs is designed in this way, I think they also do not hope we to manage the lifetime ourself.

Anyway, this bug can be fixed with the unsafe init

### DIFF
--- a/src/host/pipewire/stream.rs
+++ b/src/host/pipewire/stream.rs
@@ -36,7 +36,7 @@ impl PwInitGuard {
     pub(crate) fn new() -> Self {
         let mut count = PW_INIT_COUNT.lock().unwrap_or_else(|e| e.into_inner());
         if *count == 0 {
-            pw::init();
+            unsafe { pw::sys::pw_init(std::ptr::null_mut(), std::ptr::null_mut()) }
         }
         *count += 1;
         Self
--- a/src/host/pipewire/stream.rs
+++ b/src/host/pipewire/stream.rs
@@ -36,6 +36,7 @@ impl PwInitGuard {
     pub(crate) fn new() -> Self {
         let mut count = PW_INIT_COUNT.lock().unwrap_or_else(|e| e.into_inner());
         if *count == 0 {
+            // pw::init() uses a OnceCell, preventing re-init after deinit.
             unsafe { pw::sys::pw_init(std::ptr::null_mut(), std::ptr::null_mut()) }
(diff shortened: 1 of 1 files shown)


revonateB0T:
Thanks! It works, fixes both https://github.com/RustAudio/cpal/issues/1158 & https://github.com/RustAudio/cpal/issues/1160

revonateB0T:
I still doubt we have sufficient reason to deinit pipewire. If users import rodio, they're going to use it, and frequent init/deinit is worse than just never deiniting at all.
@roderickvd 
If we decided to deinit it anyway, the overhead should be documented.

roderickvd:
Thanks for this fix. I added a short code comment on the `pw::init()` re-init behavior, and will merge right after.

> I still doubt we have sufficient reason to deinit pipewire. If users import rodio, they're going to use it, and frequent init/deinit is worse than just never deiniting at all.

We can't assume how cpal is used. For example, there may be applications like DAWs that allow selecting and switching backends at runtime. For these cases it's nice to be able to deinit and not hog resources, even if limited. Same pattern as ALSA.

PR #1167, opened 2026-04-18
feat(error): unified errors with precise kinds, messages, and no panics

roderickvd:
This is my last design goal for v0.18: a unified error type inspired by `std::io::Error`.

This new `cpal::Error` replaces `HostUnavailable`, `DevicesError`,`SupportedStreamConfigsError`, and all other per-operation error types in favor of a single `Error`/`ErrorKind` with an optional descriptive message.

While there, I took the opportunity to reduce the number of poisoned-lock panics, and making the timestamps on PipeWire infallible.

### DIFF
--- a/src/error.rs
+++ b/src/error.rs
@@ -147,6 +147,7 @@ impl From<ErrorKind> for Error {
 
 /// Extension trait for attaching a context message to a [`Result`] whose error converts into
 /// [`cpal::Error`].
+#[allow(dead_code)]
 pub(crate) trait ResultExt<T> {
     /// Converts the error via [`Into<cpal::Error>`] and prepends `msg`, yielding
     /// `"<msg>: <original error>"` as the message.
--- a/src/error.rs
+++ b/src/error.rs
@@ -144,3 +144,20 @@ impl From<ErrorKind> for Error {
         Self::new(kind)
     }
 }
+
+/// Extension trait for attaching a context message to a [`Result`] whose error converts into
+/// [`cpal::Error`].
+pub(crate) trait ResultExt<T> {
+    /// Converts the error via [`Into<cpal::Error>`] and prepends `msg`, yielding
+    /// `"<msg>: <original error>"` as the message.
+    fn context(self, msg: &'static str) -> Result<T, Error>;
+}
+
+impl<T, E: Into<Error>> ResultExt<T> for Result<T, E> {
+    fn context(self, msg: &'static str) -> Result<T, Error> {
+        self.map_err(|e| {
+            let e = e.into();
+            Error::with_message(e.kind(), format!("{msg}: {e}"))
+        })
+    }
+}
--- a/src/error.rs
+++ b/src/error.rs
@@ -1,3 +1,4 @@
+use std::borrow::Cow;
 use std::error::Error as StdError;
 use std::fmt::{Display, Formatter};
 
@@ -96,7 +97,7 @@ impl Display for ErrorKind {
 #[derive(Clone, Debug, PartialEq, Eq, Hash)]
 pub struct Error {
     kind: ErrorKind,
-    message: Option<String>,
+    message: Option<Cow<'static, str>>,
 }
 
 impl Error {
@@ -109,7 +110,7 @@ impl Error {
     }
 
     /// Create a new error with the given kind and a human-readable message.
-    pub fn with_message(kind: ErrorKind, message: impl Into<String>) -> Self {
+    pub fn with_message(kind: ErrorKind, message: impl Into<Cow<'static, str>>) -> Self {
         Self {
             kind,
             message: Some(message.into()),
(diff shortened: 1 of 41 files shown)


tachibanayui:
While we working on revamping Errors, should we consider capture [backtrace](https://doc.rust-lang.org/std/backtrace/index.html) and/or retriving [source](https://doc.rust-lang.org/std/error/trait.Error.html#error-source) (inner error)? It makes debugging easier when we have the original error from host backends instead of a generic  `Other` error

roderickvd:
In this implementation I tried to address that use case by capturing `e.to_string()` from every backend error into the `message` field, which covers the "what was the original error?" question without capturing its source.

If in addition to that we'd add `source()`, then that'd have some downsides: it would require storing `Box<dyn Error>` or `Arc<dyn Error + Send + Sync>`, breaking the `Clone + PartialEq + Eq + Hash` derives. The platform-specific inner types also aren't part of cpal's public API, so downcasting to them would create cross-platform coupling.

I feel that with the `message` field, those downsides aren't worth it.

tachibanayui:
I see, in that case I think we should make sure that host implementations don't forget to attach "context" when propogating error up, which I think we could do with some `From<T>` implementations or an `.context()` method similar to `anyhow`. Also this maybe insignificant but maybe the `message` field could use `Cow<'static, str>` to save a few allocs on static error messages (I don't know other hosts very well but I see wasapi could use a few)

roderickvd:
> I see, in that case I think we should make sure that host implementations don't forget to attach "context" when propogating error up, which I think we could do with some `From<T>` implementations or an `.context()` method similar to `anyhow`.

I think we have that, right? There _are_ `From<T>` implementations, and there _is_ a `with_message` to add that context.

> Also this maybe insignificant but maybe the message field could use Cow<'static, str> to save a few allocs on static error messages (I don't know other hosts very well but I see wasapi could use a few)

Good idea, I'll look into it.

tachibanayui:
I meant something like this: 
```rs
// I type this directly in github so treat this as more like pseudo-code
impl Error {
	pub fn context(self, msg: &str) -> Self {
	    self.message = format!("{}\nCaused by: {}", msg, self);
	    self
	}
}
```

It is similar to `host::wasapi::windows_err_to_cpal_err_message` but generalized to support adding context.

roderickvd:
I'm not sure the "Caused by:" format or similar is the right fit for cpal. As a general-purpose library, formatting error messages should be the downstream app's choice. If the concern is that backends might lose context when propagating errors, I think they already handle that today via `with_message`.

tachibanayui:
> If the concern is that backends might lose context when propagating errors, I think they already handle that today via `with_message.`

Well the function I proposed is just a helper for propogating errors so when we propogate errors we dont have to do: 
```rs
pub(crate) fn build_output_stream_raw_inner() {
    // -- snip --
    let max_frames_in_buffer = audio_client.GetBufferSize().map_err(|e| {
	    windows_err_to_cpal_err_message::<BuildStreamError>(
	        e,
	        "failed to obtain buffer size: ",
	    )
    })?;

    // We can shorten it into
    let max_frames_in_buffer = audio_client.GetBufferSize().context("failed to obtain buffer size")?;
}
```
Although to have the above syntax work we need to redefine the helper function to an extension trait
```rs
pub trait ErrorExt{...}
impl<T, E: Display> ErrorExt for Result<T, E> {
    fn context(self, msg: &str) -> Result<T, Error> {...}
}
```
If a specific host want to process the inner error, they have option to format the string manually and call `with_message`.
I saw `host::wasapi::windows_err_to_cpal_err_message` get used quite frequently so it might be worth it to have the function available crate-wide

roderickvd:
Ah, as _convenience_ method. Yeah, we can do that.

roderickvd:
I'm going to merge this so it gets some exposure. Please open an issue for any further points.

src/error.rs
```diff
@@ -144,3 +144,20 @@ impl From<ErrorKind> for Error {
         Self::new(kind)
     }
 }
+
+/// Extension trait for attaching a context message to a [`Result`] whose error converts into
+/// [`cpal::Error`].
+pub(crate) trait ResultExt<T> {
+    /// Converts the error via [`Into<cpal::Error>`] and prepends `msg`, yielding
+    /// `"<msg>: <original error>"` as the message.
+    fn context(self, msg: &'static str) -> Result<T, Error>;
```
tachibanayui, line 153: Is there a particular reason it need to be `'static`?

roderickvd: We didn't need more now but we can make it `impl Display`.

tachibanayui: I think we could remove the lifetime here and it would still compile. We don't need the static lifetime for the extension to work and it restrict the user to only string literals
```suggestion
    fn context(self, msg: &str) -> Result<T, Error>;
```
> we can make it impl Display.

This could work too

roderickvd: I'll change it to `impl Display` on master. Thanks for thinking along.

PR #1169, opened 2026-04-18
Add `windows-core` crate to allow usage of #[implement] macro

tachibanayui:
This pull request add `windows-core` crate to allow usages of `#[implment]` macro for implementing COM interfaces.
I updated the CI workflows to also locks `windows-core` to test compatibility. I also added `activate_audio_interface_sync` function which is currently unused so that CI can fail on version mismatch and future features can use it to activate virtual devices (unblock #1027).


roderickvd:
Thanks for creating this PR!

The `windows-core` dependency addition seems the right idea. I am not sure about `activate_audio_interface_sync` though. It seems to conflict with #1027's version, what'd be your proposal?

My proposal would be to merge this without `activate_audio_interface_sync` and then update #1027. Let me know what you think.

tachibanayui:
I added the function to make sure CI fail if for some reason the `windows` and `windows-core` version mismatch so it is technically not required to have it in the PR. If it is not needed, please confirm so I can remove it.
sniped output from [a failing CI build](https://github.com/tachibanayui/cpal/actions/runs/24605356828/job/71950596774?pr=2)
```rs
error[E0407]: method `from_inner_ref` is not a member of trait `windows_core::IUnknownImpl`
   --> src\host\wasapi\device.rs:300:5
    |
300 |     #[windows::core::implement(Audio::IActivateAudioInterfaceCompletionHandler)]
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ not a member of trait `windows_core::IUnknownImpl`
    |
    = note: this error originates in the attribute macro `windows::core::implement` (in Nightly builds, run with -Z macro-backtrace for more info)

error[E0438]: const `INNER_OFFSET_IN_POINTERS` is not a member of trait `windows_core::IUnknownImpl`
   --> src\host\wasapi\device.rs:300:5
    |
300 |     #[windows::core::implement(Audio::IActivateAudioInterfaceCompletionHandler)]
    |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ not a member of trait `windows_core::IUnknownImpl`
    |
    = note: this error originates in the attribute macro `windows::core::implement` (in Nightly builds, run with -Z macro-backtrace for more info)
```

About #1027, it is implemented only to support intializing default virtual device, as such the `params: Option<*const StructuredStorage::PROPVARIANT>` cannot be set by callers. To avoid having to duplicate this function (eg: when we implement process capture, the params is used to store the processId to capture) I think the function should only abstract away the async mechanism.

roderickvd:
> I added the function to make sure CI fail if for some reason the `windows` and `windows-core` version mismatch so it is technically not required to have it in the PR. If it is not needed, please confirm so I can remove it.

If we get #1027 landed then that would ensure that in the CI too, right? Let's take it out of this one, then revive #1027 and continue the discussion there with @philpax.

roderickvd:
Thanks. If you would get the CI to pass and let me know, we should be good to merge.

tachibanayui:
ah sorry i forgot to run clippy before commiting. CI should be passing now

philpax:
Unfortunately, I'm not sure this resolves the two issues I had when I tried this :disappointed: 

1. There's no guarantee that Cargo will resolve the same `windows-core` and `windows` versions for downstream consumers; you've had to do the same hack I did, which was to force-downgrade it in CI, but that implies that it won't happen naturally for consumers
2. `implement` was feature-gated in versions of the crate prior to 0.61, so any attempt to use it will fail with versions that fall under that; conversely, specifying the feature will then fail for >=0.61.

Would love for there to be a way out of this, but I was banging my head against this for the longest time and couldn't find anything obvious.

tachibanayui:
> There's no guarantee that Cargo will resolve the same windows-core and windows versions for downstream consumers; you've had to do the same hack I did, which was to force-downgrade it in CI, but that implies that it won't happen naturally for consumers

For the consumer to force cpal to use the same `windows` as `cpal`'s, they already have to use `cargo update` to unify manually.  This is the output if we only have `windows` (before the PR) on a consumer crate with dependency windows = "0.59.0" and cpal = { path = ".." }.

```
> cargo tree
test-user-crate v0.1.0 (E:\Projects\TestProjects\Rust\cpal\test-user-crate)
├── cpal v0.18.0 (E:\Projects\TestProjects\Rust\cpal)
│   ├── dasp_sample v0.11.0
│   └── windows v0.62.2
│       ├── windows-collections v0.3.2
│       │   └── windows-core v0.62.2
│       │       ├── windows-implement v0.60.2 (proc-macro)
│       │       │   ├── proc-macro2 v1.0.106
│       │       │   │   └── unicode-ident v1.0.24
│       │       │   ├── quote v1.0.45
│       │       │   │   └── proc-macro2 v1.0.106 (*)
│       │       │   └── syn v2.0.117
│       │       │       ├── proc-macro2 v1.0.106 (*)
│       │       │       ├── quote v1.0.45 (*)
│       │       │       └── unicode-ident v1.0.24
│       │       ├── windows-interface v0.59.3 (proc-macro)
│       │       │   ├── proc-macro2 v1.0.106 (*)
│       │       │   ├── quote v1.0.45 (*)
│       │       │   └── syn v2.0.117 (*)
│       │       ├── windows-link v0.2.1
│       │       ├── windows-result v0.4.1
│       │       │   └── windows-link v0.2.1
│       │       └── windows-strings v0.5.1
│       │           └── windows-link v0.2.1
│       ├── windows-core v0.62.2 (*)
│       ├── windows-future v0.3.2
│       │   ├── windows-core v0.62.2 (*)
│       │   ├── windows-link v0.2.1
│       │   └── windows-threading v0.2.1
│       │       └── windows-link v0.2.1
│       └── windows-numerics v0.3.1
│           ├── windows-core v0.62.2 (*)
│           └── windows-link v0.2.1
└── windows v0.59.0
    ├── windows-core v0.59.0
    │   ├── windows-implement v0.59.0 (proc-macro)
    │   │   ├── proc-macro2 v1.0.106 (*)
    │   │   ├── quote v1.0.45 (*)
    │   │   └── syn v2.0.117 (*)
    │   ├── windows-interface v0.59.3 (proc-macro) (*)
    │   ├── windows-result v0.3.4
    │   │   └── windows-link v0.1.3
    │   ├── windows-strings v0.3.1
    │   │   └── windows-link v0.1.3
    │   └── windows-targets v0.53.5
    │       └── windows_x86_64_msvc v0.53.1
    └── windows-targets v0.53.5 (*)
```

`windows` appear with 2 version regardless of `windows-core` presence. 
Also I don't see a problem with it appears twice in the tree, cpal doesn't expose anything about it's backend native api so we don't have worry about passing value between 2 versions. And as for the increased compile time, you can use the `cargo update` to unify dependencies.  It's sad that cargo will not unify dependencies when it is breaking semver senmantics, eventhough we specifically tell it that it is supported.

> implement was feature-gated in versions of the crate prior to 0.61, so any attempt to use it will fail with versions that fall under that; conversely, specifying the feature will then fail for >=0.61.

Hmm, I have the `activate_audio_interface_sync` before and the CI seems to pass on both 0.59, 0.60 and 0.61.3 https://github.com/RustAudio/cpal/actions/runs/24606810025/job/71954321337?pr=1169. Maybe I'm missing something?

roderickvd:
Superseded by #1183 that does a similar trick in the CI. Thanks for this pointer!

PR #1171, opened 2026-04-18
fix(pipewire): verify nodes are available

### LINKED ISSUE #1170: CPAL should skip the PipeWire backend when PulseAudio and PipeWire are both running on the system and conflicting
See: https://github.com/marin-m/SongRec/issues/265

The PipeWire backend will produce a "no node available" error when this issue happens with the corresponding defective setup

I think that for better user experience we should switch to trying the PulseAudio or ALSA backend depending on what is enabled, rather than displaying this error

roderickvd:
Don't add default devices when there aren't any nodes available.

Fixes #1170

### DIFF
--- a/src/host/pipewire/device.rs
+++ b/src/host/pipewire/device.rs
@@ -780,6 +780,12 @@ pub fn init_devices() -> Option<Vec<Device>> {
 
     mainloop.run();
 
+    // If PipeWire connected but discovered no real audio nodes, it cannot route any streams. Treat
+    // this as unavailable so the caller can fall back to PulseAudio or ALSA.
+    if discovered.borrow().is_empty() {
+        return None;
+    }
+
     let settings = settings.take();
 
     // Build the three synthetic default devices and apply global clock settings to them.


marin-m:
I don't have what to reproduce the exact defective setup easily but you requested my review so lgtm, some trivial questions
* Are you sure that the "no node available" error message may not be returned/thrown by the `pipewire` crate earlier?
* Are all the resources correctly freed in case of failure?

Regards,

marin-m:
In any case I can confirm that PipeWire support doesn't break on my setup

roderickvd:
> I don't have what to reproduce the exact defective setup easily but you requested my review so lgtm, some trivial questions

Yeah I saw that it was one of your users who ran into it. I'll merge this and consider it fixed. Let me know if you get any feedback from him/her.

> * Are you sure that the "no node available" error message may not be returned/thrown by the `pipewire` crate earlier?

If I'm understanding your question correctly: yes, that error message came from PipeWire itself, not from cpal. But is triggered when cpal tries to open a default device that PipeWire cannot route to any audio node.

> * Are all the resources correctly freed in case of failure?

PipeWire is uninitialized by cpal when the caller drops its last PipeWire host.

Issue #1175, opened 2026-04-21
[macOS] unplugging headphones causes audio to stop


hystericca:
# summary
pretty much what it says in the title

https://github.com/user-attachments/assets/36357ba2-00ff-43a2-8627-066236fde026

*when the audio stops is when I unplug my headphones*

https://github.com/user-attachments/assets/270f293e-fd41-4529-a032-d49474f497c2

*same thing here*

143mailliw:
The application that the second video is of is mine. It seems like the issue is that for what ever reason on macOS the device disconnection code doesn't run if the device that the cpal stream was created with is disconnected. (eg. if I plug in my headphones after starting the application disconnecting and reconnecting works fine an unlimited number of times, but if I plug my headphones in *before* starting the application *and then disconnect them* audio stops playing entirely for no discernible reason)

roderickvd:
From code review I see that the code is currently broken in both directions: default devices don't get automatically rerouted upon disconnection, and specifically selected devices _do_ get automatically routed.

Probably we would like the inverse:
- when we opened the default device, and that gets disconnected, reroute to the new default device
- when we opened a specific device, and that gets disconnected, emit a `DeviceNotAvailable` error

And when there's no more default devices to route to, then also emit a `DeviceNotAvailable` error.

sandreas:
@roderickvd 

Thanks for the update. Please keep in mind it might be possible that disconnecting a device does not always mean falling back to the default one is the right choice. 

Example:
I'm working on a portable audio player with a board, that has a default audio chip, that I'm not even using at all (not connected). So disconnecting my USB-C 3.5mm Adapter would mean that there is no real audio output connected. I'd be happy, if I could handle disconnects by running explicit code like showing a dialog in my UI (e.g. Audio disconnected, Playback has been paused).


So I'd prefer a default behavior AND the possibility to hook into the event of disconnecting a device and deciding what should happen.

roderickvd:
Yes, I understand. Or how about emitting a new error kind like `DeviceChanged` over the existing error fallback?

In the meantime, when we fix the code to the “sane defaults” as above, you could work around your case by choosing that USB adapter specifically.

sandreas:
> Yes, I understand. Or how about emitting a new error kind like DeviceChanged over the existing error fallback?

Maybe... like a quickfix? However, disconnecting a device does not *feel* like an error would represent this correctly, since this is expected behavior, but for a first solution I think this might be enough.

[...]

Issue #1012, opened 2025-09-21
[Question] Handling device disconnects


JamesMoreau:
So I have a music player application and I would like for the user to be able to select the audio output device.

The user could switch devices or unplug it in the middle of a track. So, I need to be able to rebuild the stream at any point.

However, I noticed, at least on macOS, that upon disconnecting a device (in this case my headphones), the audio gets automatically rerouted to another device all on its own. This is nice but now I no longer know what the current device actually is. Which means that my application state, which display the current device, is now out of sync.

Is there a way to turn off this feature, or is it something the OS does? Otherwise, is it possible to listen for a particular stream event? Or maybe just even be able to query the current output device given a particular stream?

Thanks.

roderickvd:
It's currently hardcoded: https://github.com/RustAudio/cpal/blob/1c9c642bb31907df03e4d62063bc344fde539591/src/host/coreaudio/macos/mod.rs#L673

Could be knob to add in #1010.

sandreas:
@roderickvd 
I would also appreciate this due to the nature of USB-C Audio adapters that could be plugged out (accidentally or intentionally) in the middle of audio playback / Streaming.

I think a "DeviceChange" enum containing the device that changed and the nature of the change could be enough for a first implementation
- DeviceChange::Disconnect(cpal::Device)
- DeviceChange::Connect(cpal::Device)

along with `cpal::on_device_change(callback)`...

roderickvd:
I see now that this is a bug. Let's continue over at #1175.

roderickvd:
Please give PR #1178 a shot and report back there.

PR #1178, opened 2026-04-26
fix(coreaudio): reroute when default output changes

roderickvd:
For streams opened on the default output device, when that default device changes, reroute to it and emit `DeviceChanged` or `DeviceNotAvailable` if no default device is available anymore.

For streams opened on specific devices, pin to that device even if it was the default.

Fixes:
- #1012
- #1175


roderickvd:
On a merge streak, so merging this now to let it get some more exposure - let me know how it goes!

PR #1183, opened 2026-04-27
feat(wasapi): auto-rerouting of default device streams

roderickvd:
This takes all the work by @philpax from #1027 and @sidit77 before him with #754 to support seamless rerouting of default device streams. When it does, it fires an `ErrorKind::DeviceChanged` event on the error callback.

Before, we were stuck in dependency hell. `windows` v0.59 and v0.60 are over a year old now, so for cpal v0.18 I think it's time to move on.

As I'm not a Windows user myself, this PR was prepared with Claude Code in hand. Real-world verification necessary!

Supersedes:
- #754
- #1027
- #1169

### DIFF
--- a/.github/workflows/platforms.yml
+++ b/.github/workflows/platforms.yml
@@ -554,6 +554,7 @@ jobs:
         run: |
           cargo generate-lockfile
           cargo update -p windows --precise ${{ matrix.version }}
+          cargo update -p windows-core --precise ${{ matrix.version }}
           echo "Locked windows crate version:"
           cargo tree | grep "windows v" || echo "Windows crate not found in dependency tree"
           echo "Cargo.lock entry:"
--- a/.github/workflows/platforms.yml
+++ b/.github/workflows/platforms.yml
@@ -554,7 +554,14 @@ jobs:
         run: |
           cargo generate-lockfile
           cargo update -p windows --precise ${{ matrix.version }}
-          cargo update -p windows-core --precise ${{ matrix.version }}
+          # After pinning windows there may be two windows-core versions in the lockfile
+          # (the direct dep stays at the latest allowed, the transitive dep is what windows
+          # needs). Find the stale one dynamically and replace it with the needed version.
+          NEEDED=$(awk -F'"' '/^name = "windows-core"$/{f=1} f && /^version = /{print $2; f=0}' Cargo.lock \
+            | sort -V | head -1)
+          STALE=$(awk -F'"' '/^name = "windows-core"$/{f=1} f && /^version = /{print $2; f=0}' Cargo.lock \
+            | sort -V | tail -1)
+          [ "$NEEDED" != "$STALE" ] && cargo update -p "windows-core@$STALE" --precise "$NEEDED"
           echo "Locked windows crate version:"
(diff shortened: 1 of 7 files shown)


philpax:
Happy to see the tiebreak in favour of updating `windows`! Time solves all problems :slightly_smiling_face:

 I'd appreciate a `Co-Authored-By` on the commit if possible (and sidit might, too).

Anyhow, I notice that this PR adds a notification listener for device switching, and wires it up for the stream; if you're doing this, you can skip the entire `ActivateAudioInterfaceAsync`  dance and implement audio switching through the notification mechanism, as [Chromium](https://chromium.googlesource.com/chromium/src/media/+/refs/heads/main/audio/win/audio_device_listener_win.cc) [does](https://chromium.googlesource.com/chromium/src/media/+/refs/heads/main/audio/win/audio_manager_win.cc) (there's more to the picture here, but this is what I vaguely remember from my own investigation).

This would also mean that it would work on pre-Windows-8 installs. The reason I didn't do this is because I didn't feel comfortable doing the required refactoring for the stream (iirc), but you may be more empowered to do so (or, at least, Claude might :laughing:)

roderickvd:
> I'd appreciate a `Co-Authored-By` on the commit if possible (and sidit might, too).

Absolutely! Done. Should have done that in the first place.

> Anyhow, I notice that this PR adds a notification listener for device switching, and wires it up for the stream; if you're doing this, you can skip the entire `ActivateAudioInterfaceAsync` dance and implement audio switching through the notification mechanism, as [Chromium](https://chromium.googlesource.com/chromium/src/media/+/refs/heads/main/audio/win/audio_device_listener_win.cc) [does](https://chromium.googlesource.com/chromium/src/media/+/refs/heads/main/audio/win/audio_manager_win.cc) (there's more to the picture here, but this is what I vaguely remember from my own investigation).
> 
> This would also mean that it would work on pre-Windows-8 installs. The reason I didn't do this is because I didn't feel comfortable doing the required refactoring for the stream (iirc), but you may be more empowered to do so (or, at least, Claude might 😆)

Appreciate you reading along, and for those direct pointers to Chromium.

If I understand well, that's a different mechanism indeed: it doesn't let the OS reroute, but does so manually in the run loop by tearing down and re-creating a stream. I am doubting whether that's what we want:
- it differs from what the other hosts do
- it could cause a slight gap between teardown and reinitialization
- it would not transparently deal with format changes

With Windows 8 from 2012 and Vista before it from 2007, I'd be OK with not supporting Vista and older.

roderickvd:
Without any further comments can I assume this PR is working well? If so, I'm going to merge.

philpax:
Unfortunately, I am also not using Windows these days, so I can't test - might be worth testing in a VM? Otherwise, I think it's conceptually fine.

roderickvd:
Thanks for the response. @tachibanayui how I does this work for you?

tachibanayui:
I tested on Windows 11 and default audio routing seems to work well, default audio device changed error event also fires when I change the default output device using the feedback example. Unfortunately I don't have 2 mics to test the input side but they use the same mechanism for notification so it will likely work.

roderickvd:
Thanks. Merging now.

PR #1184, opened 2026-04-28
fix(coreaudio): crashes from early initialization

### LINKED ISSUE #1182: Crashes on macOS for some audio drivers
## Problem
I'm experiencing some stochastic crashes when enumerating my audio devices and getting their names via `.description()`. Starting an output stream also crashes (I'd say 35% of the time). Apparently, some audio driver is accessing some dead mutex, so I assumed it would be a bug within the audio driver. Nonetheless, going back to v0.16 of CPAL made the crashes almost disappear (it happened only very few times), so I figure something else must be going on.

Sadly, these crashes are making the 0.17.x completely unusable in my case.

## Environment
CPAL version 0.17.3.
Mac Mini M2 with MacOS Tahoe 26.4.1.
My main audio interface in this thread is this [professional audio device](https://www.merging.com/products/interfaces/merging+anubis).

This issue appeared much more often in Release Mode than in Debug Mode.

## Analysis
After a longer debugging session, I found something that seems to be problematic for my specific driver, **but also just looks a bit wrong in general**: Audio Units are always being instantiated and initialized _directly afterwards_. This is triggered inside `coreaudio-rs` every time. After getting an already initialized audio unit, CPAL is currently setting different things regarding the stream like sample format and such (or just disposing the instance if no stream is to be started). This order does not comply with the standard usage of CoreAudio, because initialization makes more sense only AFTER setting all stream parameters is done (allocations within the audio device happen at initialization and they depend on settings). I believe, this is the root of the problem.

I rewrote that (while also using a dev branch in `coreaudio-rs`), and all crashes are gone (validated with a significant amount of testing). Additionally, everything else also works just as expected, confirming the idea that audio unit initialization is not necessary for every instance and every usage.

PS: I'm actually not quite sure why the crashes do not happen as often with v0.16, I didn't look much into the differences between the function after I solved it by the strategy already discussed.

## Proposal
This should be changed to defer the initialization only before really starting the audio stream AND avoid it altogether if unnecessary (for example when executing `.description()` because querying device info does not even require such initialization). This requires a PR on the side of `coreaudio-rs` in order to change the API. It would also be a much cleaner approach in my opinion.

I'm aware that it's probably a hardware bug, but I think it's still an improvement for the crate.

### Pre-requirement
As already mentioned, this requires new functionality from `coreaudio-rs`. This is the [PR](https://github.com/RustAudio/coreaudio-rs/pull/152). Once that's merged, I'll gladly proceed with a PR on this side, while also updating to the latest version of `coreaudio-rs`, of course.

roderickvd:
Configuring the stream after the `AudioUnit` is initialized breaks the CoreAudio contract and caused crashes on certain drivers.

Fixes #1182

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -135,6 +135,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **CoreAudio**: Fix undefined behaviour and silent failure in loopback device creation.
 - **CoreAudio**: Poisoned stream locks now return `ErrorKind::StreamInvalidated` instead of 
   panicking.
+- **CoreAudio**: Fix crashes on certain drivers due to early initialization.
 - **JACK**: Fix input capture timestamp using callback execution time instead of cycle start.
 - **JACK**: Poisoned error callback mutex no longer silently drops subsequent error notifications.
 - **PulseAudio**: Poisoned locks now exit the thread gracefully instead of panicking.
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -115,7 +115,7 @@ pipewire = { version = "0.9", optional = true, features = ["v0_3_53"] }
 
 [target.'cfg(target_vendor = "apple")'.dependencies]
 mach2 = "0.6"
-coreaudio-rs = { version = "0.14", default-features = false, features = [
+coreaudio-rs = { version = "0.14.2", default-features = false, features = [
     "core_audio",
     "audio_toolbox",
 ] }
--- a/src/host/coreaudio/ios/mod.rs
+++ b/src/host/coreaudio/ios/mod.rs
@@ -425,11 +425,10 @@ fn setup_stream_audio_unit(
     }
 
     let mut audio_unit = create_audio_unit()?;
+    audio_unit.uninitialize()?;
 
     if is_input {
-        audio_unit.uninitialize()?;
         configure_for_recording(&mut audio_unit)?;
-        audio_unit.initialize()?;
     }
 
     // Set the stream format in interleaved mode
@@ -443,6 +442,8 @@ fn setup_stream_audio_unit(
 
     let asbd = asbd_from_config(config, sample_format);
     audio_unit.set_property(kAudioUnitProperty_StreamFormat, scope, element, Some(&asbd))?;
+
+    audio_unit.initialize()?;
 
     Ok(audio_unit)
 }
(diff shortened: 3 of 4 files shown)


nico-franco-gomez:
I think the changes here would already solve the problem when using the `.description()` method. Nonetheless, I also experienced some occasional crashes while initializing the audio unit on creating a single output stream, meaning that probably some stale state was still in place when the audio unit was initialized in `coreaudio-rs`. The [PR](https://github.com/RustAudio/coreaudio-rs/pull/152) has already been merged and I expect the publishing of the [new version](https://github.com/RustAudio/coreaudio-rs/pull/153) very soon. Therefore, I'd advocate for waiting for that to be available and then avoid the `uninitialize` steps, because unwanted allocations on the audio devices are still happening in this case. That's the cleaner approach imo, but of course, the decision lies on your side.

roderickvd:
Sure, if it’ll be released soon then we can wait for that. I’d like to include it in cpal v0.18. 

Besides the coreaudio-rs PR, to verify my understanding, does my PR fix it for you completely?

simlay:
> he https://github.com/RustAudio/coreaudio-rs/pull/152 has already been merged and I expect the publishing of the https://github.com/RustAudio/coreaudio-rs/pull/153 very soon.

Sorry. I got distracted earlier (I'm a new father and very tired) and forgot to push the merge button. [coreaudio-rs 0.14.2 is published.](https://crates.io/crates/coreaudio-rs/0.14.2).

nico-franco-gomez:
I tested the PR right now, and I still could reproduce a crash due to the early initialization inside coreaudio-rs when starting an output stream. Since that's not yet avoided here, it makes sense that it still happens.

But given the new coreaudio-rs version is already available, it would be straight-forward to integrate that in here, right?

Additionally, there's a function called `audio_unit_from_device` in macos/device.rs. There's the exact same functionality in the public API of coreaudio-rs. I would also suggest to get rid of that function altogether and just import the other one...

roderickvd:
Thanks. I’ll update soon.

roderickvd:
Hope this does it - feedback appreciated as I can't reproduce it locally.

roderickvd:
I'm going to merge this and close https://github.com/RustAudio/cpal/issues/1182https://github.com/RustAudio/cpal/issues/1182. Please re-open that ticket if there are remaining issues.

nico-franco-gomez:
Ok, I'll have access to my test setup in a couple of days, so I'll notify if I manage to reproduce the crashes again.

PR #1189, opened 2026-05-03
fix(pulseaudio): release server-side stream on Stream::drop

### LINKED ISSUE #1188: Each pulseaudio playback leaks a PA stream
`cpal::Stream::new_playback` spawns a thread that runs `block_on(stream.play_all())`. 
Internally `play_all` is `source_eof().await?; drain().await?;` : it waits for the cpal callback to return 0 bytes. But the `CallbackWrapper` in the pulseaudio crate (0.3.1) always returns `buf.len()`, never 0. 

So that thread blocks forever, holds an `Arc<InnerPlaybackStream>` clone, and `InnerPlaybackStream::drop` (which sends `DeletePlaybackStream`) never fires. 

**Result: every playback = one leaked PA stream plus one stuck OS thread.**

`Stream::drop` only calls `handle.cancel()`, which unblocks the latency-monitor thread, not the `play_all` thread.

Image

knz:
The PulseAudio backend spawns a driver thread that calls `pulseaudio::PlaybackStream::play_all`, which awaits source EOF before draining and exiting. EOF for a `PlaybackSource` is signalled by `poll_read` returning `0`, but the data-callback wrapper used by `new_playback` always reports the full buffer length to satisfy cpal's no-short-write contract. The driver thread therefore parks in `source_eof` for the lifetime of the process.

Because that thread holds a `PlaybackStream` clone (an `Arc<InnerPlaybackStream>`), the inner Arc count never drops to zero, `InnerPlaybackStream::drop` never fires, and the `DeletePlaybackStream` command it would otherwise queue is never sent. Each `Stream` created and dropped by the user thus leaks one server-side stream and one OS thread; long-running clients accumulate `pactl list short sink-inputs` entries until PulseAudio runs out of channels.

Queue a `DeletePlaybackStream` from `Stream::drop` so the reactor removes the stream state, drops the source's `eof_tx` channel, and wakes the driver thread with a cancellation error. `now_or_never` mirrors `InnerPlaybackStream::drop` and avoids blocking in `Drop`.

Fixes #1188.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -173,6 +173,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **WASAPI**: Fix callbacks firing before `build_*_stream` returns the `Stream` handle.
 - **WebAudio**: Fix duplicated callbacks on repeated `play()` calls.
 - **WebAudio**: Report errors through the callback instead of panicking.
+- **PulseAudio**: Poisoned locks now exit the thread gracefully instead of panicking.
 
 ## [0.17.3] - 2026-02-18
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -173,9 +173,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **WASAPI**: Fix callbacks firing before `build_*_stream` returns the `Stream` handle.
 - **WebAudio**: Fix duplicated callbacks on repeated `play()` calls.
 - **WebAudio**: Report errors through the callback instead of panicking.
-- **PulseAudio**: Poisoned locks now exit the thread gracefully instead of panicking.
-- **PulseAudio**: Fix server-side stream leak when a `Stream` is dropped.
-- **PulseAudio**: Remove spurious "PulseAudio client disconnected" warnings during stream cleanup.
 
 ## [0.17.3] - 2026-02-18
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -174,6 +174,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **WebAudio**: Fix duplicated callbacks on repeated `play()` calls.
 - **WebAudio**: Report errors through the callback instead of panicking.
 - **PulseAudio**: Poisoned locks now exit the thread gracefully instead of panicking.
+- **PulseAudio**: Fix server-side stream leak when a `Stream` is dropped.
 
 ## [0.17.3] - 2026-02-18
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -175,6 +175,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **WebAudio**: Report errors through the callback instead of panicking.
 - **PulseAudio**: Poisoned locks now exit the thread gracefully instead of panicking.
 - **PulseAudio**: Fix server-side stream leak when a `Stream` is dropped.
+- **PulseAudio**: Remove spurious "PulseAudio client disconnected" warnings during stream cleanup.
 
 ## [0.17.3] - 2026-02-18
 
--- a/src/host/pulseaudio/stream.rs
+++ b/src/host/pulseaudio/stream.rs
@@ -67,7 +67,7 @@ impl Drop for Stream {
                 // Help the play_all driver thread terminate by
                 // queueing a delete, which causes the reactor to drop
                 // the source's eof_tx. We need to do this because
-                // pool_read always reports a non-empty buffer.
+                // poll_read always reports a non-empty buffer.
                 let _ = stream.clone().delete().now_or_never();
             }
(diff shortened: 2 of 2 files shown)


knz:
NB: cleanup is asynchronous (the delete is queued, not awaited), so a `Drop` returning does not guarantee the server-side stream is gone yet. That mirrors `InnerPlaybackStream::drop` and avoids `block_on` in destructors.  If you prefer strict synchronous cleanup, we can use `block_on` instead.

roderickvd:
Thank you for the analysis and follow-up fix! This seems the right approach.

One concern: the driver thread is never joined. `drop` returns before the thread has processed the cancellation, so `DeletePlaybackStream` may not reach the server if the process exits shortly after.

Fixing this requires storing the worker thread's `JoinHandle` and joining it with self-join detection to prevent deadlocks if a user's error callback drops the `Stream`. In #1187 I'm fixing a similar issue for PipeWire. Within that large PR the key lines are these:

```rust
if handle.thread().id() != std::thread::current().id() {
    let _ = handle.join();
}
```

Would you take that along with this PR?

knz:
Done; added a commit, see if you like it.

knz:
hm I sense there's something wrong. let me investigate further.

knz:
Okay this patch is actually fine!

(while testing it I ran into the unrelated issue #1190.  I've been able to work around that issue in my own code, but it may be worth investigating further.)

knz:
There's a new failure in CI, but it appears unrelated to my change.

src/host/pulseaudio/stream.rs
```diff
@@ -51,21 +52,46 @@ enum StreamInner {
     Record(pulseaudio::RecordStream, Instant, LatencyHandle),
 }
 
-pub struct Stream(StreamInner);
+pub struct Stream {
+    inner: StreamInner,
+    workers: Vec<std::thread::JoinHandle<()>>,
```
roderickvd, line 59: If this is always two workers then `[JoinHandle<()>; 2]` would be slightly more efficient and convey that intent.

src/host/pulseaudio/stream.rs
```diff
@@ -51,21 +52,46 @@ enum StreamInner {
...
 
 impl Drop for Stream {
     fn drop(&mut self) {
-        match &mut self.0 {
-            StreamInner::Playback(_, _, handle) | StreamInner::Record(_, _, handle) => {
-                handle.cancel()
+        match &mut self.inner {
+            StreamInner::Playback(stream, _, handle) => {
+                handle.cancel();
+                // Tear down the play_all driver thread spawned in
```
roderickvd: This reads as quite verbose. Consider trimming it to a line or two.

src/host/pulseaudio/stream.rs
```diff
@@ -219,9 +245,18 @@ impl Stream {
         // when the stream is stopped by the user.
         let stream_clone = stream.clone();
         let error_callback_clone = error_callback.clone();
-        std::thread::spawn(move || {
+        let cancel_driver = handle.cancel.clone();
+        let driver_handle = std::thread::spawn(move || {
             if let Err(e) = block_on(stream_clone.play_all()) {
-                emit_error(&error_callback_clone, Error::from(e));
+                // Stream::drop sets the cancel flag and then queues
```
roderickvd: Seems like it could be more to the point, too.

CHANGELOG.md
```diff
@@ -144,6 +144,19 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **JACK**: Fix input capture timestamp using callback execution time instead of cycle start.
 - **JACK**: Poisoned error callback mutex no longer silently drops subsequent error notifications.
 - **PulseAudio**: Poisoned locks now exit the thread gracefully instead of panicking.
+- **PulseAudio**: Fix server-side stream leak when a `Stream` is dropped: the
```
roderickvd: Much too long. Please ensure consistency with the existing entries.

roderickvd (changes requested):
Thanks for the PR. Please take care to keep the documentation style consistent with the rest and ensure what seems to be your LLM follows that.

Don't worry about the unrelated Rust nightly failure in the CI.

src/host/pulseaudio/stream.rs
knz: when recording (not playback) there is only one thread: `latency_handle`. I believe we need the Vec here.

src/host/pulseaudio/stream.rs
knz: done.

src/host/pulseaudio/stream.rs
knz: done.

CHANGELOG.md
knz: my apologies; done.

knz (commented):
Thank you for your feedback!

knz:
I'll handle the rebase; hold on

knz:
rebase completed

roderickvd:
Thank you!

PR #1194, opened 2026-05-09
fix: add init timeouts for PipeWire and PulseAudio

### LINKED ISSUE #1192: CPAL 0.18 devel with PipeWire enabled and PipeWire+PulseAudio running not working on FuriOS - Seems to block quietly on initialization
Hello,

As I mentioned on #1145 I got two user reports with logs about this:

* https://github.com/marin-m/SongRec/issues/296
* https://github.com/marin-m/SongRec/issues/276

These reports seem to suggest that the call to `cpal::default_host();` hangs and that CPAL hence doesn't even reach initialization (cf. https://github.com/marin-m/SongRec/issues/296#issuecomment-4408280665)

These reports also seem to suggest that CPAL 0.17 stable works on this platform

roderickvd:
(Hopefully) fixes #1192

### DIFF
--- a/src/host/pulseaudio/mod.rs
+++ b/src/host/pulseaudio/mod.rs
@@ -1,4 +1,4 @@
-use std::time::Duration;
+use std::{sync::mpsc, time::Duration};
 
 use futures::executor::block_on;
 use pulseaudio::protocol;
@@ -15,6 +15,8 @@ use crate::{
     SampleRate, StreamConfig, SupportedBufferSize, SupportedStreamConfig,
     SupportedStreamConfigRange,
 };
+
+const INIT_TIMEOUT: Duration = Duration::from_secs(2);
 
 const MIN_SAMPLE_RATE: SampleRate = 8000;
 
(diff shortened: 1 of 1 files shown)


marin-m:
works on my setup

roderickvd:
> works on my setup

Just to be sure: was your setup hanging before too and is it now fixed?

marin-m:
no issue on my setup but you requested me to test on the other ticket

roderickvd:
Yes would be great to hear from your users that suffered from it!

marin-m:
My users are reporting that switching the default backend to ALSA in Flathub builds fixed the issue for them so I guess that the hanging point is likely to be here (if you want them to try with the non-Flathub builds feel free to ask them on the original tickets).

FYI the next release of SongRec should try both ALSA and PipeWire as a default backend, falling back to the next one when anything fails even in the late device acquisition process, and the first backend to be tried will heuristically depend on the distribution channel + a CLI flag (see https://github.com/marin-m/SongRec/issues/306)

roderickvd:
Sounds good. I'll get this merged soon and hopefully it will aid https://github.com/marin-m/SongRec/issues/306 or - best case - make it superfluous. Feel free to open a new issue if things pop up again.

marin-m:
Noted, pushed your fix in rel 0.7.2 in the meantime

PR #1195, opened 2026-05-09
feat(WASAPI): allow setting raw stream mode

### LINKED ISSUE #1173: WASAPI: Raw stream mode
On Windows, the audio captured (through `build_input_stream`) seems to be passed through a default audio filter. I thought I had a bug in my code, because the sound recorded by `cpal` kept cutting out. This was particularly strange because sound from the same microphone in other apps that also didn't use exclusive mode (e.g. Discord) was fine. This issue did not happen on Linux. I found out that this issue was that Windows does audio processing by default. I did some further digging, and it seems that `cpal` does not use the `IAudioClient2::SetClientProperties`, which allows the `AUDCLNT_STREAMOPTIONS_RAW` flag to be set.

infiniwave:
Closes #1173. Has not yet been tested on other platforms than Windows.

### DIFF
--- a/src/host/asio/stream.rs
+++ b/src/host/asio/stream.rs
@@ -993,6 +993,7 @@ fn check_config(
         channels,
         sample_rate,
         buffer_size,
+        ..
     } = config;
 
     // Validate buffer size if `Fixed` is specified. This is necessary because ASIO's
--- a/src/host/wasapi/device.rs
+++ b/src/host/wasapi/device.rs
@@ -1,10 +1,10 @@
 use crate::{
     error::ResultExt,
     host::{com::ComString, ErrorCallbackArc},
-    BufferSize, Data, DeviceDescription, DeviceDescriptionBuilder, DeviceDirection, DeviceId,
-    DeviceType, Error, ErrorKind, FrameCount, InputCallbackInfo, InterfaceType, OutputCallbackInfo,
-    SampleFormat, SampleRate, StreamConfig, SupportedBufferSize, SupportedStreamConfig,
-    SupportedStreamConfigRange, COMMON_SAMPLE_RATES,
+    AudioProcessing, BufferSize, Data, DeviceDescription, DeviceDescriptionBuilder,
+    DeviceDirection, DeviceId, DeviceType, Error, ErrorKind, FrameCount, InputCallbackInfo,
+    InterfaceType, OutputCallbackInfo, SampleFormat, SampleRate, StreamConfig, SupportedBufferSize,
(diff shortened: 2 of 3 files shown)


roderickvd:
Thank you for this PR. Selectable raw mode is definitely something we should support. The question is of API design; a few different ideas have been proposed (builder in #1010, trait extensions in #1074, yours). I first want to get v0.18 out of the door and then pick up that discussion and decide how.

1rhino2 (commented):
windows review notes on this tip.

what works
- the WASAPI call site looks right: SetClientProperties with AUDCLNT_STREAMOPTIONS_RAW before Initialize, gated on IAudioClient2.
- cargo check --all-targets is green on the PR branch here (x86_64-pc-windows-gnu).
- solves the real pain from #1173 (shared-mode capture going through the system AEC/NS/AGC path).

design tension with #1220
- the v0.19 tracker lists RAW mode under "work out extension traits" (#1173), same bucket as ASIO control panel / PipeWire NO_CONVERT.
- this PR puts AudioProcessing on the shared StreamConfig instead. that is a breaking public-API addition and may not match the extension-trait direction @roderickvd is pushing in #1074.
- before merge i'd want an explicit call from maintainers: StreamConfig field vs WasapiDeviceExt / stream-option trait. if extension traits win, this should be reshaped rather than landing as-is.

other nits
- ~96 commits behind master; needs rebase. master still has bare StreamConfig { channels, sample_rate, buffer_size } literals (e.g. coreaudio ios) that will need the new field or `..`.
- no CHANGELOG / UPGRADING note for the breaking StreamConfig shape.
- SetClientProperties(...).ok() makes PreferRaw a silent no-op on failure. at least a debug log (or surface UnsupportedOperation when PreferRaw was requested and the call failed / IAudioClient2 missing) would help apps that think they got raw.
- AudioClientProperties leaves eCategory at Default. worth confirming against current windows-rs / WASAPI docs whether category should be set explicitly when Options=RAW.

couldn't finish a live PreferRaw smoke here: this host is windows-gnu without a working dlltool/link path, so binaries don't link even though check passes. happy to re-run once i have an msvc or mingw linker, or if someone else posts a short capture A/B.

roderickvd:
I haven’t decided yet on the extensions traits design. I’m working on a discussion piece on my other machine, that I don’t have access to right now.

When something like a qualified majority (no hard rule) of hosts support raw and/or exclusive with the same semantics, then it may warrant something on `StreamConfig` regardless - could live alongside extension traits too. (Is raw / exclusive / hog / passthrough mode whatever something in stream though or on the device?)

Long story short, we need to work it out properly before we can commit to a design.

roderickvd:
A proposal for the extension trait API is up at #1334

PR #1197, opened 2026-05-09
fix(alsa): check PCM RT-safety before promotion

### LINKED ISSUE #1193: coredump with default build of current master on Ubuntu 20.04LTS Pro
When starting an application compiled with the current master CPAL with default features on Ubuntu 20.04LTS, this is what happens:

`
Limiet op processortijd is overschreden (geheugendump gemaakt) English: Processor limit exceeded, coredumped.
`
apparently caused by the by default enabled real-time audio thread feature.

When running the same app with sudo this does not happen.

I did not try to change limits.conf, as you can't expect that from your "normal" user.

In my opinion this feature should not be enabled by default, it should be opt-in, and certainly not silently coredump.

The same problem does *not* happen on Debian Trixie with default CPAL features, I did not try other distributions.

roderickvd:
This should fix #1193 from the following issue:

When the user's ALSA device routes to PulseAudio through the `pcm_pulse` plugin, and cpal promotes the ALSA thread to realtime, then the core is dumped with `SIGXCPU`. The sequence is this:

1. `realtime-dbus` promotes the ALSA audio thread to RT via rtkit
2. rtkit grants `SCHED_FIFO` priority to that thread while enforcing `RLIMIT_RTTIME`
3. libpulse is not RT-safe and either spins or does expensive syscalls from the RT thread, burning through the `RLIMIT_RTTIME` budget
4. Kernel fires SIGXCPU and dumps core

This PR implements PCM type whitelisting for types that are RT-safe.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -115,6 +115,7 @@ jack = { version = "0.13", optional = true }
 
 [target.'cfg(any(target_os = "linux", target_os = "dragonfly", target_os = "freebsd", target_os = "netbsd"))'.dependencies]
 alsa = "0.11"
+alsa-sys = "0.4"
 libc = "0.2"
 audio_thread_priority = { version = "0.35", optional = true, default-features = false }
 jack = { version = "0.13", optional = true }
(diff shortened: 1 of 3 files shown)


roderickvd:
Note that this disables realtime promotion for `pipewire-alsa` even though that _is_ RT-safe, unlike `pulseaudio-alsa`. This errs on the safe side: both are PCM type `IOPLUG` and we can't distinguish between them further. 

Use the dedicated `pipewire` host and cpal feature flag if you want a realtime thread on PipeWire.

PR #1198, opened 2026-05-11
fix(coreaudio): aggregate-device uuid collision + auto start to true

umerhd:
Two independent bugs in macOS LoopbackDevice, one commit each.

  1. The aggregate-device UID was a hardcoded string literal, so concurrent LoopbackDevice instances collide on the same identity, and a crash before Drop leaves a stale registration that blocks the next launch. Suffix the UID with PID and a per-process atomic counter.
  2. kAudioAggregateDeviceTapAutoStartKey was set to false, so the sub-tap is not engaged when the aggregate device opens. The input stream looks live (valid sample rate, valid timestamps) but every buffer is silence. Flip to true.

### DIFF
--- a/src/host/coreaudio/macos/loopback.rs
+++ b/src/host/coreaudio/macos/loopback.rs
@@ -4,7 +4,10 @@ use std::{
     ffi::{c_void, CStr},
     mem::MaybeUninit,
     ptr::NonNull,
+    sync::atomic::{AtomicU32, Ordering},
 };
+
+static AGGREGATE_INSTANCE_COUNTER: AtomicU32 = AtomicU32::new(0);
 
 use objc2::{rc::Retained, AnyThread};
 use objc2_core_audio::{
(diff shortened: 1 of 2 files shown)


src/host/coreaudio/macos/loopback.rs
```diff
@@ -4,8 +4,11 @@ use std::{
     ffi::{c_void, CStr},
     mem::MaybeUninit,
     ptr::NonNull,
+    sync::atomic::{AtomicU64, Ordering},
 };
 
+static AGGREGATE_INSTANCE_COUNTER: AtomicU64 = AtomicU64::new(0);
```
roderickvd: `AtomicU32` would still hold 4 billion instances 😉

src/host/coreaudio/macos/loopback.rs
roderickvd: Should become `@YES` then?

roderickvd (commented):
Thanks for catching and fixing this! Two minor points if you'd take a look at them. And if you would also add a short entry to `CHANGELOG.md`?

src/host/coreaudio/macos/loopback.rs
```diff
@@ -168,14 +180,16 @@ fn to_cfstring(cstr: &'static CStr) -> CFRetained<CFString> {
 ///     @kAudioAggregateDeviceUIDKey :
 ///         @"com.josephlyncheski.MiniMetersAggregateDevice",
 ///     @kAudioAggregateDeviceTapListKey : taps,
-///     @kAudioAggregateDeviceTapAutoStartKey : @NO,
+///     @kAudioAggregateDeviceTapAutoStartKey : @YES,
 ///     // If we set this to NO then I believe we need to make the Tap public as
```
roderickvd: This comment can then be deleted, right?

roderickvd:
Thank you!

PR #1214, opened 2026-05-20
fix(alsa): hanging double-opens with BufferSize::Fixed om some drivers

### LINKED ISSUE #1157: alsa: xrun on RPi Zero since v0.17
While trying to update to the 0.17.0 version in my project [UbiHome](https://github.com/UbiHome/UbiHome/pull/104) (because of a [sendspin](https://github.com/Sendspin/sendspin-rs) update) I am getting the following errors in (roughly) this order on a Raspberry Pi Zero W Rev 1.1 using `alsa:plughw:CARD=sndrpihifiberry,DEV=0` (with `SupportedStreamConfig { channels: 2, sample_rate: 44100, buffer_size: Range { min: 1, max: 1 }, sample_format: F32 }`

```
Audio stream error: A backend-specific error has occurred: get_htstamp `10943.138735000` was earlier than get_trigger_htstamp `0.0`
Audio stream error: Buffer underrun/overrun occurred.
Audio stream error: A backend-specific error has occurred: `alsa::poll()` returned POLLERR
```

Each error probably occurs when adding an audio chunk to the playback buffer? See the [full log](https://github.com/user-attachments/files/26660943/example.log) to get a feeling for what I mean. 

Is this a cpal problem or should the sendspin library handle this version differently?

I initially updated from version 0.15.2, but while tracking down the behavior also tried 0.16.0 (which still works). 

I also saw https://github.com/RustAudio/cpal/issues/1134 already but I think because I am compiling musl I never hit that bug.

roderickvd:
The range validation for `BufferSize::Fixed` was calling `default_output_config()` before opening the stream, which opened and closed the PCM device once just to get the buffer size range. This may leave the driver in a bad state for the subsequent streaming open.

Second, to defend against hanging processes in such an event, the worker now checks PCM state when `poll()` times out.

Fixes #1157

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -88,6 +88,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **ALSA**: Streams no longer start automatically on creation; call `play()` manually.
 - **ALSA**: The `realtime` feature now skips RT promotion for ineligible PCM types (null,
   I/O plugins, wrapper types); `audio_thread_priority` promoted unconditionally.
+- **ALSA**: Supported configurations are now enumerated up to 64 channels (AES10 maximum).
 - **ASIO**: `Device::driver`, `asio_streams`, and `current_callback_flag` are no longer `pub`.
 - **ASIO**: Timestamps now include driver-reported hardware latency.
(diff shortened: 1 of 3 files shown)


DanielHabenicht:
As I wrote in the issue there are no apparent changes for me, but it also does not negatively effect the current behavior on the rasperry pi zero w, so I would say its better than before.

PR #1229, opened 2026-06-01
feat: add duplex stream API (interface only; backends to follow)

gulbrand:
Introduce the public duplex stream contract without any backend implementation. Hosts inherit trait defaults: supports_duplex() returns false and build_duplex_stream_raw returns UnsupportedOperation.

I have tested this with coreaudio/macos for a few months now as part of: https://github.com/RustAudio/cpal/pull/1096 (which I want to close in favor of this PR and a follow-up to add the coreaudio/macos support).

I am just now starting to test with ASIO on windows (still setting up my lab for this). The interface in this PR appears to be a good match for ASIO and coreaudio/macos. You can see the draft changes here for ASIO -- but these are totally untested and not ready for review: https://github.com/gulbrand/cpal/pull/5

My plan is to get this API PR merged, close PR 1096, finish the coreaudio/macos rebase/merge conflict resolve and open a new PR for the coreaudio/macos implementation.

As for ASIO support, I can't commit to that yet, but the interface looks ok to me from a high level.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,6 +44,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `realtime` feature for real-time audio thread scheduling without a D-Bus build dependency.
 - `StreamTrait::now()` to query the current instant on the stream's clock.
 - `StreamTrait::buffer_size()` to query the stream's current buffer size in frames per callback.
+- Added duplex stream API (interface only; backends to follow).
 - `SAMPLE_RATE_CD` (44100 Hz) and `SAMPLE_RATE_48K` (48000 Hz) constants.
 - `SupportedStreamConfigRange::try_with_standard_sample_rate()` and `with_standard_sample_rate()`
   to select 48 kHz or 44.1 kHz from a range.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -44,7 +44,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `realtime` feature for real-time audio thread scheduling without a D-Bus build dependency.
 - `StreamTrait::now()` to query the current instant on the stream's clock.
 - `StreamTrait::buffer_size()` to query the stream's current buffer size in frames per callback.
-- Added duplex stream API (interface only; backends to follow).
+- `DeviceTrait::build_duplex_stream()` and `supports_duplex()` for synchronized capture and playback on a shared clock (no backend support yet).
 - `SAMPLE_RATE_CD` (44100 Hz) and `SAMPLE_RATE_48K` (48000 Hz) constants.
 - `SupportedStreamConfigRange::try_with_standard_sample_rate()` and `with_standard_sample_rate()`
   to select 48 kHz or 44.1 kHz from a range.
--- a/src/duplex.rs
+++ b/src/duplex.rs
@@ -1,0 +1,47 @@
+use crate::{ChannelCount, InputStreamTimestamp, OutputStreamTimestamp, SampleRate};
+
+/// Information relevant to a single call to the user's duplex stream data callback.
+///
+/// Combines the input and output timestamps for the callback. Because a duplex stream's input and
+/// output share a single clock, both timestamps are drawn from the same time source.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+pub struct DuplexCallbackInfo {
+    input_timestamp: InputStreamTimestamp,
+    output_timestamp: OutputStreamTimestamp,
+}
+
+impl DuplexCallbackInfo {
+    /// Construct a `DuplexCallbackInfo` from its input and output timestamps.
+    pub fn new(
+        input_timestamp: InputStreamTimestamp,
+        output_timestamp: OutputStreamTimestamp,
+    ) -> Self {
+        Self {
+            input_timestamp,
+            output_timestamp,
+        }
+    }
+
+    /// The timestamp for the captured input data passed to the callback.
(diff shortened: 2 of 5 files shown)


StantonMatt (commented):
I took a pass over this as an API-only PR, not as backend validation.

This looks like a reasonable interface step to me. Existing hosts keep the default `supports_duplex() == false`, and `build_duplex_stream_raw()` defaults to `UnsupportedOperation`, so the PR is not accidentally making any backend claim support. The dynamic `platform::Device` forwarding also matches the existing input/output dispatch pattern, and `DuplexCallbackInfo` follows the current callback-info accessor style while keeping input/output timestamps separate.

The one API tradeoff I would call out before merge is that both `build_duplex_stream<T>` and `build_duplex_stream_raw()` use a single `SampleFormat` for input and output. That seems consistent with the synchronized ASIO/CoreAudio direction described here, but if a future backend needs different native sample formats for capture and playback, this API would need a follow-up shape rather than just a backend implementation.

I also checked the visible PR jobs individually: formatting, docs-Linux, Linux/macOS/Windows/iOS/tvOS/Android/WASM/WASI/PipeWire, sanitizer, and clippy jobs are green; only the release/publish docs jobs are skipped. No blocker from me on the interface/backward-compatibility side.

src/duplex.rs
```diff
@@ -0,0 +1,47 @@
...
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+pub struct DuplexStreamConfig {
+    /// The number of input (capture) channels.
+    pub input_channels: ChannelCount,
+    /// The number of output (playback) channels.
+    pub output_channels: ChannelCount,
+    /// The sample rate driving both directions.
+    pub sample_rate: SampleRate,
+    /// The desired buffer size, in frames per callback.
+    pub buffer_size: crate::BufferSize,
```
roderickvd: Nitpick: for consistency, you could import `BufferSize` into scope.

src/traits.rs
```diff
@@ -160,6 +162,18 @@ pub trait DeviceTrait: PartialEq + Eq + Hash + Debug + Display + Send + Sync {
             .is_ok_and(|mut iter| iter.next().is_some())
     }
 
+    /// True if the device can build a synchronized duplex stream where the captured input and
+    /// rendered output share a single clock.
+    ///
+    /// Returning `true` is a contract that input and output sides will run from one device-level
+    /// callback, or an OS driver aggregate (such as an Aggregate Device on MacOS).
```
roderickvd: Nitpick: macOS (note capitalization). Several occurrences in this file.

CHANGELOG.md
```diff
@@ -36,6 +19,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `realtime` feature for real-time audio thread scheduling without a D-Bus build dependency.
 - `StreamTrait::now()` to query the current instant on the stream's clock.
 - `StreamTrait::buffer_size()` to query the stream's current buffer size in frames per callback.
+- Added duplex stream API (interface only; backends to follow).
```
roderickvd: For style, we're not starting the line with "Added" for the other entries. Consider something like: 
> `DeviceTrait::build_duplex_stream()` and `supports_duplex()` for synchronized capture and playback on a shared clock (no backend support yet).

src/traits.rs
```diff
@@ -407,10 +421,106 @@ pub trait DeviceTrait: PartialEq + Eq + Hash + Debug + Display + Send + Sync {
...
+    /// the default implementation returns [`ErrorKind::UnsupportedOperation`].
+    ///
+    /// See [`build_duplex_stream`](Self::build_duplex_stream) for parameter and error
+    /// documentation.
+    ///
+    /// [`ErrorKind::UnsupportedOperation`]: crate::ErrorKind::UnsupportedOperation
+    fn build_duplex_stream_raw<D, E>(
+        &self,
+        _config: DuplexStreamConfig,
+        _sample_format: SampleFormat,
```
roderickvd: Would it be possible for a duplex stream to have its halves with different sample formats? If so, how likely would that be, and would we then want to support that or document it as a limitation?

roderickvd (commented):
Good approach to take it step-by-step.

src/traits.rs
gulbrand: I don't know. Maybe an aggregate device would allow for each direction to have a different sample format. 

Its easy to add this as a just in case.

gulbrand:
> I took a pass over this as an API-only PR, not as backend validation.
> 
> This looks like a reasonable interface step to me. Existing hosts keep the default `supports_duplex() == false`, and `build_duplex_stream_raw()` defaults to `UnsupportedOperation`, so the PR is not accidentally making any backend claim support. The dynamic `platform::Device` forwarding also matches the existing input/output dispatch pattern, and `DuplexCallbackInfo` follows the current callback-info accessor style while keeping input/output timestamps separate.
> 
> The one API tradeoff I would call out before merge is that both `build_duplex_stream<T>` and `build_duplex_stream_raw()` use a single `SampleFormat` for input and output. That seems consistent with the synchronized ASIO/CoreAudio direction described here, but if a future backend needs different native sample formats for capture and playback, this API would need a follow-up shape rather than just a backend implementation.
> 
> I also checked the visible PR jobs individually: formatting, docs-Linux, Linux/macOS/Windows/iOS/tvOS/Android/WASM/WASI/PipeWire, sanitizer, and clippy jobs are green; only the release/publish docs jobs are skipped. No blocker from me on the interface/backward-compatibility side.

Thank you! I will add the SampleFormat so both input and output have their own.

roderickvd:
Merging now!

roderickvd:
Copilot's review points make sense. I can commit them to `develop` directly.

PR #1231, opened 2026-06-02
fix(jack): resolve compilation errors

m-hilgendorf:


roderickvd:
Continuing from Discord, should we still up the minimum version in `Cargo.toml`?

m-hilgendorf:
I think that's a good idea. `jack@0.13.5` introduced a breaking change that the jack host backend relies on (`Client::sample_rate() -> u32`, used to be `usize`).

roderickvd:
Done in 7d4aedb.

PR #1232, opened 2026-06-03
Better implementation of example: remove latency

fengints:
passing pointer reference wihout the buffer is even faster.
But i don't know how to do it.

### DIFF
--- a/examples/feedback.rs
+++ b/examples/feedback.rs
@@ -2,10 +2,8 @@
 //!
 //! Assumes that the input and output devices can use the same stream configuration and that they
 //! support the f32 sample format.
-//!
-//! Uses a delay of `LATENCY_MS` milliseconds in case the default input and output streams are not
-//! precisely synchronised.
-
+use std::sync::Arc;
+use std::sync::Mutex;
 use clap::Parser;
 use cpal::{
     traits::{DeviceTrait, HostTrait, StreamTrait},
@@ -26,10 +24,6 @@ struct Opt {
     /// The output audio device to use
     #[arg(short, long, value_name = "OUT")]
     output_device: Option<String>,
-
-    /// Specify the delay between input and output
-    #[arg(short, long, value_name = "DELAY_MS", default_value_t = 150.0)]
-    latency: f32,
 
     /// Use the JACK host. Requires `--features jack`.
     #[arg(long, default_value_t = false)]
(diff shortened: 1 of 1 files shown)


roderickvd:
Thanks for the contribution! I'm not sure this is the right direction though. The latency buffer isn't optional: input and output callbacks run on separate threads with no guaranteed timing, so without the pre-fill the output callback will fire before the input has produced anything. That can cause glitches or silence when they drift far enough.

How would the mutex help? The guard is dropped immediately so I do not see how it provides synchronization. If the guard _were_ maintained until the end of the block, then the callbacks waiting for each other does not remove the need for pre-fill when one callback fires with nothing to consume yet.

Closing this one for now, as the current design is intentional. Feel free to reopen if I'm missing something.

fengints:
@roderickvd
Tested with alsa and jack backend. All clear. except 100% cpu usage, which i have problems even without all this.

https://doc.rust-lang.org/std/sync/struct.Mutex.html
So by the time he is copying. next thread already waits. and so on
```
{
    mtx_2.lock().unwrap();

}\\Mutex lock drops here
```

Problem with this is 
if thread 1 is (99.9% under load), trying to aquire lock. And second thread (~0-1% load) hits twice.
not relevant btw. if thread 1 loads that much, copying is also under trouble.
current implementation doesn't have defence against that problem either.

fengints:
@roderickvd 
That's not over. I found the way 2 replace pointers. 
But im hearing only noise. either this lib | jack | alsa are not ready for this.

That is the results for 1024 buffer size. >2.6 ghz laptop
Time taken to copy: 27.924µs
Time taken to ptr: 904ns

roderickvd:
As your lock guard is not assigned to any variable, it will be dropped immediately. Take a close look at the example you linked.

fengints:
@roderickvd

be not strict
i come from csharp, that would have worked there with manual unlocking.
Official documention says mutex drops at the end of the block.
having in mind rust's tight optimizations, your words can be true

roderickvd:
Locking a mutex returns a guard. If you do not use that result, or keep it around explicitly, it is dropped immediately. In the examples you see it's being used and _that's_ why the _guard_ drops at the end of the block.

```rust
{
    foo.lock().unwrap(); // guard discarded; lock freed immediately
    do_something();
}
```

```rust
{
    let _guard = foo.lock().unwrap();
    do_something();
} // _guard dropped here; lock freed here
```

PR #1234, opened 2026-06-04
Update feedback.rs

fengints:
optimizations may have took the place.

wrapped on mutexes

### DIFF
--- a/examples/feedback.rs
+++ b/examples/feedback.rs
@@ -2,10 +2,6 @@
 //!
 //! Assumes that the input and output devices can use the same stream configuration and that they
 //! support the f32 sample format.
-//!
-//! Uses a delay of `LATENCY_MS` milliseconds in case the default input and output streams are not
-//! precisely synchronised.
-
 use clap::Parser;
 use cpal::{
     traits::{DeviceTrait, HostTrait, StreamTrait},
@@ -15,6 +11,7 @@ use ringbuf::{
     traits::{Consumer, Producer, Split},
     HeapRb,
 };
+use std::sync::Mutex;
 
 #[derive(Parser, Debug)]
 #[command(version, about = "CPAL feedback example", long_about = None)]
@@ -26,10 +23,6 @@ struct Opt {
     /// The output audio device to use
     #[arg(short, long, value_name = "OUT")]
     output_device: Option<String>,
-
-    /// Specify the delay between input and output
-    #[arg(short, long, value_name = "DELAY_MS", default_value_t = 150.0)]
-    latency: f32,
 
     /// Use the JACK host. Requires `--features jack`.
     #[arg(long, default_value_t = false)]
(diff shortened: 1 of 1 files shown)


fengints:
Btw i tested reference passing the buffers.

1st and last bytes.

1 thread pass to data.
data from alsa: 0.016326904, 0.022766113
2 thread, data
data to jack: 0.016326904, 0.0

I assume drop called on 1st array. I hear only noise on output.
best results [buffer 1024]  [>2.6Ghz] (x30):
    Time taken to copy: 27.924µs
    Time taken to ptr: 904ns

examples/feedback.rs
```diff
@@ -102,36 +95,26 @@ fn main() -> anyhow::Result<()> {
...
-    for _ in 0..latency_samples {
-        // The ring buffer has twice as much space as necessary to add latency here,
-        // so this should never fail
-        producer.try_push(f32::EQUILIBRIUM).unwrap();
-    }
+    // Heap access, usually slower.
+    // you may use static buffers for higher performance.
+    let ring = HeapRb::<f32>::new(config.sample_rate as usize);
+    let (producer, consumer) = ring.split();
+    let pr_producer = Mutex::new(producer);
```
roderickvd, line 108: Creating two separate mutexes, then moving them into separate callbacks, does nothing.

examples/feedback.rs
```diff
@@ -102,36 +95,26 @@ fn main() -> anyhow::Result<()> {
...
+    let pr_consumer = Mutex::new(consumer);
 
     let input_data_fn = move |data: &[f32], _: &InputCallbackInfo| {
-        if producer.push_slice(data) < data.len() {
-            eprintln!("output stream fell behind: try increasing latency");
-        }
+        let mut producer = pr_producer.lock().unwrap();
```
roderickvd, line 112: Why do you want to introduce locking in a audio callback that should be RT-safe? It seems a regression.

examples/feedback.rs
```diff
@@ -102,36 +95,26 @@ fn main() -> anyhow::Result<()> {
...
-
-    // Pre-fill with silence equal to the length of the delay.
-    for _ in 0..latency_samples {
-        // The ring buffer has twice as much space as necessary to add latency here,
-        // so this should never fail
-        producer.try_push(f32::EQUILIBRIUM).unwrap();
-    }
+    // Heap access, usually slower.
+    // you may use static buffers for higher performance.
+    let ring = HeapRb::<f32>::new(config.sample_rate as usize);
```
roderickvd, line 106: This ignores channel count.

roderickvd (changes requested):

roderickvd:
I am afraid that I have to pass on this one, too. The mutexes are not what we want. `ringbuf`'s `split()` is already lock-free SPSC, which _is_ what we want.

What is your rationale for working on this feedback example (and #1232 before it)? Dropping the pre-fill and the under/overrun warnings also removes the jitter handling the example is meant to demonstrate.
  
Real-time audio threading can be counterintuitive. I appreciate you taking the time to get involved, but unless I'm missing something there may be other issues more worthwhile to tackle.

fengints:
@roderickvd don't do this ('putting 0 bytes upfront') was the answer.
must be a better way.

roderickvd:
It’s a basic example. You can set the latency to 0 if you are sure that there is no clock drift between the input and output devices (e.g same hardware clock or a system aggregate device with drift correction).

Issue #879, opened 2024-04-30
[Question/Feature] Get channel names of a device for certain platforms


nico-franco-gomez:
Some hosts such as CoreAudio or ASIO support having string names for each input/output channel of a device (besides their index). So far, I have not seen this functionality implemented in CPAL, but it would be very useful for a project where users can select input or output channels based on their names rather than the index. Is this a feature that would be worth implementing? Is there maybe a reason why this hasn't been done yet?

nico-franco-gomez:
Hello! I went ahead and wrote some code for a project where the channel names are provided for the aforementioned hosts. Would it be of interest here, even though it may not be available to all platforms? If so, I could start a PR with this functionality since more people might find it useful

roderickvd:
We've got `device_by_id` now with stable device IDs, I'd not be against a PR that would introduce something like `device_by_name`. Or are you saying it's on `Stream` rather than on `Device`?

nico-franco-gomez:
Well, a `device_by_name` would certainly be nice. But what I was talking about here was regarding names of the channels within a `Device`. There is, for instance, in ASIO: `ASIOGetChannelInfo()`, you would get a struct and there you have a `char[32]` which contains some name/description for the channel.

If you find this interesting, I'll try to start a PR with my implementation. As stated before, I've only done it for ASIO and CoreAudio. I'm actually not certain that the other hosts even support having channel names...

roderickvd:
Could you sketch out the API with a small example?

nico-franco-gomez:
I imagine something like this
```rust
pub trait DeviceTrait {
    fn get_channel_name(&self, channel_index: u16, input_channel: bool) -> Result<String, ChannelNameError>;
    // Alternatively
    fn get_input_channel_name(&self, channel_index: u16) -> Result<String, ChannelNameError>;
    fn get_output_channel_name(&self, channel_index: u16) -> Result<String, ChannelNameError>;

    // Optional: Getting all at the same time
    fn get_channel_names(&self, input_channels: bool) -> Result<Vec<String>, ChannelNameError>;
}

#[derive(Clone, Debug, PartialEq, Eq, Hash)]
pub enum ChannelNameError {
    NotImplemented,
    ChannelIndexOutOfRange,
    BackendSpecific { err: BackendSpecificError },
}

// Usage might look like
fn main() {
    assert!("Left" == device.get_channel_name(0, false).unwrap());
    assert!("Right" == device.get_channel_name(1, false).unwrap());

[...]

PR #1254, opened 2026-06-25
Channel names: CoreAudio and ASIO

nico-franco-gomez:
This PR refers to this [issue](https://github.com/RustAudio/cpal/issues/879).

I implemented a version of this to showcase how I imagined it. When I wrote the issue, I hadn't realized that the error convention had changed from 0.16 to 0.17. This PR adheres to the current error convention.

### DIFF
--- a/asio-sys/src/bindings/mod.rs
+++ b/asio-sys/src/bindings/mod.rs
@@ -943,6 +943,18 @@ impl Driver {
         let mut dcb = DRIVER_EVENT_CALLBACKS.lock().unwrap();
         dcb.retain(|&(id, _)| id != rem_id);
     }
+
+    /// Returns the name of the channel at the given index.
+    ///
+    /// `channel` is a 0-based channel index. `is_input` selects the input (`true`) or output
+    /// (`false`) direction.
+    ///
+    /// The driver must already be loaded (i.e. this `Driver` instance must be alive).
+    pub fn channel_name(&self, channel: i32, is_input: bool) -> Result<String, AsioError> {
+        let _guard = self.inner.lock_state();
+        let info = asio_channel_info(channel, is_input)?;
+        Ok(driver_name_to_utf8(&info.name).into_owned())
+    }
 }
 
 impl DriverState {
--- a/asio-sys/src/bindings/mod.rs
+++ b/asio-sys/src/bindings/mod.rs
@@ -7,7 +7,7 @@ pub mod errors;
 #[cfg(target_os = "windows")]
 use std::os::raw::c_long;
 use std::{
-    ffi::{CStr, CString},
+    ffi::CString,
     os::raw::{c_char, c_double, c_void},
     ptr::null_mut,
     sync::{
@@ -1149,7 +1149,15 @@ fn stream_data_type(is_input: bool) -> Result<AsioSampleType, AsioError> {
 ///
 /// This converts to utf8.
 fn driver_name_to_utf8(bytes: &[c_char]) -> std::borrow::Cow<'_, str> {
-    unsafe { CStr::from_ptr(bytes.as_ptr()).to_string_lossy() }
+    let length = bytes
+        .iter()
+        .position(|&byte| byte == 0)
+        .unwrap_or(bytes.len());
+    let bytes = bytes[..length]
+        .iter()
+        .map(|&byte| byte as u8)
+        .collect::<Vec<_>>();
+    String::from_utf8_lossy(&bytes).into_owned().into()
 }
 
 /// Convert an `ASIOTimeStamp` (high and low 32-bit halves) to a `u64` nanosecond value.
--- a/asio-sys/src/bindings/mod.rs
+++ b/asio-sys/src/bindings/mod.rs
@@ -981,6 +981,18 @@ impl Driver {
         drop(dcb);
         drop(removed);
     }
+
+    /// Returns the name of the channel at the given index.
+    ///
+    /// `channel` is a 0-based channel index. `is_input` selects the input (`true`) or output
+    /// (`false`) direction.
+    ///
+    /// The driver must already be loaded (i.e. this `Driver` instance must be alive).
+    pub fn channel_name(&self, channel: i32, is_input: bool) -> Result<String, AsioError> {
+        let _guard = self.inner.lock_state();
+        let info = asio_channel_info(channel, is_input)?;
+        Ok(driver_name_to_utf8(&info.name).into_owned())
+    }
 }
 
 impl DriverState {
--- a/asio-sys/src/bindings/mod.rs
+++ b/asio-sys/src/bindings/mod.rs
@@ -1119,6 +1119,10 @@ fn asio_get_buffer_sizes() -> Result<BufferSizes, AsioError> {
 /// Retrieve the `ASIOChannelInfo` associated with the channel at the given index on either the
 /// input or output stream (`true` for input).
 fn asio_channel_info(channel: c_long, is_input: bool) -> Result<ai::ASIOChannelInfo, AsioError> {
+    if channel < 0 {
+        return Err(AsioError::InvalidInput);
+    }
+
     let mut channel_info = ai::ASIOChannelInfo {
         // Which channel we are querying
         channel,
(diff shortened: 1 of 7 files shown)


roderickvd:
Sorry for the late reply on this one.

I think this is a worthwhile addition as a trait method. You could add a default implementation to the trait method to prevent the duplication at each backend that doesn't support it.

Before we go further, could you rebase this onto the `develop` branch? We're staging upcoming feature work there rather than master right now. I intend to promote it to `master` for a v0.19 release after we push out a v0.18.2 bug fix release.

Happy to take a deeper look once it's rebased!

nico-franco-gomez:
No problem regarding the delay.

I rebased and implemented the comments while trying to keep a compact, readable git history. Not sure why that specific Linux check is failing, though...

src/host/coreaudio/macos/device.rs
```diff
@@ -1116,3 +1138,45 @@ pub(crate) fn get_device_buffer_frame_size(
     )?;
     Ok(frames as usize)
 }
+
+fn get_channel_name_for_device(
+    device_id: AudioDeviceID,
+    channel_index: u16,
```
roderickvd, line 1081: Querying an out-of-range `channel_index` will return a `BackendError` I think, when `InvalidInput` would be preferred.

src/host/asio/device.rs
```diff
@@ -213,6 +235,13 @@ impl Iterator for Devices {
                         .filter(|&r| driver.can_sample_rate(r.into()).unwrap_or(false))
                         .collect();
 
+                    let input_channel_names: Box<[String]> = (0..channels.ins)
+                        .map(|ch| driver.channel_name(ch, true).unwrap_or_default())
```
roderickvd: Question: rather than defaulting to an empty string, what if we returned something recognizable like "Channel {ch}"?

asio-sys/src/bindings/mod.rs
```diff
@@ -943,6 +943,18 @@ impl Driver {
...
+    /// Returns the name of the channel at the given index.
+    ///
+    /// `channel` is a 0-based channel index. `is_input` selects the input (`true`) or output
+    /// (`false`) direction.
+    ///
+    /// The driver must already be loaded (i.e. this `Driver` instance must be alive).
+    pub fn channel_name(&self, channel: i32, is_input: bool) -> Result<String, AsioError> {
+        let _guard = self.inner.lock_state();
+        let info = asio_channel_info(channel, is_input)?;
+        Ok(driver_name_to_utf8(&info.name).into_owned())
```
roderickvd, line 994: I think this was already the case elsewhere with other names, but it's occurring to me that this will UB if there's ever a driver that doesn't NUL-terminate.

src/host/asio/device.rs
```diff
@@ -213,6 +235,13 @@ impl Iterator for Devices {
                         .filter(|&r| driver.can_sample_rate(r.into()).unwrap_or(false))
                         .collect();
 
+                    let input_channel_names: Box<[String]> = (0..channels.ins)
```
roderickvd: We may want to check that the number of channels is greater than (or equal to) 0. I remember that ASIO often returns an `i32` and we wouldn't want a negative value overflowing this.

src/host/coreaudio/macos/device.rs
```diff
@@ -1116,3 +1138,45 @@ pub(crate) fn get_device_buffer_frame_size(
...
+    input: bool,
+) -> Result<String, Error> {
+    let mut channel_name: *mut CFString = std::ptr::null_mut();
```
roderickvd: As this `CFString` stuff is getting used more often, we could consider having a helper for it.

src/host/coreaudio/macos/device.rs
```diff
@@ -690,6 +694,24 @@ impl Device {
             .map(|mut configs| configs.next().is_some())
             .unwrap_or(false)
     }
+
+    fn get_channel_name(&self, channel_index: u16, input: bool) -> Result<String, Error> {
+        if input && !self.supports_input() {
```
roderickvd: `supports_input/output` aren't free, so if it's just to detail the error message then maybe we should just use the `input` argument but not re-query the device (or cache it, but that may be scope creep).

src/host/asio/device.rs
```diff
@@ -26,6 +26,8 @@ pub struct Device {
     input_sample_format: Option<SampleFormat>,
     output_sample_format: Option<SampleFormat>,
     supported_sample_rates: Box<[SampleRate]>,
+    input_channel_names: Box<[String]>,
```
roderickvd, line 29: CoreAudio doesn't seem to cache this. What's the preferred approach? Lazily like CoreAudio or caching it during enumeration here?

asio-sys/src/bindings/mod.rs
```diff
@@ -943,6 +943,18 @@ impl Driver {
         let mut dcb = DRIVER_EVENT_CALLBACKS.lock().unwrap();
         dcb.retain(|&(id, _)| id != rem_id);
```
roderickvd: cpal doesn't use `get_` for accessors.

asio-sys/src/bindings/mod.rs
```diff
@@ -943,6 +943,18 @@ impl Driver {
...
     }
+
+    /// Returns the name of the channel at the given index.
+    ///
+    /// `channel` is a 0-based channel index. `is_input` selects the input (`true`) or output
+    /// (`false`) direction.
+    ///
+    /// The driver must already be loaded (i.e. this `Driver` instance must be alive).
+    pub fn channel_name(&self, channel: i32, is_input: bool) -> Result<String, AsioError> {
```
roderickvd, line 991: Not sure about the `input: bool` or `is_input: bool` arguments in public functions. In the rest of cpal, that's split between `supports_input/output`, `default_input/output_config`, etc. That's more readable than `channel_name(1, true)` - what argument isn't self-explanatory.

roderickvd (changes requested):
Thanks again, please find my first review attached.

I've promoted `develop` to `master`, so please rebase on `master` where we'll continue the 0.19.0 work. This should be easier to the last rebase, as it's just continuing the `develop` branch under a new name, basically.

src/host/coreaudio/macos/device.rs
nico-franco-gomez: Good point, I introduced a new check for that in here

src/host/asio/device.rs
nico-franco-gomez: I introduced these default names, but I thought about the case where the driver leaves that unfilled knowingly, so I'd argue that for the low-level crate, the caller should decide what the default name in that case should be. What do you think?

asio-sys/src/bindings/mod.rs
nico-franco-gomez: Good catch

src/host/asio/device.rs
nico-franco-gomez: Done

src/host/coreaudio/macos/device.rs
nico-franco-gomez: Done

src/host/coreaudio/macos/device.rs
nico-franco-gomez: True. Since there's now the channel_index check, the error message is descriptive enough so that these checks are redundant.

asio-sys/src/bindings/mod.rs
nico-franco-gomez: Here you're referring to the `get_channel_name` method, right?

asio-sys/src/bindings/mod.rs
nico-franco-gomez: before rewriting this completely, what do you think could be a better API? Maybe `input_channel_name(1)` and `output_channel_name(1)` ?

src/host/asio/device.rs
nico-franco-gomez: I honestly prefer to do it lazily, my rationale is the following: the operation for CoreAudio is low-cost and can be done on demand. Since most users won't need it, I prefer for a lower-level crate to avoid such implicit operations. The operation for ASIO on the other hand is not low-cost because it requires loading the driver and potentially unloading the previous one. It's also risky if there's a callback running in another thread. That's why I find it reasonable to cache it there, but here, I would argue it's unnecessary.

PR #1261, opened 2026-07-09
chore(alsa): bump to 0.12

sertonix:
alsa 0.12 includes fixes for compiling on time64 enabled libc: https://github.com/diwic/alsa-rs/pull/157

CC @Gelbpunkt

Edit: The clippy-Windows failure looks unrelated since this MR does not change anything for windows.

PR #1262, opened 2026-07-10
Declare the AVAudioSession dependencies on visionOS

SkeletonGamer:
The CoreAudio backend gates its AVAudioSession code path on every non-macOS Apple platform:

```rust
#[cfg(all(target_vendor = "apple", not(target_os = "macos")))]
```

but `block2`, `objc2-foundation` and `objc2-avf-audio` are declared only for `ios` and `tvos`.
On visionOS the code is compiled while the crates are absent, so the build stops at:

```
error[E0432]: unresolved import `objc2_avf_audio`
error[E0432]: unresolved import `block2`
error[E0432]: unresolved imports `objc2_foundation::NSNotification`, `objc2_foundation::NSNotificationCenter`
```

This looks like an oversight in #1155, which added `tvos` to that list. Everything below is already
in place: coreaudio-rs gained visionOS support in RustAudio/coreaudio-rs#148, and mach2 0.6, the
version cpal requires, builds for every Apple vendor.

Builds, on master with rustc 1.97 and Xcode 26.6:

| target | before | after |
| --- | --- | --- |
| aarch64-apple-darwin | ok | ok |
| aarch64-apple-ios | ok | ok |
| aarch64-apple-tvos | ok | ok |
| aarch64-apple-visionos | fails to build | ok |

Runtime, on the visionOS 26.2 simulator (`xcrun simctl spawn`), built for
`aarch64-apple-visionos-sim`:

- `examples/enumerate` reports the CoreAudio host and the default device, 48 kHz, F32, 2 channels out.
- `examples/beep` opens an output stream and the sine is audible.
- rodio at master, which already requires cpal 0.18, plays a `SineWave` through
  `DeviceSinkBuilder::open_default_sink` and exits cleanly.

I have not run this on Vision Pro hardware, only on the simulator.

I did not widen the cfg to `all(target_vendor = "apple", not(target_os = "macos"))`, which is what
the code gate says. watchOS still fails further down inside coreaudio-rs
(`cannot find value buffer_frame_size`), so claiming it in the manifest would only move the error.
Adding `visionos` alone keeps the manifest honest about what builds today.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,10 +16,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- **visionOS**: The CoreAudio backend now builds. Its AVAudioSession dependencies were only
-  declared for iOS and tvOS.
 - Timestamps now stay monotonic across device and graph changes.
 - **ALSA**: A nonzero but sub-millisecond stream timeout is no longer treated as a non-blocking poll.
+- **visionOS**: The CoreAudio backend now builds.
 - **WASAPI**: Reported buffer sizes are no longer off by one frame.
 
 ## [0.18.1] - 2026-06-07
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **visionOS**: The CoreAudio backend now builds. Its AVAudioSession dependencies were only
+  declared for iOS and tvOS.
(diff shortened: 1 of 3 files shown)


SkeletonGamer:
The failing `clippy-Linux` and `clippy-Windows` checks are unrelated to this PR, which only touches
`Cargo.toml` and `CHANGELOG.md`. `clippy::question_mark` now fires on `src/host/alsa/mod.rs:610` and
`src/host/asio/device.rs:183` with Rust 1.97, released yesterday, and CI runs with `-D warnings`.
#1261 hits the same failure, and the runs that predate 1.97 are green.

#1263 fixes both.

roderickvd:
Thanks! Made some minor edits to the changelog and readme. Merging now.

PR #1263, opened 2026-07-10
Fix clippy::question_mark on Rust 1.97 (ALSA and ASIO)

SkeletonGamer:
Rust 1.97, released on July 9, extends `clippy::question_mark` to `match` expressions whose arms
map cleanly onto `?`. CI runs clippy with `-D warnings`, so `clippy-Linux` and `clippy-Windows`
now fail on master and on every open pull request:

```
error: this `match` expression can be replaced with `?`
   --> src/host/alsa/mod.rs:610:35
error: this `match` expression can be replaced with `?`
   --> src\host\asio\device.rs:183:13
```

Both are mechanical. In the ASIO iterator the `match` on `self.drivers.next()` becomes
`let name = self.drivers.next()?;`, which returns `None` from `next` exactly as the old
`None => return None` arm did. The rest of that hunk is reindentation, and the diff is empty once
whitespace is ignored.

Verified for ALSA by running the Linux job's own command, `cargo clippy --all --all-features
-- -D warnings`, on rust:1.97: it fails on master and passes with this change. The ASIO backend
only builds on Windows with the ASIO SDK, so I could not run clippy against it locally and I am
relying on this PR's own `clippy-Windows` job. `cargo fmt --all --check` is clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,8 +16,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
-- **ALSA**, **ASIO**: Two `match` expressions are replaced with `?`, which `clippy::question_mark`
-  started flagging in Rust 1.97.
 - Timestamps now stay monotonic across device and graph changes.
 - **ALSA**: A nonzero but sub-millisecond stream timeout is no longer treated as a non-blocking poll.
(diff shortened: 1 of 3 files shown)


roderickvd:
And thanks again 😄 
Removed the changelog entry here: clippy fixes are not observable changes.

PR #1272, opened 2026-07-14
feat: duplex support for pipewire

Decodetalkers:
still in draft, I need to test more

### DIFF
--- a/src/host/pipewire/stream.rs
+++ b/src/host/pipewire/stream.rs
@@ -33,8 +33,8 @@ use pipewire::{
 };
 
 use crate::{
-    CallbackInfo, Data, Error, ErrorKind, FrameCount, SampleFormat, StreamConfig, StreamInstant,
-    StreamTimestamp,
+    CallbackInfo, Data, DuplexCallbackInfo, DuplexStreamConfig, Error, ErrorKind, FrameCount,
+    SampleFormat, StreamConfig, StreamInstant, StreamTimestamp,
     host::{
         ErrorCallbackArc, Notify, emit_error, equilibrium::fill_equilibrium, frames_to_duration,
         latch::Latch, try_emit_error,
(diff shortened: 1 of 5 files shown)


src/host/pipewire/stream.rs
```diff
@@ -1129,7 +1240,453 @@ where
...
+
+                // Pre-fill only the active region with equilibrium before handing it to the
+                // callback.
+                let active = &mut samples[..frames * stride];
+                fill_equilibrium(active, user_data.sample_format);
+
+                let data = active.as_mut_ptr() as *mut ();
+                let mut data =
+                    unsafe { Data::from_parts(data, n_samples, user_data.sample_format) };
+                user_data.publish_data_duplex_out(stream, frames, &mut data, xrun);
```
Decodetalkers, line 1606: because input and output are in two channels, so I cannot send input and output in the same time.. what is about other protocols?

roderickvd:
@gulbrand

roderickvd:
I've promoted `develop` to `master`, so please rebase on `master` where we'll continue the 0.19.0 work.

Decodetalkers:
I found this pr https://gitlab.freedesktop.org/pipewire/pipewire-rs/-/merge_requests/132/

Maybe I should wait for it?

roderickvd:
Indeed that would be preferred. Otherwise we have to interface with SPA and the filter API ourselves.

src/host/pipewire/stream.rs
roderickvd: Probably because this is the stream API and we should use the filter API.

roderickvd:
Oh but in hindsight, that's a PR has been sitting there for four years with no movements for over a year. So either:
1. someone would have to pick up that effort again
2. we could use the `pw_filter_*` stuff ourselves from `pipewire-sys`

Decodetalkers:
have made a pr for pipewire https://gitlab.freedesktop.org/pipewire/pipewire-rs/-/merge_requests/276, but still problems that I haven't solved, can you also take a look @roderickvd 

it seems not that easy to use the ffi part directly, so I think I still need to make the pr merged first. There are still many problems I cannot solve, so I will wait for the review from pipewire-rs first

Issue #1260, opened 2026-07-08
`HostId` enum Refactor


cbrownsey:
The `HostId` enum currently depends very heavily on the features `cpal` is compiled with, as each variant is only available if the corresponding backend is supported in the current build. This means that writing code which can switch between different backends, or which even just wants to be aware of which backend is currently being used must convert it into a string, and work with that. This is error prone and laborious.

I propose changing the `HostId` to be `#[non_exhaustive]`, and to include every backend which is supported by `cpal` under any possible combination of features. The `SUPPORTED_HOSTS` associated constant, and `is_supported` method will surface the same information as the `ALL_HOSTS` constant currently does before these changes, with `is_supported` being equivalent to being included in `SUPPORTED_HOSTS`. The `available_hosts` iterator method and `is_available` method will surface the same information that the `available_hosts` function currently exposes, without the mandatory `Vec` allocation.

The items `HostId`, `ALL_HOSTS`, `available_hosts`, `default_host`, and `host_from_id` will also be pulled out from behind any feature gates to improve the documentation UX, as they aren't feature-dependent. This does add another failure mode to `host_from_id`, that being passing the `HostId` of an unsupported backend, in which case it will return `ErrorKind::UnsupportedOperation`.

`HostId` is also given a `Default` implementation, which is equivalent to the id of the host returned by `default_host`. This is primarily for the sake of the implementation, as opposed to being something users are inspecting, but it may also be useful for that purpose.

I'm not necessarily proposing deprecating `ALL_HOSTS` and `available_hosts`, although they would be made redundant should this be implemented, so they could be deprecated.

```rust
#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
#[non_exhaustive]
enum HostId {
    AAudio,
    Alsa,
    Asio,
    AudioWorklet,
    CoreAudio,
    Custom,
    Jack,
    Null,
    Wasapi,
    WebAudio,
}

impl HostId {
    /// Every `HostId` supported by the current configuration of cpal.
    const SUPPORTED_HOSTS: &[Self];
    
    /// Returns `true` if the given host is supported by the current configuration of cpal.
    const fn is_supported(&self) -> bool;
    
    /// Returns `true` if the given host is currently available on this platform.
    fn is_available(&self) -> bool;
    
    /// Returns an iterator over the hosts which are currently available on this platform.
    const fn available_hosts() -> AvailableHostsIter;
}

struct AvailableHostsIter { .. }

[...]

PR #1274, opened 2026-07-15
`HostId` enum Refactor

cbrownsey:
Original Issue: #1260 

This implements the change to having `HostId` include a variant for all possible hosts under any compilation configuration.

This also adds an implementation of `TryFrom<HostId> for Host`, which seemed the best way to have that. This also keeps `available_hosts`, `host_from_id`, and `ALL_HOSTS`, although they no longer seem relevant to keep anymore.

### DIFF
--- a/examples/beep.rs
+++ b/examples/beep.rs
@@ -75,15 +75,15 @@ fn main() -> anyhow::Result<()> {
     // cargo run --release --example beep --features jack -- --jack
     let host = if opt.jack {
         jack_host_id
-            .and_then(cpal::host_from_id)
+            .and_then(TryFrom::try_from)
             .expect("make sure `--features jack` is specified, and the platform is supported")
     } else if opt.pulseaudio {
         pulseaudio_host_id
-            .and_then(cpal::host_from_id)
+            .and_then(TryFrom::try_from)
             .expect("make sure `--features pulseaudio` is specified, and the platform is supported")
     } else if opt.pipewire {
         pipewire_host_id
-            .and_then(cpal::host_from_id)
+            .and_then(TryFrom::try_from)
             .expect("make sure `--features pipewire` is specified, and the platform is supported")
     } else {
         cpal::default_host()
--- a/src/platform/host_id.rs
+++ b/src/platform/host_id.rs
@@ -90,7 +90,7 @@ impl std::str::FromStr for HostId {
     /// # Errors
     ///
     /// - [`ErrorKind::UnsupportedOperation`] if the string does not name a
-    /// valid `HostId.
+    ///   valid `HostId.
     ///
     /// [`ErrorKind::UnsupportedOperation`]: crate::ErrorKind::UnsupportedOperation
     fn from_str(s: &str) -> Result<Self, Self::Err> {
--- a/src/platform/host_id.rs
+++ b/src/platform/host_id.rs
@@ -1,0 +1,195 @@
+use crate::platform;
+
+/// A unique identifier for each host supported by CPAL.
+///
+/// Not all hosts in this enum are available at runtime, or are even supported
+/// by the current platform. This can be checked with `is_available` or
+/// `is_supported` respectively.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+#[non_exhaustive]
+pub enum HostId {
+    AAudio,
+    Alsa,
+    Asio,
+    AudioWorklet,
+    CoreAudio,
+    Custom,
+    Jack,
+    Null,
+    PipeWire,
+    PulseAudio,
+    Wasapi,
+    WebAudio,
+}
+
+impl HostId {
+    /// All hosts supported by CPAL on this platform.
+    pub const SUPPORTED_HOSTS: &[HostId] = {
+        // This is a hack to prevent rustdoc from referencing the
+        // implementation const in its output.
+        let _ = 1 + 2;
+
+        super::SUPPORTED_HOSTS
+    };
+
+    /// Returns the human-readable host name.
+    pub const fn name(&self) -> &'static str {
+        match self {
+            HostId::AAudio => "AAudio",
+            HostId::Alsa => "ALSA",
+            HostId::Asio => "ASIO",
+            HostId::AudioWorklet => "AudioWorklet",
+            HostId::CoreAudio => "CoreAudio",
+            HostId::Custom => "Custom",
+            HostId::Jack => "JACK",
+            HostId::Null => "Null",
+            HostId::PipeWire => "PipeWire",
+            HostId::PulseAudio => "PulseAudio",
+            HostId::Wasapi => "WASAPI",
+            HostId::WebAudio => "WebAudio",
+        }
+    }
+
+    /// Checks if the given `HostId` is supported on this platform.
+    pub const fn is_supported(&self) -> bool {
+        super::is_supported_impl(*self)
+    }
+
+    /// Checks if the given `HostId` is currently available.
+    pub fn is_available(&self) -> bool {
+        super::is_available_impl(*self)
+    }
+
+    /// Iterates over all the `HostId`s currently available on this platform.
+    ///
+    /// The availability check is performed when `next` is called, not when
+    /// this function is called.
+    pub fn available_hosts() -> AvailableHostsIter {
+        AvailableHostsIter(Self::SUPPORTED_HOSTS.iter())
+    }
+}
+
+impl Default for HostId {
+    fn default() -> Self {
+        super::default_host_id()
+    }
+}
+
+impl std::fmt::Display for HostId {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.write_str(&self.name().to_ascii_lowercase())
+    }
+}
+
+impl std::str::FromStr for HostId {
+    type Err = crate::Error;
+
+    /// Parse a host identifier from its string representation (e.g. `"alsa"`,
+    /// `"coreaudio"`). This conversion is case-insensitive.
+    ///
+    /// # Errors
+    ///
+    /// - [`ErrorKind::UnsupportedOperation`] if the string does not name a
+    /// valid `HostId.
+    ///
+    /// [`ErrorKind::UnsupportedOperation`]: crate::ErrorKind::UnsupportedOperation
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        macro_rules! match_str_case_insensitive {
+            (
+                $s:expr => {
+                    $( $l:literal => $e:expr, )*
+                    _ => $f:expr $(,)?
+                }
+            ) => {
+                {
+                    let s = $s;
+
+                    if false { unreachable!() }
+
+                    $(
+                        else if $l.eq_ignore_ascii_case(s) { $e }
+                    )*
+
+                    else { $f }
+                }
+            };
+        }
+
+        match_str_case_insensitive! {
+            s => {
+                "AAudio" => Ok(HostId::AAudio),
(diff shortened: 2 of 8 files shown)


cbrownsey:
Ack! Rookie mistake. All green now, but I'm not sure if I'm entirely keen on pushing as-is. I think the `HostId` enum should be pulled out of the platform module and be re-exported at the crate root. Additionally, I'm thinking that the `available_hosts`, `host_from_id` and `ALL_HOSTS` items should be canned. But this is what I outlined in the initial issue, so that's what I'm committing right now.

1rhino2 (commented):
static pass on the HostId refactor (no runtime concerns here).

direction looks right for #1260: a single non_exhaustive HostId with every variant, plus is_supported / is_available / available_hosts(), and TryFrom<HostId> for Host. that is cleaner than cfg-gated enums for cross-platform user code.

a few things to nail before merge
- AvailableHostsIter::next currently returns hosts where is_supported() is true, but the docs on available_hosts / HostId::available_hosts talk about availability. supported != available (daemon down, etc). looks like the iterator should call is_available(), or the naming/docs need to match whichever filter you want.
- Display writes ascii-lowercase ("wasapi") while FromStr examples/docs show mixed case like "WASAPI" / "ALSA". case-insensitive parse is fine; worth one golden test that Display round-trips through FromStr.
- body notes available_hosts / host_from_id / ALL_HOSTS may be obsolete. if they stay, mark deprecated in the same PR so the migration path is obvious; if they go, UPGRADING.md needs a short note.
- needs a rebase check against current master (this sat a bit).

i can help poke windows HostId::Wasapi / Asio availability after rebase if useful.

roderickvd:
Valid point that supported isn’t equal to available. But to your last point, in which cases would that be different for WASAPI or ASIO?

cbrownsey:
I've made some changes. I decided to go for removing `available_hosts`, `ALL_HOSTS`, and `host_from_id`. I added a very short section about how to go from a `HostId` to a `Host`, which seems enough to me. 

Thank you for spotting that I did use the wrong method in `AvailableHostsIter`, not sure what that was about. And I did remove the lowercasing in the `Display` method. I kept that from the previous implementation, but I never did like it as it felt like a very unnecessary allocation. I also added a test to ensure that the Display implementation round-trips, as you mentioned.

I've also updated some of the example files to have them compile with my changes, but they do still have the `cfg`s in them that this whole exercise is to remove, so I'll have those fixed up tomorrow. Other than that, I'm comfortable with the changes as they are.

examples/beep.rs
```diff
@@ -75,15 +75,15 @@ fn main() -> anyhow::Result<()> {
     // cargo run --release --example beep --features jack -- --jack
     let host = if opt.jack {
         jack_host_id
-            .and_then(cpal::host_from_id)
+            .and_then(TryFrom::try_from)
```
roderickvd, line 78: I like using `TryFrom`. How about keeping `host_from_id` around for one more release cycle and tagging it as deprecated?

src/platform/mod.rs
```diff
@@ -882,23 +796,14 @@ mod platform_impl {
...
     );
 
-    /// The default host for the current compilation target platform.
-    pub fn default_host() -> Host {
-        #[cfg(feature = "pipewire")]
-        if <PipeWireHost as crate::traits::HostTrait>::is_available() {
-            if let Ok(host) = PipeWireHost::new() {
-                return host.into();
-            }
+    pub(crate) fn default_host_id() -> HostId {
```
roderickvd, line 799: This is silently removing the fallback paths checking whether PipeWire and PulseAudio aren't just supported but actually available. If they aren't, rather than falling back to ALSA, it will now panic.

src/platform/host_id.rs
```diff
@@ -0,0 +1,225 @@
...
+
+impl Default for HostId {
+    fn default() -> Self {
+        super::default_host_id()
+    }
+}
+
+impl std::fmt::Display for HostId {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.write_str(self.name())
```
roderickvd, line 90: Though you did say you don't like the allocation, I'm not clear why you want to drop the convenience contract and consistency of using `.to_ascii_lowercase()`.

If we do change then it needs to be mentioned in the changelog, because it will break existing string-compares and hash keys. But I'd first like to settle changing at all.

src/platform/mod.rs
```diff
@@ -931,27 +835,17 @@ mod platform_impl {
...
-    /// where `AudioContext` is unavailable.
-    pub fn default_host() -> Host {
-        assert!(
-            WebAudioHost::is_available(),
-            "WebAudio is not available in this context; \
-             AudioContext requires a Window (not a Worker or Service Worker)"
-        );
-        WebAudioHost::new().unwrap().into()
+    pub(crate) fn default_host_id() -> HostId {
+        HostId::WebAudio
```
roderickvd, line 848: What's your rationale for dropping the assertion that was there?

src/platform/host_id.rs
```diff
@@ -0,0 +1,225 @@
...
+///
+/// Not all hosts in this enum are available at runtime, or are even supported
+/// by the current platform. This can be checked with `is_available` or
+/// `is_supported` respectively.
+///
+/// # Getting a Host
+/// If you have a `HostId` that you would like to turn into an instance of a [`Host`],
+/// you can use the `TryFrom<HostId>` implementation of [`Host`].
+///
+/// ```ignore
```
roderickvd, line 13: We used to have a more exhaustive and compiled doctest here.

src/platform/host_id.rs
```diff
@@ -0,0 +1,225 @@
...
+/// let host = Host::try_from(HostId::Jack);
+/// ```
+///
+/// [`Host`]: crate::Host
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
+#[non_exhaustive]
```
roderickvd, line 19: We should ensure that the changelog contains entries for all breaking items. Beyond the changelog, this warrants an entry in the upgrading guide as well.

Further, we should ensure all examples and docs compile, and that docs are consistent with the new implementation and practices.

src/platform/mod.rs
```diff
@@ -63,13 +67,35 @@ pub use crate::host::custom::{Device as CustomDevice, Host as CustomHost, Stream
 macro_rules! impl_platform_host {
     ($($(#[cfg($feat: meta)])? $HostVariant:ident $($HostName:literal)? => $Host:ty),* $(,)?) => {
```
roderickvd, line 68: Is `$HostName` dead now?

src/platform/host_id.rs
```diff
@@ -0,0 +1,225 @@
...
+    ///
+    /// [`ErrorKind::HostUnavailable`]: crate::ErrorKind::HostUnavailable
+    /// [`ErrorKind::UnsupportedOperation`]: crate::ErrorKind::UnsupportedOperation
+    /// [`ErrorKind::BackendError`]: crate::ErrorKind::BackendError
+    fn try_from(value: HostId) -> Result<Self, Self::Error> {
+        crate::platform::host_from_id_impl(value)
+    }
+}
+
+pub struct AvailableHostsIter(std::slice::Iter<'static, HostId>);
```
roderickvd, line 173: I think it'd be consistent with `Devices` and `Supported*Configs` to re-export this and derive `Debug + Clone` on it.

src/platform/host_id.rs
```diff
@@ -0,0 +1,225 @@
...
+                    )*
+
+                    else { $f }
+                }
+            };
+        }
+
+        match_str_case_insensitive! {
+            s => {
+                "AAudio" => Ok(HostId::AAudio),
```
roderickvd, line 130: Mirroring `FromStr` and `Display` is normal in Rust, I know, but just to point out that the previous macro did that automatically. Not a blocker for me but worth considering to re-add.

src/platform/host_id.rs
```diff
@@ -0,0 +1,225 @@
...
+    Wasapi,
+    WebAudio,
+}
+
+impl HostId {
+    /// All hosts supported by CPAL on this platform.
+    pub const SUPPORTED_HOSTS: &[HostId] = {
+        // This is a hack to prevent rustdoc from referencing the
+        // implementation const in its output.
+        let _ = 1 + 2;
```
roderickvd, line 40: What happens without this?

roderickvd (changes requested):
Thanks for the rebase and refactor. Please find my first review attached. Happy to iterate with you to a solution that, this time, hopefully stands the test of time!

roderickvd:
I've promoted `develop` to `master`, so please rebase on `master` where we'll continue the 0.19.0 work.

roderickvd:
@cbrownsey friendly nudge to ask when you'd have the bandwidth to advance this?

cbrownsey:
I do apologize, I've recently just gotten a bit slammed with university. I should be able to get it rebased in the next few days or so.

src/platform/host_id.rs
cbrownsey: Nothing major happens, just that in the rustdoc page it shows `const SUPPORTED_HOSTS: &[HostId] = { super::SUPPORTED_HOSTS }`, and since `SUPPORTED_HOSTS` is a private implementation detail, it doesn't actually link to anything. I personally dislike how it looks, and think it makes it look like the documentation is broken, so adding a trivial statement tricks rustdoc into hiding it.

src/platform/host_id.rs
cbrownsey: `Devices` doesn't actually implement `Clone`, and none of them actually implement `Debug`. This is probably an oversight, so I'll add the impls to `AvailableHostsIter`. 
I will push back just a little on re-exporting it though. It's an iterator struct that will almost never be named, and so I don't feel is worth cluttering the root namespace with it.

examples/beep.rs
cbrownsey: Fair call. I'll re-add `host_from_id` and `available_hosts`, and mark them as deprecated.

src/platform/mod.rs
cbrownsey: I don't believe that's the case. `HostId::is_available` does end up calling `HostTrait::is_available`. `is_supported` is the one which would return true even if the backends aren't running.

src/platform/mod.rs
cbrownsey: This function is now an implementation detail for `HostId::default`. The place where this panic would now happen is on line 192 of `host_id.rs`. I can add it back there, but I don't think this function should never panic.

src/platform/mod.rs
cbrownsey: I believe so. I'll remove that now.

src/platform/host_id.rs
cbrownsey: I'm so sorry, I'm not quite sure what you're asking here. Just that the `Display` impl should be lowercased? I can add that back in.

src/platform/host_id.rs
cbrownsey: I think I was working on a slightly too fancy macro solution for this one. I'm happy to re-add it, I have no opinions either way, except that I feel this kind of use-case shouldn't have to have an allocation. But frankly that's just me being persnickety.

src/platform/host_id.rs
cbrownsey: That is true, but that was an example of matching on the cross-platform strings. This PR is to make the enum itself cross-platform, so it felt like a bad example to have. I took it out but couldn't think of anything proper to replace it with.

src/platform/host_id.rs
cbrownsey: I'll do a thorough documentation pass.

PR #1275, opened 2026-07-15
Remove `is_format_supported` calls in wasapi `build_*_stream_inner`

edwloef:
"During stream creation we’re gating on `is_format_supported` when we shouldn’t. This is inconsistent with enumeration. The resampler would handle it."

I finally got the opportunity to verify this works.


roderickvd:
Thanks. I'll push a changelog entry to your branch and merge it.

Issue #922, opened 2024-10-20
Add a (serialisable) means to uniquely identify a device


abey79:
## Motivating use case

I'm writing a GUI program where the user is offered a choice of input devices and can pick one. The program should "remember" the user choice across launch, so it must serialise some identifying data about the device such that it can be retrieved again at next launch (assuming it still exists).

As far as I understand, the only piece of identifying data available for a `Device` is its `name()`.

## Problem

Device names are not unique, at least on macOS if you have two identical (e.g.) monitors, both equipped with mics and/or speakers. I happen to be in this situation and this code returns the following result:

```rust
let devices: Vec<_> = cpal::default_host().input_devices().unwrap().collect();
for device in devices {
    println!("Input device: {}", device.name().unwrap());
}
```

Result:
```
Input device: <iphone name> Microphone
Input device: Studio Display Microphone
Input device: Studio Display Microphone
Input device: MacBook Pro Microphone
Input device: Microsoft Teams Audio
Input device: iPad audio
Input device: ZoomAudioDevice
```

## Solution

Ideally, `Device`s should have some kind of id that can be stored and checked against when trying to retrieve the same device at a later stage.

## Related

- Rejected attempt at implementing `Device::id()`: https://github.com/RustAudio/cpal/pull/537

dheijl:
I simply add the index in the list to the device name when serializing.

abey79:
I'll probably have to do something like that, though I might save the `(name, n)`, to refer to the `n`th device named `name`.

roderickvd:
Done in #1014.

Issue #551, opened 2021-03-15
FR: Add volume and mute controls to Device


micolous:
I want to be able to use this library to control the volume and mute state of input and output `Device`s, ie:

```rust
let input_device = host.default_input_device();
println("microphone mute state: {}", input_device.get_mute());
input_device.set_mute(true);

let output_device = host.default_output_device();
println!("output device volume: {} dB, {}%",
    output_device.get_volume_db(),
    output_device.get_volume_scalar() * 100);

// Set output volume to -10dB
output_device.set_volume_db(-10.);
// Set output volume to "scalar" 50% of maximum
output_device.set_volume_scalar(0.5);
```

Be-ing:
I'm not sure this should be cpal's responsibility. For ideal gain staging, volume should be set in hardware. I doubt it is possible to do this in a portable way. Of course cpal could simply apply a gain reduction to the samples as a fallback, but that would likely get confusing for the user if it does not interact with the OS or sound server's volume control.

Be-ing:
Duplicate of #436?

micolous:
I'm interested in very simple, device-level controls.  I think #436 is asking about than mixing audio sources from different applications.

Device-level mixer controls is fairly straight forward to implement, and I've had a crack at doing this for [CoreAudio (macOS, based on coreaudio-rs)](https://github.com/micolous/footswitch/blob/main/client/src/macos.rs) and [Windows APIs](https://github.com/micolous/footswitch/blob/main/client/src/windows.rs).  Doing Linux wouldn't be too difficult either, but I'd need to get a Linux box running that has audio devices. 😄 

Moving this functionality "upstream" as it were into a library like this means that I don't need to worry so much about writing `unsafe` platform-specific code, and that everyone in the Rust ecosystem can benefit from it.

GoodWork0903:
Hi everyone,
I also want to control device volume using this library.
Because we can use this library Cross Platform.
Can I use CPAL to control volume? 
Thank you

Ralith:
That sounds like a significant increase in scope. CPAL is already starved for maintenance resources at its current scope.

GoodWork0903:
I see.
I think that CPAL can extend

abique:
I think it is out of scope.
You can't do it with ASIO, and if you want it for your own buffers, just do it yourself.

xephyris:
Stumbled upon this thread while researching for this feature for a project. 

[...]

PR #1276, opened 2026-07-17
Re-export volume-related `alsa` types for better control and avoiding version discrepancies

loxoron218:
On an application I'm working in, I would like to have better control of the system volume to allow users to choose between bit-perfect (system-wide) and resampled (application only) output modes. Right now, I need to add `alsa` as a separate dependency, which leads to version discrepancies between `cpal` (which currently uses "0.11") and the latest `alsa` version ("0.12").

That's why I though re-exporting the volume-related types would avoid this issue. What do you think about it? :)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -94,7 +94,7 @@ dasp_sample = "0.11"
 [dev-dependencies]
 anyhow = "1.0"
 hound = "3.5"
-ringbuf = "0.4"
+ringbuf = "0.5.1"
 clap = { version = ">=4.6, <5", features = ["derive"] }
 
 # `windows` and `windows-core` must be the same minor version.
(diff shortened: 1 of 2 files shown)


roderickvd:
Thanks for opening this, and yeah, WASAPI users have this same version-lockstep pain with `windows`/`windows-core`, so if we were to go this route, I'd want a general fix rather than one crate at a time.

However, I don't think re-exporting either the specific types or the whole `alsa` crate gets us there where we want. 
Cargo already dedupes compatible semver ranges, and interface-free re-export becomes permanent public API the moment anyone stores that type in their own code. I don't want `alsa`/`windows` version bumps become breaking changes for cpal itself.

Instead, we went through this in #551 and landed on volume control living outside cpal, as an extension crate keyed off a stable device ID (#922). You can use that to access the mixer. If we ever add a device-level escape hatch, I'd rather it extend the `into_inner()`/`as_inner()` pattern from #660.

So I am going to close this for now, but am genuinely thankful for your first contribution!

PR #1283, opened 2026-07-21
fix(pulseaudio,pipewire): build on targets without 64-bit atomics

### LINKED ISSUE #1277: cpal fails to build for 32bit systems
I'm trying to build cpal on a bit of an exotic platform (`mipsel-unknown-linux-musl` - a Hiby R1 audio player) and unfortunately parts of the code seem to rely on 64bit data structures. This results in:

```
error[E0432]: unresolved import `std::sync::atomic::AtomicU64`
 --> /cargo/registry/src/index.crates.io-1949cf8c6b5b557f/cpal-0.18.1/src/host/pulseaudio/stream.rs:3:30
  |
3 |         atomic::{AtomicBool, AtomicU64, Ordering},
  |                              ^^^^^^^^^ no `AtomicU64` in `sync::atomic`
  |
help: a similar name exists in the module
  |
3 -         atomic::{AtomicBool, AtomicU64, Ordering},
3 +         atomic::{AtomicBool, AtomicU32, Ordering},
  |

```

I'm not sure this can be solved easily, but it may be worth a closer look. Should happen on every 32bit target. Let me know if you need more info.

Best,

sandreas

roderickvd:
PulseAudio's latency tracking stores two microsecond counters in `AtomicU64`, which doesn't exist in `std` on exotic or legacy targets without native 64-bit atomics. This PR adds fallback to `portable-atomic` on such targets.

PipeWire hid the same error through its `last_quantum` field, but we can narrow that to `u32` without any problem or pulling in another dependency.

Closes #1277

### DIFF
--- a/.github/workflows/platforms.yml
+++ b/.github/workflows/platforms.yml
@@ -34,10 +34,10 @@ env:
   MSRV_WASM: "1.85"
   MSRV_WINDOWS: "1.85"
 
-  PACKAGES_LINUX: libasound2-dev libjack-jackd2-dev libdbus-1-dev libpipewire-0.3-dev
+  PACKAGES_LINUX: libasound2-dev libdbus-1-dev libjack-jackd2-dev libpipewire-0.3-dev
 
   # System packages required by the docs.rs feature set and available on docs.rs build environment.
-  PACKAGES_DOCS_RS: libasound2-dev libjack-jackd2-dev libdbus-1-dev
+  PACKAGES_DOCS_RS: libasound2-dev libdbus-1-dev libjack-jackd2-dev
 
   ANDROID_COMPILE_SDK: "30"
   ANDROID_BUILD_TOOLS: "30.0.3"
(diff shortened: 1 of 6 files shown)


roderickvd:
@sandreas

sandreas:
Awesome thank you, try to provide feedback tomorrow.

sandreas:
@roderickvd 

Sorry for the delay, I'm currently fighting with my toolchain, so it may take some time. I try to keep you up to date.

sandreas:
@roderickvd 
Awesome, this does the trick. Thank you very much. 
I checked this branch works and re-checked, that master fails to be sure, so it definitely compiles now.

Would it be useful to add a 32bit target to the workflows ensuring that this cannot happen again?

Maybe I have a little extra to think about (not this part of the code - this is an external APIs ID known to be 32bit integer, but maybe other parts of the code storing the amount of samples):

Let's say we have an audio book with 28 hours duration and a sampling rate of 44100Hz - that would make 28×3600×44.100=**4.445.280.000** samples right?
Would these still fit in a 32bit uint? 
- Samples: 4.445.280.000
- Max 32bit uint: 4.294.967.295

This was a problem on old iPod Nanos (although they had used int32, so 28 hours with 22050Hz was too much, maybe we can prevent this from happening in cpal, before it appears. Just a thought ;-)

roderickvd:
> @roderickvd Awesome, this does the trick. Thank you very much. I checked this branch works and re-checked, that master fails to be sure, so it definitely compiles now.

Great, thanks for the feedback!

> Would it be useful to add a 32bit target to the workflows ensuring that this cannot happen again?

Yeah, I've been able to add a MIPS target in [40b15a5](https://github.com/RustAudio/cpal/pull/1283/commits/40b15a5f69c0d82b6046c55fcd4a85f84efc8ea3) through a Bookworm cross-build. Let's see how long that lasts. 

> Let's say we have an audio book with 28 hours duration and a sampling rate of 44100Hz - that would make 28×3600×44.100=**4.445.280.000** samples right? Would these still fit in a 32bit uint?

To answer that question we must discern between position counters (and in which unit) and buffer sizes.                                                                      

PipeWire's `last_quantum` is a buffer size (frames per callback), safe as `u32`.

PulseAudio's latency tracking is a position counter in microseconds, which would overflow after ~71 minutes as `u32`, so that one stays a real 64-bit atomic.

So as far as I can see, this is all accounted for now.

sandreas:
> So as far as I can see, this is all accounted for now.

Great work. Thank you very much!

PR #1284, opened 2026-07-21
fix(jack): make DeviceId stable across restarts

### LINKED ISSUE #1279: What to do with JACK device names?
Currently they contain the PID (e.g. `cpal_client_82935_in`), but that leads to unstable serialization of `DeviceId` across application restarts, and therefore breaks configs that save device choices.

roderickvd:
`Device::id()` and `description()` derived from the JACK client name, which embeds the process ID. That broke `DeviceId`'s documented restart-stability contract. Use direction (input/output) instead.

Fixes #1279


roderickvd:
@edwloef

edwloef:
Looks good I think :)

PR #1285, opened 2026-07-22
fix(alsa): timespec segfault with 64-bit time_t on 32-bit systems

roderickvd:
This stacks with https://github.com/diwic/alsa-rs/pull/158.

@d3d9


roderickvd:
It fails to compile because you need to patch `alsa-rs` first.

roderickvd:
Despite the ongoing discussion over at https://github.com/diwic/alsa-rs/pull/158, this patch seems correct, so I plan on merging it unless someone feels differently.

roderickvd:
Changing base to `develop` because this requires `alsa-rs` 0.12(.1). cpal 0.18 is bound to `alsa-rs` 0.11 for SemVer-reasons.

PR #1289, opened 2026-07-26
ios: recover the stream after an audio session interruption

benface:
## The problem

On iOS an audio session interruption — a call, an alarm, Siri — stops the audio unit through CoreAudio without going through cpal. `StreamInner::playing` is left reading `true` for a unit that is no longer running, so `StreamTrait::play` short-circuits:

```rust
if !stream.playing {
    stream.audio_unit.start()?;
    stream.playing = true;
}
```

An interrupted stream therefore can never be resumed for the lifetime of the process: `play()` returns `Ok(())` and produces silence. `AVAudioSessionInterruptionNotification` is not among the notifications `SessionEventManager` watches, so nothing is reported either and the caller has no way to discover why. Reactivating the session with `setActive:YES` does not help on its own — the session comes back and the unit stays stopped.

## The change

Observe `AVAudioSessionInterruptionNotification` alongside the route-change and media-services notifications already handled. On the beginning of an interruption, stop the stream through the normal path — which brings the flag back in step so a later `play` starts the unit again — and report a `StreamInvalidated` error.

Resuming stays the caller's decision, since Apple's guidance is that the app reads `AVAudioSessionInterruptionOptionShouldResume`, so the stream is left stopped rather than restarted here.

`SessionEventManager` is built before the `Stream`, so it takes a slot that `Stream::new` fills with a `Weak<Mutex<StreamInner>>`. The observer blocks already gate on the readiness latch, which is released later still, so the slot is always populated by the time one can act — reusing the existing ordering invariant rather than adding a second.

If you would rather not touch the state layout, a reporting-only variant is a much smaller diff, though callers would still be unable to recover.

## Verification

Reproduced and confirmed on an iPhone against a shipping app: the stop-then-start this performs is what restores audio after an interruption.

- `cargo check --target aarch64-apple-ios` clean
- `cargo clippy --target aarch64-apple-ios --lib` clean
- `cargo fmt` clean
- The three `--all-targets` test errors on this target (`default_output_device`, `OutputCallbackInfo`, `SampleRate`) are pre-existing on `master` and unrelated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **AudioWorklet**: Fix stale audio output when a data callback wrote a partial buffer.
 - **CoreAudio**: Default-output streams now report xrun status.
 - **CoreAudio**: Fix stale audio output when a data callback wrote a partial buffer.
+- **iOS**: Streams stopped by an audio session interruption now resume when it ends, instead of staying silent until rebuilt.
 - **JACK**: Streams with more channels than physical system ports no longer fail to build.
 - **JACK**: Channel enumeration is no longer capped at the physical system port count.
 - **JACK**: `Device::id` no longer embeds the process ID, so `DeviceId` is now stable across application restarts.
--- a/src/host/coreaudio/ios/mod.rs
+++ b/src/host/coreaudio/ios/mod.rs
@@ -190,11 +190,8 @@ impl DeviceTrait for Device {
         let latency_frames = Arc::new(AtomicUsize::new(input_latency_frames()));
 
         let error_callback: ErrorCallbackArc = Arc::new(Mutex::new(error_callback));
-        let session_manager = SessionEventManager::new(
-            error_callback.clone(),
-            Latch::new(),
-            Some((latency_frames.clone(), true)),
-        );
+        let session_error_callback = error_callback.clone();
+        let session_latency_frames = latency_frames.clone();
 
         // Set up input callback
         setup_input_callback(
@@ -208,13 +205,21 @@ impl DeviceTrait for Device {
             },
         )?;
 
-        let stream = Stream::new(
-            StreamInner {
-                playing: false,
-                audio_unit,
-            },
-            session_manager,
+        let inner = Arc::new(Mutex::new(StreamInner {
+            playing: false,
+            interrupted: false,
+            audio_unit,
+        }));
+        let session_manager = SessionEventManager::new(
+            session_error_callback,
+            Latch::new(),
+            Some((session_latency_frames, true)),
+            Arc::downgrade(&inner),
         );
+        let stream = Stream {
+            inner,
+            session_manager,
+        };
         stream.signal_ready();
         Ok(stream)
     }
@@ -245,11 +250,8 @@ impl DeviceTrait for Device {
         let latency_frames = Arc::new(AtomicUsize::new(output_latency_frames()));
 
         let error_callback: ErrorCallbackArc = Arc::new(Mutex::new(error_callback));
-        let session_manager = SessionEventManager::new(
-            error_callback.clone(),
-            Latch::new(),
-            Some((latency_frames.clone(), false)),
-        );
+        let session_error_callback = error_callback.clone();
+        let session_latency_frames = latency_frames.clone();
 
         // Set up output callback
         setup_output_callback(
@@ -263,31 +265,32 @@ impl DeviceTrait for Device {
             },
         )?;
 
-        let stream = Stream::new(
-            StreamInner {
-                playing: false,
-                audio_unit,
-            },
-            session_manager,
+        let inner = Arc::new(Mutex::new(StreamInner {
+            playing: false,
+            interrupted: false,
+            audio_unit,
+        }));
+        let session_manager = SessionEventManager::new(
+            session_error_callback,
+            Latch::new(),
+            Some((session_latency_frames, false)),
+            Arc::downgrade(&inner),
         );
+        let stream = Stream {
+            inner,
+            session_manager,
+        };
         stream.signal_ready();
         Ok(stream)
     }
 }
 
 pub struct Stream {
-    inner: Mutex<StreamInner>,
+    inner: Arc<Mutex<StreamInner>>,
     session_manager: SessionEventManager,
 }
 
 impl Stream {
-    fn new(inner: StreamInner, session_manager: SessionEventManager) -> Self {
-        Self {
-            inner: Mutex::new(inner),
-            session_manager,
-        }
-    }
-
     fn signal_ready(&self) {
         self.session_manager.signal_ready();
     }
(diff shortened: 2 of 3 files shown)


benface:
Confirmed end to end on an iPhone. The shipping app was pointed at this branch with its own `pause()`-before-`play()` workaround removed, so recovery had to come from cpal alone: after dismissing an alarm mid-session, a plain `play()` resumes audio. Without the patch the same call is a silent no-op.

src/host/coreaudio/ios/session_event_manager.rs
```diff
@@ -4,20 +4,43 @@ use std::{
...
+    let Some(value) = user_info.objectForKey(&**key) else {
+        return false;
+    };
+    let Ok(number) = value.downcast::<NSNumber>() else {
+        return false;
+    };
+    AVAudioSessionInterruptionType(number.unsignedIntegerValue())
+        == AVAudioSessionInterruptionType::Began
+}
 use crate::{
```
roderickvd, line 23: Style: this `use crate::{...}` block should live with the rest of the imports at the top, not after `is_interruption_start`.

src/host/coreaudio/ios/session_event_manager.rs
```diff
@@ -77,6 +101,43 @@ impl SessionEventManager {
         let nc = NSNotificationCenter::defaultCenter();
         let mut observers = Vec::new();
         let waiter = latch.waiter();
+        let stream: StreamSlot = Arc::new(OnceLock::new());
+
+        // An interruption — a call, an alarm, Siri — stops the audio unit through CoreAudio
```
roderickvd: This reads unedited (same goes for `mark_stopped_by_system`'s comment in `mod.rs`). Trim to the why in one line, e.g. "the OS already stopped the unit, resync `playing` so a later `play()` actually starts it again."

src/host/coreaudio/ios/session_event_manager.rs
```diff
@@ -77,6 +101,43 @@ impl SessionEventManager {
         let nc = NSNotificationCenter::defaultCenter();
         let mut observers = Vec::new();
         let waiter = latch.waiter();
+        let stream: StreamSlot = Arc::new(OnceLock::new());
```
roderickvd: Do we need the two-step `OnceLock`/`attach_stream` dance? `SessionEventManager` only gets built before `StreamInner` because of the order in `build_*_stream_raw`, nothing requires that. Build `Arc<Mutex<StreamInner>>` first and pass `Arc::downgrade(&inner)` straight into `SessionEventManager::new()`, then `StreamSlot`, `OnceLock`, and `attach_stream` can all go.

src/host/coreaudio/ios/session_event_manager.rs
```diff
@@ -77,6 +101,43 @@ impl SessionEventManager {
...
+        // the caller has no way to tell. Stopping through the normal path resyncs the two, so a
+        // later `play` starts the unit again. Whether to resume is the caller's decision — the
+        // notification's `ShouldResume` option is theirs to read — so this reports the event and
+        // leaves the stream stopped.
+        {
+            let cb = error_callback.clone();
+            let w = waiter.clone();
+            let slot = stream.clone();
+            let block = RcBlock::new(move |notif: NonNull<NSNotification>| {
+                if !w.is_released() || !unsafe { is_interruption_start(notif.as_ref()) } {
```
roderickvd: This only handles `Began`, `Ended` is silently dropped, that's actually the useful half. Read `AVAudioSessionInterruptionOptionKey`/`ShouldResume` here: if set, reactivate the session and start the unit again the same way `play()` does, right from this closure using the same `Weak` already in scope. If not set, leave it stopped, that's just a normal `pause()`'d stream. Without that, a cpal user has to observe `AVAudioSessionInterruptionNotification` themselves just to get their stream back, which defeats the point of cpal hiding the platform bits.

src/host/coreaudio/ios/session_event_manager.rs
```diff
@@ -77,6 +101,43 @@ impl SessionEventManager {
...
+                    return;
+                }
+                if let Some(inner) = slot.get().and_then(Weak::upgrade) {
+                    if let Ok(mut inner) = inner.lock() {
+                        inner.mark_stopped_by_system();
+                    }
+                }
+                emit_error(
```
roderickvd: Let's drop this `emit_error` call. The data callback not firing is already the signal that playback stopped, we don't need a bespoke error on top of it. `StreamInvalidated` would be the wrong pick here anyway, it's documented as "must be rebuilt" which isn't true.

roderickvd (changes requested):
Thanks for tracking this down and verifying on real hardware, that's a real bug.

Per `CONTRIBUTING.md` I don't mind AI usage, but you do need to hold it to the same bar as hand-written code, and some of the comments here read unedited, see inline. Couple of points on the interruption handling itself too.

And if you would also add a short entry to `CHANGELOG.md`?

benface:
Thanks @roderickvd, you were right about `Ended`. Handling only `Began` meant every caller still had to observe the notification themselves.

Updated:

- `Ended` reads `AVAudioSessionInterruptionOptionKey` and, when `ShouldResume` is set, reactivates the session and starts the unit again from the same closure. Only streams the interruption stopped resume, so one the caller had paused stays paused.
- Dropped the `OnceLock` / `attach_stream` two-step. Both builders create the `Arc<Mutex<StreamInner>>` first and pass `Arc::downgrade` into `SessionEventManager::new`.
- Dropped the `emit_error`, and the `StreamInvalidated` kind with it.
- Moved the `use crate::{...}` block up with the other imports, trimmed the two comments, added a CHANGELOG entry.

Also folded the two different ways of reading `userInfo` into one helper, and inlined `Stream::new` now that it does not do anything.

Confirmed on device again: app pointed at this branch with both its own workaround and its interruption observer removed, so recovery had to come from cpal alone. Music comes back after dismissing an alarm.

src/host/coreaudio/ios/session_event_manager.rs
benface: No, it was only needed because the manager was built first. Both builders now create the stream first and pass `Arc::downgrade` straight in, so `StreamSlot`, the `OnceLock` and `attach_stream` are all gone.

src/host/coreaudio/ios/mod.rs
```diff
@@ -341,9 +344,33 @@ impl StreamTrait for Stream {
 
 struct StreamInner {
     playing: bool,
+    /// Set while an interruption is what stopped the unit, so only those resume on its end.
```
roderickvd, line 347: Remove comments where the code is self-explanatory.

src/host/coreaudio/ios/mod.rs
roderickvd, line 325: `pause` now needs to set `stream.interrupted = false`. Otherwise, calling `pause` during an interruption would have the end-of-interruption event restart the stream when you don't want to.

src/host/coreaudio/ios/session_event_manager.rs
```diff
@@ -1,39 +1,54 @@
...
+fn with_stream(stream: &Weak<Mutex<StreamInner>>, f: impl FnOnce(&mut StreamInner)) {
+    if let Some(inner) = stream.upgrade() {
+        if let Ok(mut inner) = inner.lock() {
+            f(&mut inner);
+        }
+    }
+}
+
+/// Reads the number stored under `key` in a notification's `userInfo`.
+unsafe fn user_info_number(notification: &NSNotification, key: Option<&NSString>) -> Option<usize> {
```
roderickvd, line 38: I think this doesn't need `unsafe` any longer?

CHANGELOG.md
```diff
@@ -30,15 +30,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **AudioWorklet**: Fix stale audio output when a data callback wrote a partial buffer.
 - **CoreAudio**: Default-output streams now report xrun status.
 - **CoreAudio**: Fix stale audio output when a data callback wrote a partial buffer.
+- **iOS**: Streams stopped by an audio session interruption now resume when it ends, instead of staying silent until rebuilt.
```
roderickvd, line 33: Could this be shorter, like: "Streams now resume automatically when an audio session interruption ends."

src/host/coreaudio/ios/mod.rs
```diff
@@ -341,9 +344,33 @@ impl StreamTrait for Stream {
...
+    interrupted: bool,
```
roderickvd, line 348: Instead of an `interrupted` field next to `playing`, we should probably use an enum like:

```rust
enum PlaybackState {
    Stopped,
    Playing,
    Interrupted,
}
```

This will also prevent race conditions when one flag is written, the other not yet, and one of the closures is called.

roderickvd (changes requested):
Thanks for processing the previous round. Having worked through the various states, I think we need some further changes to prevent race conditions.

roderickvd:
Superseded by #1295. I want to have it included in 0.18.2 that I want to release momentarily. I'll keep you as co-author @benface, thanks for the initial report and work.

PR #1291, opened 2026-07-28
fix(alsa): use PCM state for suspend handling

### LINKED ISSUE #1290: error[E0425]: cannot find value `ESTRPIPE` in crate `libc`
Hi, I'm trying to compile [rustconn](https://github.com/totoshko88/RustConn) (v. 0.19.5), which depends on cpal crate, on FreeBSD-14.4, but it fails with the following error:

```
error[E0425]: cannot find value `ESTRPIPE` in crate `libc`
    --> /usr/local/poudriere/ports/default/net/rustconn/work/RustConn-0.19.5/cargo-crates/cpal-0.18.1/src/host/alsa/mod.rs:1125:42
     |
1125 |         Err(err) if err.errno() == libc::ESTRPIPE => return try_resume(&stream.handle),
     |                                          ^^^^^^^^
     |
    ::: /usr/local/poudriere/ports/default/net/rustconn/work/RustConn-0.19.5/cargo-crates/libc-0.2.189/src/unix/bsd/freebsdlike/mod.rs:630:1
     |
 630 | pub const ESPIPE: c_int = 29;
     | ----------------------- similarly named constant `ESPIPE` defined here
     |
help: a constant with a similar name exists
     |
1125 -         Err(err) if err.errno() == libc::ESTRPIPE => return try_resume(&stream.handle),
1125 +         Err(err) if err.errno() == libc::ESPIPE => return try_resume(&stream.handle),
     |
```

The same happens while compiling other programs such as [noctavox](https://github.com/Jaxx497/NoctaVox).

Do you have any suggestion?

Thanks.

roderickvd:
Compare state instead of `errno() == libc::ESTRPIPE` which is only a Linux thing.

Fixes #1290

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - Timestamps now stay monotonic across device and graph changes.
 - **ALSA**: A nonzero but sub-millisecond stream timeout is no longer treated as a non-blocking poll.
+- **ALSA**: Fix compilation on FreeBSD, DragonFly, and NetBSD.
 - **AudioWorklet**: Fix `Stream` operations to work when called from any thread.
 - **AudioWorklet**: Fix stale audio output when a data callback wrote a partial buffer.
 - **CoreAudio**: Default-output streams now report xrun status.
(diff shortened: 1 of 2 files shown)


roderickvd:
Merging this right away; please reopen #1290 if it still fails.

PR #1297, opened 2026-07-31
fix(asio): share duplex buffer state across Device handles for the same driver

### LINKED ISSUE #1287: Bug: ASIO modified feedback example no mic input
I was trying to run a modified version of the feedback example specifically for ASIO, the regularl feedback example works with no issues, but when adding ASIO to the equation it is ran to completion but there's no audio playing, either the mic or the headphones, but based on another project I have tried that uses cpal I think the issue stems from the mic.

OS: Windows 11
Audio Interface: UMC22
Branch: master
ASIO SDK version: ASIO-SDK_2.3.4
ASIO Driver (whatever u call this): ASIO4ALL

<details>

<summary>Script</summary>

```rs
use ringbuf::{
    traits::{Consumer, Producer, Split},
    HeapRb,
};

fn main() {

    let input_device = host
        .default_input_device()
        .expect("failed to find input device");
    let output_device = host
        .default_output_device()
        .expect("failed to find output device");

    let config = input_device.default_input_config().unwrap().config();

    let latency_frames = 0.1 * config.sample_rate as f32;
    let latency_samples = latency_frames as usize * config.channels as usize;

    let ring = HeapRb::<i32>::new(latency_samples * 2);
    let (mut producer, mut consumer) = ring.split();

    for _ in 0..latency_samples {
        producer.try_push(0).unwrap();
    }

    let input_data_fn = move |data: &[i32], _: &_| {
        if producer.push_slice(data) < data.len() {
            eprintln!("output stream fell behind");
        }
    };

    let output_data_fn = move |data: &mut [i32], _: &_| {
        let read = consumer.pop_slice(data);
        if read < data.len() {
            data[read..].fill(0);
            eprintln!("input stream fell behind");
        }
    };

    let input_stream = input_device
        .build_input_stream(config, input_data_fn, |err| eprintln!("{err}"), None)
        .unwrap();
    let output_stream = output_device
        .build_output_stream(config, output_data_fn, |err| eprintln!("{err}"), None)
        .unwrap();

    input_stream.play().unwrap();
    output_stream.play().unwrap();

    std::thread::sleep(std::time::Duration::from_secs(10));

    drop(input_stream);
    drop(output_stream);
}
```

</details>

roderickvd:
`default_input_device()` and `default_output_device()` each enumerate a fresh `Device` with its own empty `AsioStreams`, even when they resolve to the same loaded driver. This broke the reuse path in `get_or_create_input_stream` / `get_or_create_output_stream`: building both directions triggered a second, destructive `ASIOCreateBuffers` call that silently discarded the first direction's buffers instead of combining them into one call as designed.

This PR moves the `AsioStreams` bookkeeping onto `DriverInner`, alongside `DriverState`, so every `Driver` handle for the same loaded driver shares it.

Fixes #1287

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Timestamps now stay monotonic across device and graph changes.
 - **ALSA**: A nonzero but sub-millisecond stream timeout is no longer treated as a non-blocking poll.
 - **ALSA**: Fix compilation on FreeBSD, DragonFly, and NetBSD.
+- **ASIO**: Fix duplex streams silently dropping audio when built from separate `Device` handles.
 - **AudioWorklet**: Fix `Stream` operations to work when called from any thread.
 - **AudioWorklet**: Fix stale audio output when a data callback wrote a partial buffer.
 - **CoreAudio**: Default-output streams now report xrun status.
--- a/asio-sys/src/bindings/mod.rs
+++ b/asio-sys/src/bindings/mod.rs
@@ -133,6 +133,7 @@ struct BufferCallback(Box<dyn FnMut(&CallbackInfo) + Send>);
 /// There is only ever max one input and one output.
 ///
 /// Only one is required.
+#[derive(Debug)]
 pub struct AsioStreams {
     pub input: Option<AsioStream>,
     pub output: Option<AsioStream>,
@@ -141,6 +142,7 @@ pub struct AsioStreams {
 /// A stream to ASIO.
 ///
 /// Contains the buffers.
+#[derive(Debug)]
 pub struct AsioStream {
     /// A Double buffer per channel
     pub buffer_infos: Vec<AsioBufferInfo>,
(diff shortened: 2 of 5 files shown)


1rhino2 (commented):
code review (no ASIO hardware here so this is static only). cargo check --all-targets green on the PR branch under the default windows host.

the root cause story matches the code: get_or_create_input_stream / get_or_create_output_stream only combine via prepare_*_stream when the other direction already lives in the same AsioStreams mutex. on master, default_input_device() and default_output_device() each mint a Device with its own empty Arc, so the second build can't see the first and ASIOCreateBuffers runs again. moving streams onto DriverInner so every Driver/Device for that loaded driver shares one Arc looks like the right fix.

feedback.rs --asio flag + sample-format dispatch is a nice independent improvement for repros.

one note from #1287: after clarifying build order, the reporter said they saw no perceived difference between master and this branch. that doesn't kill the change for me (order sensitivity on master is itself a smell for the discarded-buffers failure mode), but it'd be good to get one more clean A/B with the RMS harness (input-first and output-first) logged on both tips before merge. i don't have an ASIO interface to run that myself yet.

happy to help chase any leftover edge cases if someone can share a failing recipe that still repros on this branch.

roderickvd:
Apparently this doesn't fix it yet. Debugging and discussion continuing at https://github.com/RustAudio/cpal/issues/1287.

roderickvd:
Done now! Merging when the CI passes.

PR #1300, opened 2026-08-07
fix(pipewire): fix 24bit sample format mapping

kpqi5858:
Use correct pipewire AudioFormat for cpal's 4-byte I24, U24 representation.


roderickvd:
Looks great. changelog and all! I'll find out why the CI is failing on MIPS and we'll merge. Thanks!

PR #1303, opened 2026-08-10
Feature: Adding support for retrieval of ASIO channel names - 0.3.1

SpruceCloud:


roderickvd:
Thank you for your contribution. We also had https://github.com/RustAudio/cpal/pull/1254 queued up for `develop` (anyway this will be a v0.19 feature and not a v0.18 one). Before I look further, would you look to that one and see what's the best way forward?

PR #1305, opened 2026-08-11
fix(wasapi): left-justify samples in a container wider than they are

ErwanLegrand:
A WAVEFORMATEXTENSIBLE describes a sample with two numbers: wBitsPerSample, the container, and wValidBitsPerSample, how much of it the sample fills. When they differ the valid bits sit at the *top* of the container -- "the valid bits (the actual PCM data) are left-aligned within the container", as ksmedia.h's WAVEFORMATEXTENSIBLE reference puts it. CPAL's I24 is the other way round: a dasp_sample::I24 is an i32 holding -(1 << 23)..=(1 << 23) - 1, at the bottom of its four-byte container.

So the backend handed a 24-in-32 device every sample 2^8 too small and read every sample back 2^8 too large. It reaches both share modes: supported_formats reports I24 as supported, and the two fields go to the engine's converter in shared mode and to the driver in exclusive mode alike.

Measured on a PreSonus AudioBox 22VSL (USB, Windows 11) at 24-in-32 / 48 kHz stereo, in both share modes and both directions, converting I24 to F32 on input and F32 to I24 on output. Output was 48 dB too quiet -- clean, just far below level -- and input pinned to full scale by anything above near-silence, moving only when the source was nearly silent. Instrumented on the capture side, a shared-mode I24 stream returned samples 2^8 too large, within 0.6 dB of an I32 reference on the same scale.

Render shifts up in place, between the data callback and ReleaseBuffer: that buffer is the backend's to write until it is handed back. Capture cannot do the same, since GetBuffer lends WASAPI's own packet and lends it to be read, so the samples are shifted down into a staging buffer on their way to the callback -- sized once at stream build, so the callback still allocates nothing. It is a Vec<i32> rather than a Vec<u8> because that buffer reaches the callback as a Data, whose as_slice casts to the sample type, and a Vec<u8> guarantees no alignment.

The shift is wBitsPerSample - wValidBitsPerSample read off the format handed to Initialize, not a test for I24, so every format whose container is exactly full -- I16, I32, F32 -- lands on zero by the same arithmetic and keeps its existing zero-copy path byte for byte. A container that is padded but not four bytes wide is refused rather than passed through unshifted, which would be silently wrong by the width of the padding.

The integer half is a platform-neutral module so it can be tested off Windows: the round trip, the -2^23 and 2^23 - 1 boundaries, the sign carried down, padding bits discarded, the shift actually coming from the format, a trailing partial container left alone, and 16- and 32-bit formats passed through untouched.

This was made with the assistance of AI coding agents.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **WASAPI**: Reported buffer sizes are no longer off by one frame.
 - **WASAPI**: `Stream::drop`, `play`, and `pause` no longer panic when the device is lost.
 - **WASAPI**: Output streams no longer reject formats that the built-in resampler can convert.
+- **WASAPI**: 24-bit streams are no longer 48 dB out. A container wider than the sample it carries
+  holds that sample left-justified, so `SampleFormat::I24` samples are now shifted up on their way
+  to the device and back down on their way in. Formats whose container is exactly full — `I16`,
+  `I32`, `F32` — are untouched.
 - **WebAudio**: Fix stale audio output when a data callback wrote a partial buffer.
 - **WebAudio**: Fix unsound `Send + Sync` on `Stream` when compiled with `+atomics`.
 - **WebAudio**: Fix `Host::is_available()` always returning `true`, even in non-window contexts.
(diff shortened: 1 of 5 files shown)


roderickvd:
Thanks for the contribution and LLM disclosure. Per [`CONTRIBUTING.md`](https://github.com/RustAudio/cpal/blob/master/CONTRIBUTING.md) that's OK as long as the code and style are human-vetted. Here are some things that don't yet hold up as much as I'd like to see:

- Could it be shorter?

- `src/host/container_align.rs` doesn't touch any Windows types and I wouldn't see which other host would. Please isolate it to the WASAPI host.

- The new CHANGELOG entry is four lines. Please trim it.

- `container_shift()` re-reads `wValidBitsPerSample` but the caller already has `valid_bits` a few lines up. Can that re-reading be prevented?

roderickvd:
Superseded by #1309 which is showing the way to a much simpler and preferable solution to what the LLM created here.

PR #1306, opened 2026-08-11
doc: record UnsupportedConfig on the supported-configs queries

ErwanLegrand:
`DeviceTrait::supported_input_configs` and `supported_output_configs` list `DeviceNotAvailable` and `UnsupportedOperation`, but the WASAPI backend also answers `UnsupportedConfig` from `supported_formats`, which both of them call: once when `IsFormatSupported` cannot vouch for the mix format, and once when that format maps to no `SampleFormat` cpal has.

Documentation only; no behaviour change.


roderickvd:
Merged, thanks!

PR #1309, opened 2026-08-14
fix(wasapi): shift i24 to/from MSB

LastExceed:
supersedes #1305

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -85,7 +85,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **WASAPI**: Capture streams no longer report a spurious xrun on the first buffer after starting.
 - **WASAPI**: Output streams no longer reject formats that the built-in resampler can convert.
 - **WASAPI**: Device enumeration no longer panics if the COM enumerator fails to initialize.
-- **WASAPI**: Fix 24bit samples not getting shifted from/to MSB.
+- **WASAPI**: Fix 24-bit samples not getting shifted from/to MSB.
 - **WebAudio**: Fix stale audio output when a data callback wrote a partial buffer.
 - **WebAudio**: Fix unsound `Send + Sync` on `Stream` when compiled with `+atomics`.
 - **WebAudio**: Fix `Host::is_available()` always returning `true`, even in non-window contexts.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -85,6 +85,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **WASAPI**: Capture streams no longer report a spurious xrun on the first buffer after starting.
 - **WASAPI**: Output streams no longer reject formats that the built-in resampler can convert.
 - **WASAPI**: Device enumeration no longer panics if the COM enumerator fails to initialize.
+- **WASAPI**: Fix 24bit samples not getting shifted from/to MSB.
 - **WebAudio**: Fix stale audio output when a data callback wrote a partial buffer.
 - **WebAudio**: Fix unsound `Send + Sync` on `Stream` when compiled with `+atomics`.
 - **WebAudio**: Fix `Host::is_available()` always returning `true`, even in non-window contexts.
--- a/src/host/wasapi/stream.rs
+++ b/src/host/wasapi/stream.rs
@@ -1,8 +1,7 @@
-use std::slice;
 use std::{
     mem,
     ops::ControlFlow,
-    ptr,
+    ptr, slice,
     sync::{
         Arc,
         atomic::{AtomicBool, AtomicU64, Ordering},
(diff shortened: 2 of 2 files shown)


src/host/wasapi/stream.rs
```diff
@@ -1,3 +1,4 @@
+use std::slice;
```
roderickvd: Style: please fold this into the imports below.

src/host/wasapi/stream.rs
```diff
@@ -921,6 +953,19 @@ fn process_output(
...
             );
+
+            if stream.sample_format == SampleFormat::I24 {
+                // WASAPI stores i24 in the upper bits
+                #[expect(
+                    clippy::cast_ptr_alignment,
+                    reason = "WASAPI guarantees the buffer to be aligned to a frame boundary"
+                )]
+                let buffer_slice_i32 =
+                    slice::from_raw_parts_mut(buffer.cast::<i32>(), len / size_of::<i32>());
```
roderickvd: Doesn't this divide `len` twice (incorrectly)? Compare to `process_input` calculating `byte_count / size_of::<i32>()` directly when slicing.

CHANGELOG.md
```diff
@@ -84,6 +84,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **WASAPI**: `Stream::drop`, `play`, and `pause` no longer panic when the device is lost.
 - **WASAPI**: Capture streams no longer report a spurious xrun on the first buffer after starting.
 - **WASAPI**: Output streams no longer reject formats that the built-in resampler can convert.
+- **WASAPI**: Fix 24bit samples not getting shifted from/to MSB.
```
roderickvd: Style: "24-bit" is what we normally use.

src/host/wasapi/stream.rs
```diff
@@ -815,6 +830,7 @@ fn process_input(
     stream: &StreamInner,
     capture_client: Audio::IAudioCaptureClient,
     data_callback: &mut dyn FnMut(&Data, &CallbackInfo),
+    scratch_buffer: &mut Vec<i32>,
```
roderickvd: Should be able to take a `&mut [i32]`, right?

src/host/wasapi/stream.rs
```diff
@@ -665,6 +666,15 @@ fn run_input(
...
     }
 
+    let stream = &run_ctxt.stream;
+    let scratch_len = if stream.sample_format == SampleFormat::I24 {
+        stream.max_frames_in_buffer as usize * stream.bytes_per_frame as usize / size_of::<i32>()
+    } else {
+        // The scratch buffer won't be used in this case.
+        0 // Vec::with_capacity(0) does not allocate.
+    };
+    let mut scratch_buffer = Vec::with_capacity(scratch_len);
```
roderickvd: This doesn't ever need to grow, does it? In which case we could `Box` it:

```rust
let mut scratch_buffer: Box<[i32]> = vec![0; scratch_len].into_boxed_slice();

let dst = &mut scratch_buffer[..sample_count];
dst.copy_from_slice(source_data);
for sample in dst.iter_mut() { *sample >>= 8; }
```

roderickvd (changes requested):
😅 last steps now.

src/host/wasapi/stream.rs
LastExceed: done

src/host/wasapi/stream.rs
LastExceed: WASAPI makes no guarantee about this. `max_frames_in_buffer` is just the *available* buffer space at creation time, which is not necessarily the *maximum* that `IAudioRenderClient.GetBuffer()` will ever return

src/host/wasapi/stream.rs
LastExceed: No, due to the use of `Vec::clear()`

src/host/wasapi/stream.rs
LastExceed: Oops

CHANGELOG.md
LastExceed: Fixed

src/host/wasapi/stream.rs
roderickvd: For capture, from [GetBufferSize:](https://learn.microsoft.com/en-us/windows/win32/api/audioclient/nf-audioclient-iaudioclient-getbuffersize):

> For capture clients, the buffer length determines the maximum amount of capture data that the audio engine can read from the endpoint buffer during a single processing pass. 

Does this not mean that `frames_available` can never exceed what `GetBufferSize()` reported, and that maximum is `max_frames_in_buffer`? Under that contract, `extend_from_slice` should be dead code.

For render:

> Rendering clients can use this value to calculate the largest rendering buffer size that can be requested from IAudioRenderClient::GetBuffer during each processing pass.

Do drivers violate this in practice or am I misunderstanding?
If the contract holds up, I'd prefer `Box<[i32]>` and `&mut [i32]`.

LastExceed: Oh shoot, you're right. I confused `GetBufferSize()` with the out-param in `GetBuffer()`. I'll change it

Issue #1302, opened 2026-08-08
WASAPI: cached IMMDeviceEnumerator is created in one thread's STA and used after that thread exits


joshelson:
Calling cpal's safe API on Windows can crash the process with `STATUS_ACCESS_VIOLATION` rather than returning an error.

## Symptom

In a test binary run with `--test-threads=1`, a test that enumerates devices passes when run **alone**, but the process faults when other tests run in the same binary. The fault lands on an arbitrary later test and moves between runs. Two stacks from cdb, both faulting at `+0x19d` — the vtable read:

```
INVALID_POINTER_READ_c0000005
  windows::Win32::Media::Audio::IMMDeviceEnumerator::GetDefaultAudioEndpoint
  cpal::host::wasapi::device::default_device
  cpal::host::wasapi::device::default_input_device
  cpal::traits::HostTrait::default_input_device
```

```
INVALID_POINTER_READ_c0000005
  windows::Win32::Media::Audio::IMMDeviceEnumerator::EnumAudioEndpoints
  cpal::host::wasapi::device::Devices::new
  cpal::host::wasapi::impl$1::devices
  cpal::traits::HostTrait::input_devices
```

Observed on `windows-latest` GitHub Actions runners, cpal 0.17.3, and the same code is present on `main`.

## Diagnosis

`ENUMERATOR` in `src/host/wasapi/device.rs` is a process-wide `OnceLock<Enumerator>`, created on whichever thread calls first, with `unsafe impl Send + Sync`. COM is initialised as `COINIT_APARTMENTTHREADED` through a thread-local in `src/host/com.rs` whose `Drop` calls `CoUninitialize`.

An STA belongs to the thread that created it. So once the first thread to touch cpal exits, the cached interface pointer is dangling, and every later caller reads a dead vtable. The comment at the initialisation site states the assumption that does not hold:

```rust
// COM initialization is thread local, but we only need to have COM initialized in the
// thread we create the objects in
com::com_initialized();
```

This is most visible where audio work happens on short-lived threads: whichever one touches cpal first owns the enumerator, and its exit breaks device access for the rest of the process.

## Ruled out

- `CoCreateInstance` failing is not a fit — that would panic at the `.unwrap()`, not fault at a vtable read.
- Absent audio endpoints is not a fit either — that surfaces as an HRESULT from `GetDefaultAudioEndpoint` / `EnumAudioEndpoints`, not an invalid dereference.

## Two affected paths

[...]

PR #1310, opened 2026-08-15
fix(wasapi): don't panic when the COM enumerator fails to initialize

roderickvd:
Fix for that item from #1302 where `CoCreateInstance().unwrap()` on the cached enumerator would let a safe API panic on failure.

The STA/thread-affinity bug from that issue needs a separate, bigger fix.


LastExceed (commented):
LGTM

Issue #1083, opened 2025-12-31
If ASIO fails to load, the build crashes, but I need it to be optional


Toni997:
Hello,

I am developing an audio application and I need to have ASIO support for low-latency. So, considering that, in my cargo.toml I have "cpal = { version = "0.17.0", features = ["asio"] }". However, if it fails to download SDK or ASIO doesn't get properly loaded for whatever reason, the app just crashes and never runs. Which is not ideal for me.

Errors in my example:

error: IO error for operation on C:\...\asio_sdk\host: The system cannot find the path specified. (os error 3)
error: IO error for operation on C:\...\asio_sdk\host/pc: The system cannot find the path specified. (os error 3)
error: IO error for operation on C:\...\asio_sdk\common: The system cannot find the path specified. (os error 3)

I don't know how this error happened, I already had ASIO SDK installed because I use FL Studio, but the folder isn't named "asio_sdk", it is "ASIOSDK". And because of that, it crashes, but I would like it to just fallback to WASAPI for example and log errors for debugging. ASIO isn't required for my application.

I know how to fix this error btw, I just set "CPAL_ASIO_DIR" env variable myself, that's not an issue.

Any ideas on how I can achieve this at the moment? Or if it might be possible in the future? Thanks for taking the time to read this.  Any help is greatly appreciated.

roderickvd:
Can you post a full log including the command you're using?

Then it'll be easier for me to judge whether it makes sense to make it more robust (as you're asking) or keep it failing early (arguably, an ASIO app without ASIO installed should fail). But I may not understand your use case fully.

Toni997:
> Can you post a full log including the command you're using?
> 
> Then it'll be easier for me to judge whether it makes sense to make it more robust (as you're asking) or keep it failing early (arguably, an ASIO app without ASIO installed should fail). But I may not understand your use case fully.

Hello, thanks for the reply. About the logs, sorry, I won't be home until Monday, but let me clarify a bit till then. My audio application can run on any driver, but I want to ship it with ASIO support as well as others like WASAPI. If the user doesn't want to use ASIO for any reason, I don't want the app to crash on them because they don't have it installed. Think of FL Studio for example, they allow users to choose driver at any moment. If ASIO isn't enabled, it just won't be listed as available driver.

roderickvd:
[...]

PR #1331, opened 2026-08-19
new ASIO host implementation

### LINKED ISSUE #1299: A new ASIO backend implementation
I have recently published [azo](https://crates.io/crates/azo), a re-implementation of ASIO's underlying COM API. As mentioned in #1296, the project's goal is for it to replace `asio-sys` in `cpal`'s ASIO backend. I'm creating this issue for discussion and tracking purposes.

## Motivation

`asio-sys` is bound to the limitations of Steinberg's ASIO SDK (outlined below), which are unsolvable from the outside.

### Licensing Issues
The `ASIO SDK` has been under a propietary license for most of its life, with GPL being added as an alternative just a few months ago. Neither of them is a great fit for `cpal`:
* The propietary license (under which `ASIO SDK` was introduced into `cpal`) prohibits any form of redistribution, including even the header files, which requires developers to either
  * use a build script that automatically downloads the SDK from Steinberg's website - which is questionable both legally and security wise
  * manually provide the SDK themselves - which is inconvenient
* The GPL license is copyleft, which I doubt is 100% compatible with `cpal`'s permissive license either (IANAL, correct me if I'm wrong)

### Limited Capability
Most notably, `ASIO SDK` is incapable of initializing more than 1 driver at a time. This makes using multiple devices in parallel impossible. The spec explicitly calls this out as a limitation of the implementation, rather than a design choice.¹ There are also some other minor limitations, like not supporting strings which exceed certain fixed buffers in length, but these have essentially become part of the standard at this point.

## `azo`

solves both of these problems. It is permissively licensed, and has multi-driver support without limitations. As the author I am of course heavily biased towards it, but I'm willing to do my due diligence in terms of field testing. For that, my first step was to implement `cpal`'s custom host API, and in doing so, I encountered some problems with its design that I would like to discuss here.

### No channel Mapping

~~`cpal` doesn't have a mechanism for channel mapping at the moment, so requesting `n` channels always means `0..n`. Combined with the fact that ASIO channels are always mono (by design), this effectively means that only the first input and output channel of each device can be used. The current implementation works around this by including a mixer, which combines mono channels to stereo using an intermediate buffer. However, I'm very apprehensive of going this route, as it misrepresents the device's topology. The ability to record/playback stereo streams on a device without stereo channels might cause confusion among dependents.~~

~~One might point out that other backends feature native mixers out of the box, such as WASAPI's system wide sound server, but this isn't quite comparable, because in those cases, the native mixer IS the origin/destination of the stream, as far as `cpal` is concerned.~~

_Edit: Created a separate issue for this_
* #1301 

### No Bulk Stream Creation

~~Another problem I faced is that in `cpal`s API, streams are created one at a time, but ASIO synchronizes all channels by design, including their buffer size, sample position, buffer callback, running state, etc.~~ 

_Edit: I had confused channel layout with stream separation. See also:_
* #367

~~Therefore, it isn't even possible to use the 1 input and 1 output channel that are accessible at the same time, unless there is some serious spaghetti going on under the hood.~~

_Edit: the duplex API solved this_

### Overwhelming trait requirements

~~The device and stream traits impose some requirements that I struggled to fulfill. For instance, devices being `Clone` was counter intuitive to me. I understand the notion of it being a device identifier, rather than an life handle, but this just raises the question where the actual life handle (driver instance in ASIO's case) should reside. I ended up putting it behind an `Arc` that all `Device` and `Stream` instances hold, but it felt like cheating the system.~~

_Edit: I have since learned that these are deliberate design choices for the sake of convenience_

~~The `Send` bound on `Stream` was also tricky. ASIO uses apartment threaded COM, and painfully doesn't support marshalling AT ALL, which is very hard to make thread safe. Maybe this more of a criticism of ASIO rather than of `cpal`, but it is also what we have to work with.~~

_Edit: It definitely is a problem with ASIO, not with cpal_

## Conclusion

~~Maybe it is a skill issue, but I am unable to come up with a "correct" design of fitting ASIO into `cpal`'s API to an extent that is practically useful. I don't know how well the other backends fared in this regard, would love to hear some opinions on this.~~

~~I'm actually considering to make myself familiar with the other backends before continuing with ASIO, so I can get a bigger picture of why `cpal` is the way it is, but I fear I may be underestimating the depth of this rabbit hole.~~

_Edit: This is outdated. See_
* #1331 

---

¹ASIO 2.3 Specification, section IV.1:
> [...] However, the implementation is currently limited to one active driver. Also, it implies that a driver’s name will have max. 32 characters including the terminating '\\0'.

LastExceed:
Closes #1299 
Obsoletes #1083 

Still got a couple of clippy lints to work through, but the general structure is ready for review


src/host/asio_new/ffi_callbacks.rs
```diff
@@ -0,0 +1,225 @@
...
+    [mut in_, mut out]: [simplex::WithScratch; 2]
+) -> Bare<BufferSwitchTimeInfo> {
+    let closure = move |time: *mut Time, buf_idx: c_long, direct_process: Bool| {
+        // The ASIO spec claims `direct_process` to always be true on Windows,
+        // and dropped support for other platforms. But just in case:
+        if direct_process != Bool::TRUE {
+            throw(
+                &error_callback,
+                Error::with_message(
+                    RealtimeDenied,
```
edwloef: This is the wrong error kind, I'm pretty sure.

LastExceed:
Note to self: consider tracking driver instances globally to avoid creating a new one when another of the same device already exists

LastExceed:
Status update: I keep reworking this branch over and over because implementing ASIO *correctly* turns out to be quite the challenge. The spec has several soundness holes, and the reference implementation (ASIO SDK) breaks COMs requirements left and right. While we could easily get away with cutting quite a few corners (like the ASIO SDK does), such as not running a message pump, I am determined to do Rust's correctness culture justice and not cut any of them.

I am currently working on moving some of the internal machinery of the new ASIO host (e.g. the session factory and ffi closures) out of `cpal` and into `azo`, as this branch got way bigger than I had initially envisioned it to be. I initially avoided this to keep `azo` unopinionated, but it has become increasingly apparent that there aren't really any feasible alternatives to this machinery anyway (that I can think of) without sacrificing correctness in some capacity, so I guess it isn't that opinionated after all.

PR #1335, opened 2026-08-22
Second round of correctness fixes on multiple backends

roderickvd:
This is the result from a second pass comparing cpal against CamillaDSP's backends for correctness and robustness
gaps (first one landed in 0.18.0). Their backends moved on since (`next5`), so I went through it again.

In full transparency of licensing: CamillaDSP is MPL-2.0, cpal is Apache 2.0. That should be OK: MPL's copyleft covers licensed source files but not ideas or bugs found while reading them. Nothing was copied, rather, every fix was reimplemented against cpal's conventions. @HEnquist let me know if you have any objections.

@LastExceed would you verify the ASIO and WASAPI changes? You'll note that in WASAPI I made some small changes to pave the way for exclusive mode later.


src/host/asio/stream.rs
```diff
@@ -1275,6 +1279,27 @@ unsafe fn asio_channel_slice_mut<T>(
     unsafe { std::slice::from_raw_parts_mut(buff_ptr, channel_length) }
 }
 
+// ASIOStart() plays buffer half 1 immediately, before the first bufferSwitch can fill it;
+// half 0 is covered by that first callback.
+fn prefill_output_silence(driver: &sys::Driver, asio_streams: &Mutex<sys::AsioStreams>) {
```
LastExceed: Shouldn't this use the data callback instead of generating silence? Also I think this isn't actually necessary at all:

ASIO 2.3 specification, secion II.5:
> The audio streaming starts after the `ASIOStart()` call. Prior to starting the hardware streaming the driver will issue one or more `bufferSwitch()` or `bufferSwitchTimeInfo()` callbacks to fill its first output buffer(s). Since the hardware did not provide any input data yet the input channels' buffers should be filled with silence by the driver.

To me this clearly states that pre-filling is the driver's responsibility, not the host's. In fact, if the initial callback invocations happen within `ASIOCreateBuffers()` (on the calling thread), then the silence ends up overwriting actual stream data.

CHANGELOG.md
```diff
@@ -46,6 +46,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **PipeWire**: Fix capture reading from the wrong offset in the buffer on some devices.
 - **WASAPI**: Device enumeration no longer panics if the COM enumerator fails to initialize.
 - **WASAPI**: Output streams now start with silence instead of undefined content in the render buffer.
+- **WASAPI**: Fix `I64` and `F64` incorrectly reported as supported output formats.
```
LastExceed, line 60: Do you have a source for the claim that this was incorrect? I can't find anything that forbids WASAPI devices from using these formats

src/host/wasapi/stream.rs
```diff
@@ -577,6 +577,19 @@ fn process_commands(run_context: &mut RunContext) -> Result<bool, Error> {
...
+                    // gets real data from the first callback.
+                    if let AudioClientFlow::Render { ref render_client } =
+                        run_context.stream.client_flow
+                    {
+                        // PlayStream also fires on resume from pause, where the buffer wasn't
+                        // reset and may already be full.
+                        let frames = get_available_frames(&run_context.stream)?;
+                        if frames > 0 {
+                            write_silence(render_client, &run_context.stream, frames)?;
+                        }
```
LastExceed: Here too, shouldn't we use the data callback instead of generating silence?

src/host/wasapi/device.rs
```diff
@@ -187,19 +187,15 @@ unsafe fn data_flow_from_immendpoint(endpoint: &Audio::IMMEndpoint) -> Audio::ED
 }
 
 // Given the audio client and format, returns whether the audio engine supports it natively in
-// shared mode without format conversion.
+// the given share mode without format conversion.
 pub unsafe fn is_format_supported(
     client: &Audio::IAudioClient,
     waveformatex_ptr: *const Audio::WAVEFORMATEX,
+    share_mode: Audio::AUDCLNT_SHAREMODE,
 ) -> Result<bool, Error> {
```
LastExceed, line 195: I'd generally discourage implementing things before we need them (YAGNI principle). In this case it is probably fine, but in general you never know how requirements might change, or when a planned feature ends up getting scrapped entirely, potentially resulting in code that ends up never being used

src/host/asio/stream.rs
```diff
@@ -223,8 +223,9 @@ impl Device {
             }
             last_buffer_index = callback_info.buffer_index;
 
-            // There is 0% chance of lock contention the host only locks when recreating streams.
-            let stream_lock = asio_streams.lock().unwrap();
+            let Ok(stream_lock) = asio_streams.lock() else {
+                return;
+            };
```
LastExceed, line 228: The commit message says "do not block", but `.lock()` is a blocking function, no?

LastExceed (changes requested):
Static review only so far, haven't tested anything yet. Also can't speak on the changes to non-windows backends, as I am not familiar with them

CHANGELOG.md
roderickvd: I don't, the MSDN docs are woefully quiet about what's supported and what's not. I _believe_ that 64-bit isn't a thing in WASAPI at all. As proof of the contrary I've not found any implementations of it either. Could you test on a Windows box?

src/host/wasapi/device.rs
roderickvd: Native DSD support on WASAPI is one of the v0.19 design goals, so it was easy to rig up.

src/host/asio/stream.rs
roderickvd: Indeed, I should have written "do not panic".

src/host/asio/stream.rs
roderickvd: What I based it on is https://github.com/dechamps/ASIOUtil/blob/master/BUFFERS.md which also states this and then goes on with two more pieces of the spec and conludes that:

> the host should fill buffer 1 prior to calling `ASIOStart()` ...

It does not seem filled via `bufferSwitch` callbacks.

Is this synthesis incorrect? I thought that this source was pretty authoritative but it'd be great if you could verify.

src/host/wasapi/stream.rs
roderickvd: Yeah, not sure if I got this right. As matter of fact, CamillaDSP doesn't write silence at all, but moves `start_stream()` to after the first pass of the fill loop. This proposal of mine writes silence because `process_commands()` has no access to the data callback, at the cost of one period of silence at the start. That's smaller than the real fix would be, mirroring what CamillaDSP did. What do you think?

LastExceed: I guess it is fine for now, not worth refactoring everything just for this. But probably worth creating an issue for future reference, I could imagine similar situations existing on the other backends as well (I haven't checked)

CHANGELOG.md
LastExceed: Scratch that, I just realized that we're specifically talking about formats supported by `AUDCLNT_STREAMFLAGS_AUTOCONVERTPCM`, not by the device itself. Your commit is correct, `i64` and `f64` are not supported (`IAudioClient::Initialize()` returns `E_INVALIDARG`). See also #1343

src/host/asio/stream.rs
LastExceed: That is an awesome resource, and I have a thing or two to say about it (will create an issue there). But it doesn't actually seem to contain the conclusion you quoted, Ctrl + F "it is not filled" yields 0 results. Is this paraphrased?

My interpretation is still that the driver should invoke the callback right after buffer creation to populate the buffers, but it is indeed a bit ambiguous

roderickvd: https://github.com/dechamps/ASIOUtil/blob/master/BUFFERS.md#asiostart-and-priming

src/host/wasapi/device.rs
roderickvd: Err I meant _exclusive mode_.

src/host/asio/stream.rs
roderickvd: Fixed the quote for what was my deduction.

LastExceed: I left [a thorough comment](https://github.com/dechamps/ASIOUtil/issues/1) there, I recommend you read it too. Bottom like is that the spec contradicts itself, but it can be somewhat shoehorned into what I said earlier

roderickvd: So you're recommending to undo this particular change? As you've [later tested](https://github.com/RustAudio/cpal/pull/1335#issuecomment-5410584246) everything to work OK _with_ this change I want to leave no trace of doubt.

Though I guess it wasn't audible before this change either? The only way to be sure might be to capture the very first ~ms of audio and inspect its waveform.

src/host/wasapi/stream.rs
roderickvd: I'll see how much effort it takes to refactor to move `start_stream()` to after the first pass of the loop. I like correctness, if it doesn't cost too much.

src/host/asio/stream.rs
roderickvd: Scrap that, I'll drop it.

LastExceed: Oh sorry, I somehow missed your previous comment. I'll take a look at the output after work

LastExceed: I have tested 7 drivers:

* Steinberg "built-in" driver invokes the callback multiple times before returning from `start()`, probably to top up the underlying WASAPI buffer
* My physical audio interface's driver calls it once, and even zeros the buffers right before that (so anything written before `start()` is lost)
* `FlexASIO` (which was created by the guy that wrote the post you linked) invokes the callback multiple times immediately *after* `start()` returned, long before a full period has elapsed, probably for the same reason as Steinberg's driver
* `KoordAsio` behaves roughly the same as `FlexASIO`
* `ASIO2WASAPI` prepends 100ms of silence to the output stream, way more than fits into the ASIO buffers, so no priming is needed there
* `ASIO4ALL v2` takes over 1.5 seconds to call the callback after returning from `start()`, no idea whats going on there
* `ReaRoute` (from Reaper DAW) always sets the buffer index to 0, and has some very strange timings, so I'm assuming it is broken on my machine

TLDR: Drivers take care of priming.

roderickvd: Oh wow that’s extensive. Thanks for being so thorough!

src/host/wasapi/stream.rs
LastExceed: FYI I plan to fully rewrite the WASAPI backend once we got the 0.19 API figured out, as there are quite a few things I'd like to change (e.g. upgrading to `IAudioClient3` for lower latency), so if fixing this requires some refactoring, then don't bother.

roderickvd: I have something cooking locally that’s not all too difficult. Will push and ping you soon.

roderickvd: It's force pushed and in [ca13908](https://github.com/RustAudio/cpal/pull/1335/commits/ca13908b1fffd918dfe3f8b428c81f8087307c00) if you could do a sanity check on that.

src/host/wasapi/stream.rs
```diff
@@ -576,37 +587,40 @@ fn process_commands(run_context: &mut RunContext) -> Result<bool, Error> {
...
-                    {
-                        // PlayStream also fires on resume from pause, where the buffer wasn't
-                        // reset and may already be full.
-                        let frames = get_available_frames(&run_context.stream)?;
-                        if frames > 0 {
-                            write_silence(render_client, &run_context.stream, frames)?;
+                if run_context.stream.playback_state == PlaybackState::Stopped {
+                    // PlayStream also fires on resume from pause, where the buffer wasn't reset
+                    // and may already hold real, unplayed data.
+                    let cold_start = match run_context.stream.client_flow {
```
LastExceed, line 594: Maybe rename `cold_start` to `needs_priming`, since capture streams setting `cold_start = false` is semantically incorrect

roderickvd: Yes, can do that. If you can confirm it works, I'll update the naming and merge after.

src/host/wasapi/stream.rs
LastExceed: just 1 nitpick, rest lgtm

src/host/wasapi/stream.rs
LastExceed: Hmm, pausing immediately after starting always results in 1 callback invocation slipping through, but stopping immedately after starting does not. This isn't really a problem, but it seems strange, and I can't find an explanation for this difference

roderickvd: Try 67d6c47

LastExceed: Still same

roderickvd: Hmm. Are you calling `pause()` immediately after `start()`, with no delay in between? Or does the stream play for a bit first before you pause it?

LastExceed: immediately. its simply
```rust
stream.start().unwrap();
stream.pause().unwrap();
```
vs
```rust
stream.start().unwrap();
stream.stop().unwrap();
```

roderickvd: Welp, out of ideas then too. A WASAPI thing? May as well merge this.

LastExceed: I see no problem with merging this, there's no requirement for anything specific here after all. Its just an odd observation

roderickvd: Well, I'm stubborn. Hopefully final attempt at this in a49ad9d. Would you give that a spin in different conditions? Start/pause immediately, start/play-a-bit/pause, ...

roderickvd: @LastExceed nudge if  you would let me know?

LastExceed: That did the trick. No callback invocations in either case, even when repeating 100x in a row (unless it happens to align with a buffer signal). Playback behaviour seems fine, no audible oddities in a sine wave when pausing/resuming etc. Did not look at the code yet

roderickvd: Perfect, will resolve the conflicts, merge this to `master`, and backport to `stable-0.18`.

PR #1341, opened 2026-08-23
feat(pipewire): set node.rate prop

kpqi5858:
I was debugging why `default.clock.allowed-rates` isn't working with cpal stream, and came up with this solution.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Renamed the `wasm-beep` and `audioworklet-beep` examples to `webaudio` and `audioworklet`.
 - **ALSA**: Update `alsa` dependency to 0.12.
 - **Linux**: `realtime` can now promote threads without requiring `realtime-dbus`.
+- **PipeWire**: Set `node.rate` property so that `default.clock.allowed-rates` PipeWire config works.
 
 ### Deprecated
 
--- a/src/host/pipewire/device.rs
+++ b/src/host/pipewire/device.rs
@@ -176,6 +176,8 @@ impl Device {
         // preventing phase drift between simultaneous input/output streams.
         properties.insert("node.group", format!("cpal-{}", std::process::id()));
 
+        properties.insert(*pw::keys::NODE_RATE, format!("1/{}", config.sample_rate));
+
         if let BufferSize::Fixed(buffer_size) = config.buffer_size {
             properties.insert(
                 *pw::keys::NODE_LATENCY,
--- a/src/host/pipewire/device.rs
+++ b/src/host/pipewire/device.rs
@@ -38,7 +38,7 @@ use crate::{
             stream::{
                 DefaultDeviceMonitor, PwInitGuard, SUPPORTED_FORMATS, StreamCommand, StreamData,
             },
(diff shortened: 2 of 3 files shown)


src/host/pipewire/device.rs
```diff
@@ -174,7 +174,12 @@ impl Device {
 
         // Group input and output nodes so PipeWire schedules them in the same quantum,
         // preventing phase drift between simultaneous input/output streams.
-        properties.insert("node.group", format!("cpal-{}", std::process::id()));
+        properties.insert(
+            *pw::keys::NODE_GROUP,
```
roderickvd, line 178: Good one! I also prefer to re-use constants. Any chance there's more in src/host/pipewire/utils.rs where we could use some from `*pw::keys` instead?

CHANGELOG.md
```diff
@@ -26,6 +26,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Renamed the `wasm-beep` and `audioworklet-beep` examples to `webaudio` and `audioworklet`.
 - **ALSA**: Update `alsa` dependency to 0.12.
 - **Linux**: `realtime` can now promote threads without requiring `realtime-dbus`.
+- **PipeWire**: Set `node.rate` property.
```
roderickvd: If this fixes an issue with `default.clock.allowed-rates` then is it worth to state so?

roderickvd (commented):
Thanks for your contribution!

src/host/pipewire/device.rs
```diff
@@ -174,7 +174,12 @@ impl Device {
...
+            format!("cpal-{}", std::process::id()),
+        );
+
+        properties.insert(*pw::keys::NODE_RATE, format!("1/{}", config.sample_rate));
```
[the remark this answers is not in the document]
kpqi5858: This is wrong suggestion as cpal defines Pipewire 0.3.53 as minimum os/runtime, which exposes this.

src/host/pipewire/device.rs
kpqi5858: I removed node module from utils and replaced with pw::keys.

- METADATA_NAME is exposed in pipewire-sys but not in pipewire.
- There is pw::keys::DEVICE_ICON but that one exposes `icon-name`, not `icon_name`
- I couldn't find anything other than that.

src/host/pipewire/device.rs
```diff
@@ -1019,7 +1024,7 @@ pub fn init_devices(connect_automatically: Arc<AtomicBool>) -> Option<Vec<Device
                                 Option<FrameCount>,
                                 Option<SampleRate>,
                             ) = props
-                                .get(node::LATENCY)
+                                .get(&pw::keys::NODE_LATENCY)
```
roderickvd: Nitpick: I think `*pw::keys::FOO` is more idiomatic than `&pw::keys::FOO`.

src/host/pipewire/device.rs
roderickvd: Perfect, thanks. One nit I added to the review, to use `*pw::` instead of `&pw::`.

src/host/pipewire/device.rs
kpqi5858: Done.

PR #1348, opened 2026-09-01
fix(coreaudio): honour the full timeout when waiting for a sample rate change

dylanpulver:
`set_sample_rate` waits for `RateListener` to report the new rate, and re-arms when a notification carries some other rate. The re-arm subtracted `start.elapsed()` — the total elapsed time — from the already-reduced `remaining`, so the budget shrank super-linearly and hit zero well before the caller's timeout:

```rust
remaining = remaining.checked_sub(start.elapsed())   // total elapsed, every iteration
```

With a 500 ms timeout and a listener emitting a non-matching rate every few ms, the loop gave up after **82 ms** and returned `Sample rate update timed out` on a device that was still settling. Perversely, the chattier the device, the shorter the effective timeout. Subtracting from the original `timeout` instead makes it a single deadline; the same measurement then runs the full 500 ms.

Non-matching notifications are reachable: `RateListener`'s callback re-reads the property and sends whatever it gets, including `0.0` when that read fails — which matches no target rate.

The loop is extracted into `wait_for_rate` so it can be tested without a device. Two tests: one asserts the whole timeout is spent across repeated non-matching events (fails at 82 ms on the current code, and also on a `saturating_sub` variant of it, so it pins the accounting rather than the helper); one asserts it returns as soon as the target rate arrives. `cargo test` on macOS goes 18 → 20 passing, including the existing device tests. fmt, `clippy --all --all-targets -- -D warnings` and `cargo doc` are clean.

The other backends call `recv_timeout(dur)` once and never re-arm, so this is CoreAudio-only.

Honest limit: I could not reproduce the intermediate-rate notification on real hardware — the built-in mic and speakers here settle in a single notification — so that half is unverified. The evidence is the extracted unit test and the arithmetic.


CHANGELOG.md
```diff
@@ -43,6 +43,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **ASIO**: Fix loading a driver while a previous driver was still unloading.
 - **AudioWorklet**: Fix processor construction failures not being reported to `error_callback`.
 - **AudioWorklet**: Fix dropouts in output streams when the callback buffer grows.
+- **CoreAudio**: A sample rate change no longer gives up early when the device reports other rates first; the caller's timeout is now honoured in full.
```
roderickvd: Shorter & sweeter would be:

> CoreAudio: Fix sample rate changes timing out early when the device reports other rates first.

src/host/coreaudio/macos/device.rs
```diff
@@ -1116,3 +1133,46 @@ pub(crate) fn get_device_buffer_frame_size(
...
+mod tests {
+    use std::sync::mpsc::channel;
+    use std::time::{Duration, Instant};
+
+    use super::wait_for_rate;
+
+    /// A listener can report rates other than the target before it reports the new one, e.g. a
+    /// device stepping through rates. The whole timeout must remain available across those.
+    #[test]
+    fn wait_for_rate_honours_the_full_timeout_across_repeated_events() {
```
roderickvd, line 1236: Please don't wait 500 ms in every test run.

src/host/coreaudio/macos/device.rs
```diff
@@ -184,37 +184,54 @@ fn set_sample_rate(
...
+fn wait_for_rate(
+    receiver: &Receiver<f64>,
+    target_sample_rate: SampleRate,
+    timeout: Duration,
+) -> Result<(), Error> {
+    let start = Instant::now();
+    let mut remaining = timeout;
+
+    loop {
+        match receiver.recv_timeout(remaining) {
```
roderickvd, line 225: This still manually implements the deadline by subtracting every loop iteration. In ALSA we've got something prettier by computing one deadline up front and checking `saturating_duration_since` each iteration:

```rust
fn wait_for_rate(
    receiver: &Receiver<f64>,
    target_sample_rate: SampleRate,
    timeout: Duration,
) -> Result<(), Error> {
    let deadline = Instant::now() + timeout;

    loop {
        let remaining = deadline.saturating_duration_since(Instant::now());
        if remaining.is_zero() {
            return Err(Error::with_message(
                ErrorKind::DeviceNotAvailable,
                "Sample rate update timed out",
            ));
        }

        match receiver.recv_timeout(remaining) {
            Ok(reported_rate) => {
                if (reported_rate - target_sample_rate as f64).abs() < 1.0 {
                    return Ok(());
                }
            }
            Err(RecvTimeoutError::Timeout) => {
                return Err(Error::with_message(
                    ErrorKind::DeviceNotAvailable,
                    "Sample rate update timed out",
                ));
            }
            Err(RecvTimeoutError::Disconnected) => {
                return Err(Error::with_message(
                    ErrorKind::StreamInvalidated,
                    "Sample rate listener disconnected unexpectedly",
                ));
            }
        }
    }
}
```

As bonus, this also gets rid of a blocking `recv_timeout(ZERO)` edge case.

roderickvd (changes requested):
Thanks for your PR for what indeed is a real bug. I propose another solution though, consistent with what we do in ALSA.

src/host/coreaudio/macos/device.rs
```diff
@@ -191,37 +191,57 @@ fn set_sample_rate(
...
+/// Block until the rate listener reports `target_sample_rate`, giving up after `timeout`.
+///
+/// Notifications carrying some other rate can arrive first, so `timeout` bounds the whole wait
+/// rather than each individual receive.
+fn wait_for_rate(
+    receiver: &Receiver<f64>,
+    target_sample_rate: SampleRate,
+    timeout: Duration,
+) -> Result<(), Error> {
+    let deadline = Instant::now() + timeout;
```
roderickvd, line 214: This panics on overflow. Consider:

```rust
fn wait_for_rate(
    receiver: &Receiver<f64>,
    target_sample_rate: SampleRate,
    timeout: Option<Duration>,
) -> Result<(), Error> {
    let deadline = timeout.and_then(|t| Instant::now().checked_add(t));

    loop {
        let received = match deadline {
            Some(deadline) => {
                let remaining = deadline.saturating_duration_since(Instant::now());
                if remaining.is_zero() {
                    return Err(Error::with_message(
                        ErrorKind::DeviceNotAvailable,
                        "Sample rate update timed out",
                    ));
                }
                receiver.recv_timeout(remaining)
            }
            None => receiver.recv().map_err(|_| RecvTimeoutError::Disconnected),
        };

        match received {
            // ...
        }
    }
}
```

src/host/coreaudio/macos/device.rs
```diff
@@ -191,37 +191,57 @@ fn set_sample_rate(
...
-                        ErrorKind::DeviceNotAvailable,
-                        "Sample rate update timed out",
-                    ));
-                }
-                Err(RecvTimeoutError::Disconnected) => {
-                    return Err(Error::with_message(
-                        ErrorKind::StreamInvalidated,
-                        "Sample rate listener disconnected unexpectedly",
-                    ));
+        // otherwise default to 1 second.
```
roderickvd, line 194: Pass on the `Option`; the contract is that a `timeout` of `None` is to wait indefinitely.

roderickvd (changes requested):
Thanks, almost right, see below.

roderickvd:
Thanks for taking care of the changes. I'll merge and see if I can add a little to extra to stop indefinite waits (`timeout = None`) when a device is unplugged in the meantime.

Issue #413, opened 2020-05-22
Is there still a use-case for the emscripten host (`wasm32-unknown-emscripten` target)?


mitchmindtree:
Forgive my lack of web target experience! I'm just wondering if there's still a motivating use-case for the `wasm32-unknown-emscripten` backend that would not be covered by the webaudio target (`wasm32-unknown-unknown`)? Could any folks more familiar with the web targets weigh in on this, and perhaps discuss where use-cases might differ between the two?

est31:
Originally, wasm32-unknown-unknown only supported pure rust while emscripten supported C/C++ stuff too. I don't know how good clang's native wasm support has become since, but emscripten is still actively maintained so it's likely it's still useful for *something*.

kripken:
In general emscripten is useful for C API support, for APIs like pthreads, OpenAL, SDL2, OpenGL, longjmp, etc. (and C++ APIs like exceptions too). That is, you can build a rust project that uses C libraries like those. Emscripten also does a bunch of optimizations for size by default that other toolchains usually don't (but you can do most of those manually). I'm not sure about whether those factors are relevant for this project though!

roderickvd:
It's here, working, and so far not a burden to maintain, so let's keep it for now.

roderickvd:
Belay that: Emscripten is broken and needs to be replaced with an OpenAL host: #810. @kripken's comment remains valid for why'd we want that. For other web targets, you can use the `webaudio` or `audioworklet` hosts.

Issue #92, opened 2015-12-05
Support for emscripten


tomaka:

icefoxen:
This will probably be a priority for ggez someday, right now the best(???) option appears to basically be to target the emscripten SDL2 audio or maybe OpenAL backends.

tomaka:
I didn't update this issue, but basically the problem is that the WebAudio is not low-level enough and didn't provide some critical functionnalities, although I don't remember what exactly.

icefoxen:
Conversation on a possible approach

```
14:55 < tomaka> well, there's just no way to stream data
14:55 < tomaka> their interface is high-level, they expect you to send sound files or something
14:55 < tomaka> but it's not that far away from being possible
14:55 < jpernst> there is a way to do it
14:55 < jpernst> i'm doing it in the openal impl
14:56 < tomaka> you can send buffers with arbitrary data, but there's no timeout when it's finished or something like that
14:56 < Icefoz> There's that too.
14:56 < jpernst> ahhh, right
14:56 < tomaka> I don't remember at this point what was missing
14:56 < jpernst> there is no callback, that's true
14:56 < jpernst> openal is poll based, so it works
14:56 < jpernst> but if you rely on accurate callbacks, that is a problem
14:56 < tomaka> I don't need accurate callbacks, just callbacks at all
14:57 < jpernst> you could simulate it with an interval proc
14:57 < jpernst> for example, the openal impl runs a refresh method every 25ms or so
14:57 < tomaka> I think you could use setTimeout, but since there's no way to know when a sound is finished maybe you're 2 minutes ahead of the playback and you don't know it
14:58 < tomaka> I mean, if you start playing a sound that lasts 20ms and setup a callback for 20ms later, you're going to be too short
14:58 < tomaka> so you should setup your callback for 19ms later or something like that
14:58 < jpernst> the audio context has a "current time" field
14:58 < jpernst> which you can use to find out when a sound is done
14:58 < tomaka> ah
14:59 < jpernst> i have a queue of pending buffers and flush them out once they're "finished"
15:00 < jpernst> now, it is true that webaudio doesn't actually support buffer queueing
15:00 < jpernst> so you can get clicks and pops if your buffer times aren't exactly representable as a double, or if there's lots of resampling involved
15:01 < jpernst> but if you keep them the same rate as the native rate, and using perfect durations, it does work
```

zicklag:
This can be closed now that CPAL works on web right?

roderickvd:
[...]

PR #1349, opened 2026-09-01
Enable wasm-bindgen feature with target wasm32-unknown-emscripten

DouglasDwyer:
### Summary

[The `wasm-bindgen` tool is finally getting support for integration with Emscripten.](https://github.com/wasm-bindgen/wasm-bindgen/issues/5237) The feature is still quite new, but this means that most of Rust's web ecosystem can now work with the Emscripten target. I'd like to use `cpal` in a WASM/Emscripten project, but right now the `wasm32-unknown-emscripten` target is hard-coded to use the null backend. This PR eliminates the feature gate to make the `wasm-bindgen` feature work with `wasm32-unknown-emscripten` too.

This PR exposes the `webaudio` backend but not the `audioworklet` backend. That backend relies on re-instantiating the WASM module, but the way Emscripten modules get instantiated is different, so it wouldn't work without more changes.

### Changes

- Replace `#[cfg(all(target_arch = "wasm32", target_os = "unknown", feature = "wasm-bindgen"))]` with `#[cfg(all(target_arch = "wasm32", feature = "wasm-bindgen"))]` for the `webaudio` backend
- Use `Closure::wrap_aborting` instead of `Closure::wrap` so that the code properly compiles on WASM targets with `panic=unwind`
- Bump `wasm-bindgen` dependency to `0.2.110` in order to use `Closure::wrap_aborting`
- Add an additional CI job for the Emscripten target

### Testing

In my own project, I have gotten `cpal` audio working with a Rust/Emscripten WASM module in Chrome. This PR also adds CI checks to ensure that compilation is successful.

### Related issues

#92 #413 #810

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `DeviceTrait::build_duplex_stream()`, `build_duplex_stream_raw()`, `default_duplex_config()`, and `supports_duplex()` for capture and playback from one device-level callback.
 - **AudioWorklet**: Input and duplex streams are now supported.
 - **WebAudio**: Input and duplex streams are now supported.
+- **WebAudio**: Added support for Emscripten targets via [wasm-bindgen/Emscripten integration](https://github.com/wasm-bindgen/wasm-bindgen/issues/5237).
 
 ### Changed
 
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -199,8 +199,8 @@ objc2-avf-audio = { version = "0.3", default-features = false, features = [
   "AVAudioSessionTypes",
 ] }
 
-[target.'cfg(all(target_arch = "wasm32", target_os = "unknown"))'.dependencies]
-wasm-bindgen = { version = "0.2", optional = true }
+[target.'cfg(all(target_arch = "wasm32", any(target_os = "emscripten", target_os = "unknown")))'.dependencies]
+wasm-bindgen = { version = "0.2.127", optional = true }
 wasm-bindgen-futures = { version = "0.4", optional = true }
 futures-channel = { version = "0.3", optional = true }
 futures-util = { version = "0.3", default-features = false, optional = true, features = [
--- a/README.md
+++ b/README.md
@@ -52,7 +52,8 @@ The minimum Rust version (MSRV) and minimum operating system / runtime version b
 | PipeWire | Linux, BSD | 1.85 | PipeWire 0.3.53 |
 | PulseAudio | Linux, BSD | 1.88 | — |
 | WASAPI / ASIO | Windows | 1.85 | Windows 10 |
-| WASM (`wasm32-unknown`) | WebAssembly | 1.85 | — |
+| WASM (`wasm32-unknown-unknown`) | WebAssembly | 1.85 | — |
+| WASM (`wasm32-unknown-emscripten`) | WebAssembly | 1.85 | Emscripten 6.0.3, wasm-bindgen 0.2.127 |
 | WASM (`wasm32-wasip1`) | WebAssembly | 1.85 | — |
 | WASM (`audioworklet`) | WebAssembly | nightly | — |
 
@@ -70,7 +71,7 @@ The `audioworklet` backend additionally requires `-Zbuild-std` with atomics supp
 | `pulseaudio` | Linux, BSD | PulseAudio sound server backend. Requires `libpulse-dev` (Debian/Ubuntu) or `pulseaudio-libs-devel` (Fedora). |
 | `realtime` | Android, Linux, Windows | Raises the audio callback thread to real-time or high-priority scheduling for lower latency. On Linux, requires `CAP_SYS_NICE`, root, or an `rtprio` limit granted via `limits.conf` or systemd, unless `realtime-dbus` is also enabled. |
 | `realtime-dbus` | Linux | Uses `rtkit` via D-Bus for RT scheduling on Linux desktop systems. Implies `realtime` on all platforms. Requires `libdbus-1-dev` on Linux. |
-| `wasm-bindgen` | WebAssembly (`wasm32-unknown-unknown`) | Web Audio API backend for browser-based audio; required for any WebAssembly audio support. See the `webaudio` example. |
+| `wasm-bindgen` | WebAssembly (`wasm32-unknown-emscripten`, `wasm32-unknown-unknown`) | Web Audio API backend for browser-based audio; required for any WebAssembly audio support. See the `webaudio` example. |
 
 See the [beep example](examples/beep.rs) for selecting the backend at runtime.
 
--- a/src/host/webaudio/main_thread.rs
+++ b/src/host/webaudio/main_thread.rs
@@ -36,10 +36,20 @@ mod emscripten {
         ) -> bool;
     }
 
+    /// Runs `func` on the browser main thread. For the Emscripten target,
+    /// always succeeds with [`Some`].
+    pub fn try_run<F, R>(func: F) -> Option<R>
+    where
+        F: FnOnce() -> R + Send,
+        R: Send,
+    {
+        Some(run(func))
+    }
+
     /// Run `func` on the browser main thread and return its result, blocking
     /// the caller until it completes. Runs inline when the caller already is
     /// the main thread.
-    pub fn run<F, R>(func: F) -> R
+    fn run<F, R>(func: F) -> R
     where
         F: FnOnce() -> R + Send,
         R: Send,
@@ -92,35 +102,12 @@ mod emscripten {
             slot.ret.take().expect("proxied task did not run")
         }
     }
-
-    /// Runs `func` on the browser main thread. For the Emscripten target,
-    /// always succeeds with [`Some`].
-    pub fn try_run<F, R>(func: F) -> Option<R>
-    where
-        F: FnOnce() -> R + Send,
-        R: Send,
-    {
-        Some(run(func))
-    }
 }
 
 /// Proxying implementation for `wasm32-unknown-unknown`: will check to see
 /// if closures are running on the main thread, and fail if ever called from a worker.
 #[cfg(target_os = "unknown")]
 mod unknown {
-    /// Asserts that this is the main browser thread and runs `func`.
-    pub fn run<F, R>(func: F) -> R
-    where
-        F: FnOnce() -> R + Send,
-        R: Send,
-    {
-        assert!(
-            is_main_thread(),
-            "proxying closures is not supported on wasm32-unknown-unknown"
-        );
-        func()
-    }
-
     /// Attempts to run `func`. If this was not already the main browser thread,
     /// then returns [`None`] because proxying is not possible on this target.
     pub fn try_run<F, R>(func: F) -> Option<R>
@@ -133,6 +120,19 @@ mod unknown {
         } else {
             None
         }
+    }
+
+    /// Asserts that this is the main browser thread and runs `func`.
+    fn run<F, R>(func: F) -> R
+    where
+        F: FnOnce() -> R + Send,
+        R: Send,
+    {
+        assert!(
+            is_main_thread(),
+            "proxying closures is not supported on wasm32-unknown-unknown"
+        );
+        func()
     }
 
     /// Whether this is the main browser thread.
--- a/src/host/webaudio/main_thread.rs
+++ b/src/host/webaudio/main_thread.rs
@@ -1,0 +1,142 @@
+//! Utility for invoking functions on the browser's main thread.
+//! This allows for creating audio contexts within workers,
+//! when normally having access to the window object is required.
+
+#[cfg(target_os = "emscripten")]
+pub use self::emscripten::*;
+
+#[cfg(target_os = "unknown")]
+pub use self::unknown::*;
+
+/// Proxying implementation for `wasm32-unknown-emscripten`: will send
+/// functions through Emscripten's queue to run on the main thread.
+#[cfg(target_os = "emscripten")]
+mod emscripten {
+    use std::ffi::c_void;
+
+    // Functions provided by `emscripten/proxying.h` and `emscripten/threading.h`
+    unsafe extern "C" {
+        /// Returns true if the current thread is the thread that hosts the Emscripten
+        /// runtime.
+        fn emscripten_is_main_runtime_thread() -> bool;
+
+        /// Returns the thread ID of the thread that hosts the Emscripten runtime.
+        fn emscripten_main_runtime_thread_id() -> usize;
+
+        /// Get the queue used for proxying low-level runtime work.
+        fn emscripten_proxy_get_system_queue() -> *mut c_void;
+
+        /// Enqueue `func` to be called with argument `arg` on the given queue
+        /// and thread then wait for `func` to be executed synchronously before returning.
+        fn emscripten_proxy_sync(
+            queue: *mut c_void,
+            target_thread: usize,
+            func: extern "C" fn(*mut c_void),
+            arg: *mut c_void,
+        ) -> bool;
+    }
+
+    /// Run `func` on the browser main thread and return its result, blocking
+    /// the caller until it completes. Runs inline when the caller already is
+    /// the main thread.
+    pub fn run<F, R>(func: F) -> R
+    where
+        F: FnOnce() -> R + Send,
+        R: Send,
+    {
+        /// Data that gets sent between the two browser threads.
+        struct SyncSlot<F, R> {
+            /// The function to execute.
+            func: Option<F>,
+            /// The value that was returned.
+            ret: Option<R>,
+        }
+
+        /// Callback that gets invoked on the main thread.
+        extern "C" fn trampoline<F, R>(arg: *mut c_void)
+        where
+            F: FnOnce() -> R,
+        {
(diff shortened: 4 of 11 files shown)


src/host/webaudio/mod.rs
```diff
@@ -457,7 +457,8 @@ impl DeviceTrait for Device {
                         ),
                     }
                 })
-                    as Box<dyn FnMut(AudioProcessingEvent)>);
+                    as Box<dyn FnMut(AudioProcessingEvent)>;
+                let on_audio_process = Closure::wrap_aborting(on_audio_process_fn);
```
roderickvd: `wrap_abort` instead of `wrap` would abort the whole WASM instance for regular WebAudio users, where currently they just get a JS exception.

Cargo.toml
```diff
@@ -199,8 +199,8 @@ objc2-avf-audio = { version = "0.3", default-features = false, features = [
   "AVAudioSessionTypes",
 ] }
 
-[target.'cfg(all(target_arch = "wasm32", target_os = "unknown"))'.dependencies]
-wasm-bindgen = { version = "0.2", optional = true }
+[target.'cfg(target_arch = "wasm32")'.dependencies]
```
roderickvd: This is not necessary for `wasm32-wasip1`/`wasip2`.

CHANGELOG.md
```diff
@@ -24,6 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `InputCallbackInfo`/`OutputCallbackInfo` merged into `CallbackInfo`.
 - `InputStreamTimestamp`/`OutputStreamTimestamp` merged into `StreamTimestamp`; `capture`/`playback` renamed `device`.
 - Renamed the `wasm-beep` and `audioworklet-beep` examples to `webaudio` and `audioworklet`.
+- **WebAudio**: the host and its `wasm-bindgen`/`js-sys`/`web-sys` dependencies are now gated on `target_arch = "wasm32"` rather than `all(target_arch = "wasm32", target_os = "unknown")`, so the WebAudio host works under `wasm32-unknown-emscripten` (previously it fell back to the `null` host). The AudioWorklet host stays `wasm32-unknown-unknown`-only. Raises the minimum `wasm-bindgen` to 0.2.110.
```
roderickvd: Please make this consistent with the other lines: short and observable behavior, not implementation details.

roderickvd (changes requested):
That's great! Not so long ago we removed the old Emscripten host that had become defunct. This seems like a light-weight manner to get Emscripten support back.

Beyond the changes requested in the review points, please also consider updating `README.md` with Emscripten support.

src/host/webaudio/mod.rs
DouglasDwyer: What happened was that:
- `Closure::wrap` expects an `UnwindSafe` closure whenever `panic=unwind`
- The normal `wasm32-unknown-unknown` target is typically `panic=abort`, so this compiled fine
- The new `wasm32-unknown-emscripten` target compiles with `panic=unwind`, and `dyn FnMut(AudioProcessingEvent)` is not `UnwindSafe`, so there was a compilation error

Claude's solution to this was to switch to `wrap_abort`. But it looks like we can just do `Closure::wrap(...)` directly on the lambda without needing to cast to a trait object first. The Rust type system is happy and the lambda implements `UnwindSafe`. So I've gone back to using `wrap`.

Cargo.toml
DouglasDwyer: I added a `any(target_os = "emscripten", target_os = "unknown")` wherever there was a `target_os = "unknown"` before

README.md
```diff
@@ -52,7 +52,8 @@ The minimum Rust version (MSRV) and minimum operating system / runtime version b
 | PipeWire | Linux, BSD | 1.85 | PipeWire 0.3.53 |
 | PulseAudio | Linux, BSD | 1.88 | — |
 | WASAPI / ASIO | Windows | 1.85 | Windows 8 |
-| WASM (`wasm32-unknown`) | WebAssembly | 1.85 | — |
+| WASM (`wasm32-unknown-unknown`) | WebAssembly | 1.85 | — |
+| WASM (`wasm32-unknown-emscripten`) | WebAssembly | 1.85 | Emscripten 6.0.3, wasm-bindgen 0.2.127 |
```
roderickvd, line 56: If it's `wasm-bindgen` 0.2.127 and up specifically, then should that version not also be pinned in `Cargo.toml`?

CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **ALSA**: Update `alsa` dependency to 0.12.
 - **Linux**: `realtime` can now promote threads without requiring `realtime-dbus`.
 - **PipeWire**: Set `node.rate` property so that `default.clock.allowed-rates` PipeWire config works.
+- **WebAudio**: the WebAudio host now works on `wasm32-unknown-emscripten` via [wasm-bindgen/Emscripten integration](https://github.com/wasm-bindgen/wasm-bindgen/issues/5237).
```
roderickvd: I propose to move this to the "Added" section, rephrasing it to "WebAudio: added support for Emscripten targets via ..."

src/platform/mod.rs
roderickvd, line 956: ...or now Emscripten...

src/host/webaudio/mod.rs
```diff
@@ -395,7 +395,7 @@ impl DeviceTrait for Device {
                 #[cfg(target_feature = "atomics")]
                 let current_time_bits_cb = current_time_bits.clone();
 
-                let on_audio_process = Closure::wrap(Box::new(move |event: AudioProcessingEvent| {
+                let on_audio_process_fn = Box::new(move |event: AudioProcessingEvent| {
```
roderickvd: Are these changes to make it a `_fn` and wrap it later still necessary? Seems they can be removed and reduce the diff size, unless you think it reads a lot better or I'm missing something else.

roderickvd (commented):
Thanks for the quick turnaround. Here's a few points, hopefully the last.

src/host/webaudio/mod.rs
DouglasDwyer: That's a good point. Interestingly, removing the extra `_fn` variables makes `rustfmt` want to indent the entire lambda, making the diff cluttered: 8855eb82847dd4450d287e53b0b567d8eb6a45e5. (I guess the `as Box` was preventing it from doing that before?)

For that reason I'd lean toward keeping it as-is (probably why the AI agent chose to create the extra variable in the first place) but let me know what you'd prefer.

src/platform/mod.rs
DouglasDwyer: Running on the browser's UI thread (and having access to the Window object) is still required, even with the Emscripten target. As such, I think this statement is clear and accurate - if you are running on one of Emscripten's worker threads, it won't work. We could change it to say "AudioContext requires the main UI thread" if that would be better?

CHANGELOG.md
DouglasDwyer: Moved to the added section and reworded in 40c1b2b

README.md
DouglasDwyer: Pinned to 0.2.127 in 40c1b2b

src/platform/mod.rs
DouglasDwyer: Actually, I would mention that we _could_ call [`emscripten_proxy_sync`](https://emscripten.org/docs/api_reference/proxying.h.html) when creating `Stream`s, which would allow them to be created from any Emscripten thread. I didn't do that yet because I wanted to keep the scope of this PR very small. Would you be open to that addition as well? If so, I can add it as part of this PR or in a separate PR.

roderickvd: Sure, maybe you can propose it as a separate commit, then we can judge to keep it in or do it separately.

src/host/webaudio/mod.rs
roderickvd: Same, feel free do do such formatting in another commit so it’s easier to navigate the PR diff.

DouglasDwyer: Removed the variables in 22cf4cc4f6fdfd5f6caeb655895c29a259953ef7, ran reformatting in 16dd7061985f5b912e694e0c9684c1d09303eecf

src/platform/mod.rs
DouglasDwyer: I implemented this in 7f20ece2256be65e84f5e97d0adf31e86ccd125d and 31691dfc1502331d3a1bc65a2ca6c9e94c5064f8. I will test it in my project, and also test the webaudio example in this repo to make sure it still works

DouglasDwyer: Following up - I have confirmed that the webaudio example in this repo (`wasm32-unknown-unknown`), as well as my own project (multithreaded `wasm32-unknown-emscripten`) work. I am able to get a beep going in Chrome.

src/host/webaudio/main_thread.rs
```diff
@@ -0,0 +1,142 @@
...
+    /// then returns [`None`] because proxying is not possible on this target.
+    pub fn try_run<F, R>(func: F) -> Option<R>
+    where
+        F: FnOnce() -> R + Send,
+        R: Send,
+    {
+        if is_main_thread() {
+            Some(run(func))
+        } else {
+            None
```
roderickvd: `None` maps to `ErrorKind::UnsupportedConfig` but not being able to proxy seems more like an `UnsupportedOperation. Maybe return a `Result` rather than `Option` here?

src/host/webaudio/main_thread.rs
```diff
@@ -0,0 +1,142 @@
...
+            };
+
+            let ok = emscripten_proxy_sync(
+                emscripten_proxy_get_system_queue(),
+                emscripten_main_runtime_thread_id(),
+                trampoline::<F, R>,
+                (&raw mut slot).cast(),
+            );
+
+            assert!(
```
roderickvd: How does this not panic with `try_run`?
Both this `assert!()` and the `.expect()` below.

src/host/webaudio/main_thread.rs
```diff
@@ -0,0 +1,142 @@
...
+        }
+    }
+
+    /// Asserts that this is the main browser thread and runs `func`.
+    fn run<F, R>(func: F) -> R
+    where
+        F: FnOnce() -> R + Send,
+        R: Send,
+    {
+        assert!(
```
roderickvd: This is unreachable, because it's already checked in `try_run` and there's no other way that this private function is called.

src/host/webaudio/main_thread.rs
```diff
@@ -0,0 +1,142 @@
...
+/// functions through Emscripten's queue to run on the main thread.
+#[cfg(target_os = "emscripten")]
+mod emscripten {
+    use std::ffi::c_void;
+
+    // Functions provided by `emscripten/proxying.h` and `emscripten/threading.h`
+    unsafe extern "C" {
+        /// Returns true if the current thread is the thread that hosts the Emscripten
+        /// runtime.
+        fn emscripten_is_main_runtime_thread() -> bool;
```
roderickvd, line 22: Does Emscripten define this as bool or as int?

src/host/webaudio/main_thread.rs
```diff
@@ -0,0 +1,142 @@
...
+        fn emscripten_proxy_get_system_queue() -> *mut c_void;
+
+        /// Enqueue `func` to be called with argument `arg` on the given queue
+        /// and thread then wait for `func` to be executed synchronously before returning.
+        fn emscripten_proxy_sync(
+            queue: *mut c_void,
+            target_thread: usize,
+            func: extern "C" fn(*mut c_void),
+            arg: *mut c_void,
+        ) -> bool;
```
roderickvd, line 37: Does Emscripten define this as bool or as int?

roderickvd (changes requested):
Thanks for the iteration. Some more points inside.
Additionally:
- would you add `wasm-emscripten` to the `publish-cpal` gate?
- add a CI path for threading and the proxying?

src/host/webaudio/main_thread.rs
DouglasDwyer: It's defined here, as a bool:

https://github.com/emscripten-core/emscripten/blob/main/system/include/emscripten/threading.h#L31-L34

src/host/webaudio/main_thread.rs
DouglasDwyer: It's defined here, as a bool:

[emscripten_proxy_sync](https://github.com/emscripten-core/emscripten/blob/main/system/include/emscripten/proxying.h#L60-L67)

Albeit some of the other arguments are type aliases (like `pthread_t`) but I think it would be beyond the scope to include those. I will say that it would be nice if there were an idiomatic Rust crate that wrapped these functions. I did look on crates.io, but I couldn't find any. If I end up using these functions to patch other projects/libraries, then maybe I'll make one myself eventually.

src/host/webaudio/main_thread.rs
DouglasDwyer: Yeah, originally I was going to use `run` and `try_run` in different places, but at this point only `try_run` is used, so I'll just get rid of this function

src/host/webaudio/main_thread.rs
DouglasDwyer: My intended semantics for `try_run` are:

- If it's possible to run the closure on the main browser thread, do it
- If it wasn't possible, return `null`

My intention was that `try_run` would always return `Some` on Emscripten targets, since Emscripten always has a main thread. The only case where `emscripten_proxy_sync` returns `false` [is when the target thread is null or cancelled](https://emscripten.org/docs/api_reference/proxying.h.html#functions), but I don't think that can reasonably happen for the UI thread? So I added these asserts to document invariants that I would never expect to fail. The main point of having `try_run` return `None` was for `wasm32-unknown-unknown`, where proxying isn't possible.

DouglasDwyer: Since (according to my best understanding) these `expect`s should never fail - it should always be possible to proxy something to Emscripten's runtime thread - I would push to leave them as is.

src/host/webaudio/main_thread.rs
DouglasDwyer: In the old code path, when attempting to create an input/output stream from a worker thread, it would fail on this line:

https://github.com/RustAudio/cpal/blob/master/src/host/webaudio/mod.rs#L324-L330

So I chose that error variant because it would preserve the existing behavior. But I like the idea of returning a `Result` here - much more explicit. I'll go ahead and make that change.

DouglasDwyer: Switched to `Result` in 3492db2

src/host/webaudio/main_thread.rs
DouglasDwyer: Got rid of separate `run` functions in 3492db2

src/host/webaudio/main_thread.rs
roderickvd: I understand. Keeping the `expect` indeed is better than simply unwrapping, but then why do we still need the assertion?

src/host/webaudio/main_thread.rs
roderickvd: The switch to the `Result` looks good, but the Rustdoc is stale and still talks about returning `Some` or `None` with both `try_run` implementations.

src/host/webaudio/main_thread.rs
DouglasDwyer: Good point - I have left the `expect`s, but the `assert` is unnecessary, because if the function didn't run then the `expect` on the very next line will fail. Removed the extra check in 411f5cc

src/host/webaudio/main_thread.rs
DouglasDwyer: I should have caught that. Thanks. Fixed in 411f5cc

src/host/webaudio/main_thread.rs
```diff
@@ -78,17 +78,13 @@ mod emscripten {
...
                 emscripten_main_runtime_thread_id(),
                 trampoline::<F, R>,
                 (&raw mut slot).cast(),
             );
 
-            assert!(
-                ok,
-                "emscripten_proxy_sync to the browser main thread failed"
-            );
             Ok(slot.ret.take().expect("proxied task did not run"))
```
roderickvd: Nitpick: you can use `.ok_or()` to transform an `Option` into a `Result`.

DouglasDwyer: From the way I understand your suggestion, this would look like:

```rust
slot.ret.ok_or_else(|| panic!("proxied task did not run"))
```

I'm not totally convinced that the control flow is more obvious here. With the current code, you can look and see the `expect` followed by the `Ok`, and see that it's an unwrap (or panic) followed by putting the value into a `Result`. With `ok_or_else`, the reader needs to parse the `panic!` and then the fact that it only happens in this conditionally-executing closure.

Of course, we could properly return an `Err` here for the exceptional case:

```rust
slot.ret.ok_or_else(|| Error::with_message(
    ErrorKind::BackendError,
    "proxied task did not run",
))
```

But from our previous conversations here I thought that we were keeping the panic-on-error.

---------------------------

I did notice that the `.take()` is useless here though, so I'll remove that in any case.

roderickvd: Yes, I meant that we might as well defensively return `Err` if we've got that function signature anyway.

DouglasDwyer: Done in 3f585a7

PR #1350, opened 2026-09-01
Restore reroute notifications on WASAPI

### LINKED ISSUE #1339: WASAPI: incorrect `StreamInvalidated` instead of `DeviceChanged` when default device changes
#1267 stopped emitting `DeviceChanged` claiming that WASAPI doesn't automatically reroute streams to changed default devices, but that does not seem to be true.

It's documented here:
https://learn.microsoft.com/en-us/windows/win32/coreaudio/automatic-stream-routing

I also tested this on my machine and audio *does* change device even when not re-creating the stream. I only tested with audio output, so I can't say for sure that it also works for input but the documentation claims it.

Re-creating the stream also takes longer than the automatic rerouting (an additional ~50ms in my configuration).

LastExceed:
Closes #1339

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **AudioWorklet**: Fix stale output when the data callback grows Wasm memory.
 - **JACK**: Channel enumeration is capped at the physical system port count again.
 - **WASAPI**: Device enumeration no longer panics if the COM enumerator fails to initialize.
+- **WASAPI**: Revert "Default device changes no longer report `DeviceChanged`" as it was misinformed.
 
 ## [0.18.2] - 2026-08-16
 
--- a/src/host/wasapi/stream.rs
+++ b/src/host/wasapi/stream.rs
@@ -727,11 +727,10 @@ fn boost_current_thread_priority(
     }
 }
 
-// WASAPI never rebinds the IAudioClient, so report what's actually true instead of DeviceChanged.
 fn default_device_change_error(flow: Option<Audio::EDataFlow>) -> Error {
     match flow.and_then(get_current_default) {
         None => ErrorKind::DeviceNotAvailable.into(),
-        Some(_) => ErrorKind::StreamInvalidated.into(),
+        Some(_) => ErrorKind::DeviceChanged.into(),
     }
 }
 


src/host/wasapi/stream.rs
```diff
@@ -753,7 +739,7 @@ fn process_commands_and_await_signal(
         if flag.swap(false, Ordering::Relaxed) {
             emit_error(
                 error_callback,
-                default_device_change_error(run_context.default_device_flow),
+                Error::with_message(ErrorKind::DeviceChanged, "Default audio device changed"),
```
roderickvd: This seems to throw out more than #1339 needs. `StreamInvalidated` _is_ wrong when a replacement device exists, but `default_device_change_error`'s other branch serves a different purpose:

`OnDeviceStateChanged` and `OnDeviceRemoved` only `SetEvent` when `get_current_default(self.flow).is_none()` (see the comment above them), so they only ever wake this code up for the no-replacement case. After this revert that case gets reported as `DeviceChanged` too, which contradicts its definition: "the stream remains active and no rebuild is required." There's nothing to route to there, so that isn't true.

I think the fix should be smaller than a full revert: keep `default_device_change_error`, just fix the one arm that was actually wrong:

```rust
fn default_device_change_error(flow: Option<Audio::EDataFlow>) -> Error {
    match flow.and_then(get_current_default) {
        None => ErrorKind::DeviceNotAvailable.into(),
        Some(_) => ErrorKind::DeviceChanged.into(), // was StreamInvalidated
    }
}
```

CHANGELOG.md
```diff
@@ -15,6 +15,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **AudioWorklet**: Fix stale output when the data callback grows Wasm memory.
 - **JACK**: Channel enumeration is capped at the physical system port count again.
 - **WASAPI**: Device enumeration no longer panics if the COM enumerator fails to initialize.
+- **WASAPI**: Revert "Default device changes no longer report `DeviceChanged`" as it was misinformed.
```
roderickvd, line 18: I think this should be a bit more succinct: just document observable behavior, no need to pony up about being wrong or right.

roderickvd (changes requested):

src/host/wasapi/stream.rs
LastExceed: _Edit: Not proud of this comment. Keeping it for archival purposes_

I gotta be honest: I struggle to make sense of the code. To me, `flow` being an `Option` is confusing. Every stream has a direction, why would it ever be optional? Apparently this is used to indicate whether the given device is initialized as a default device, because the flow direction is stored in the monitor (why?) and the monitor itself only exists for default devices (that part makes sense I guess). But then I have to ask: Why is it called `default_device_changed_error()` when it is also meant for non-default device streams? Also, I noticed that the monitor is the only place where the `pending_device_changed` signal ever gets set, which means the function is only ever called for default devices. Maybe that explains the name, but then we're back to questioning the optionality of the parameter. Next, it calls `get_current_default()` to check if a default device is present, even though (as you said) the signal is only set when it was already confirmed that there is none. And the list just keeps going.

The longer I look at this code, the more questions arise in my head. It all reeks of vibe coding, and that makes me sad. I don't even want to look at other backends. On the flipside, I now feel 100-fold re-affirmed that rewriting the WASAPI backend from scratch was the right decision. It just so happens that I already re-wrote the ASIO backend (for unrelated reasons), and considering that we are already planning to majorly revamp cpal's public API for v0.19, I am genuinely starting to wonder if maybe we should just go all the way and start over entirely at this point...

Ok I am getting carried away. I have reduced the PR to what you suggested, and re-tested it on hardware.

CHANGELOG.md
LastExceed: Sorry, language barrier, what does "pony up" mean in this context? If what I wrote came off as offensive, then I'm sorry

I dropped the second sentence, assuming that is what you meant

src/host/wasapi/stream.rs
roderickvd: > The longer I look at this code, the more questions arise in my head. It all reeks of vibe coding, and that makes me sad. I don't even want to look at other backends.

You seem passionate about this. I appreciate that, but do ask that we keep it forward-looking.

> I struggle to make sense of the code. To me, flow being an Option is confusing. Every stream has a direction, why would it ever be optional? 

I agree that the way it’s written, it’s trying to be smarter than it perhaps should be. 

`flow` is being used as a tri-state parameter to differentiate between `default_input_device`, `default_output_device` and an item of `Host::devices()` that the user chose explicitly and just so happens to be equal to the host’s default device at that time. For the first two cases, the intent is that cpal follows default device reroutings, for the case where a user chose a device specifically, it should stick to that device without being rerouted. An enum may convey this more clearly. 

> But then I have to ask: Why is it called default_device_changed_error() when it is also meant for non-default device streams? Also, I noticed that the monitor is the only place where the pending_device_changed signal ever gets set, which means the function is only ever called for default devices. 

The way it’s written (and being too smart about it) is why the arms leading to `default_device_changed_error()` can never be reached unless it’s a default device: like you said, the flag never gets set. 

> I now feel 100-fold re-affirmed that rewriting the WASAPI backend from scratch was the right decision.

Be careful for greenfield promises. You strike me as a seasoned enough developer to know that while code accumulates technical debt, it also accumulates a body of fixes and edge case handling. If not for reasons like an existing architecture no longer being able to accommodate growth, it may be more prudent to refactor iteratively before throwing away a trusted pair of shoes. 

Would you like to refactor this tri-state before landing this PR?

CHANGELOG.md
roderickvd: Thanks, that’s what I meant. I’m no native speaker either. “To pony up” is my attempt at looking cool instead of saying “to admit”.

src/host/wasapi/stream.rs
LastExceed: > keep it forward-looking.

You're right, that was unconstructive, and I should not have said it. Sorry

> Be careful for greenfield promises

Sorry, I have a tendency to get ahead of myself. I recognize that I certainly did here.

> it also accumulates a body of fixes and edge case handling

I know what you mean, and I'm aware that I can be quite the naive idealist at times. But with all the changes I have in mind (WinRT, `IAudioClient3`, the new API, ...), I feel like there won't be much left of the current code either way, and if we're going to replace every plank of the ship anyway, then the sum of all those iterative refactorings will ultimately have the same potential for regressions as doing it all in one go. I actually considered treating Win32 and WinRT as separate hosts at some point, but eventually decided that there is no point, since the only scenario where Win32 is actually preferrable is when WinRT is not available, which is effectively ruled out by our MSRV.

> Would you like to refactor this tri-state before landing this PR?

Maybe, but I don't fully understand the situation yet. If `default_device_changed_error()` can never be reached on non-default device streams, then what is the purpose of the tri-state?

roderickvd: All good.

> If `default_device_changed_error()` can never be reached on non-default device streams, then what is the purpose of the tri-state?

I think you're right that the `None` arm in `Option<EDataFlow>` is now unreachable and that it no longer needs to be wrapped in an `Option`.

Second, we can probably drop `default_device_flow` and use `run_context.stream.client_flow` instead.

LastExceed: I think the whole function can essentially be reduced to `ErrorKind::DeviceChanged.into()`, which then might as well be inlined, right?

I've also decided that its probably better to do 'refactorings on a separate branch, to keep the PR small and focused. If there aren't any further concerns, then I'm good with merging this

roderickvd: > I think the whole function can essentially be reduced to `ErrorKind::DeviceChanged.into()`, which then might as well be inlined, right?

The simpler, the better. 

> I've also decided that its probably better to do 'refactorings on a separate branch, to keep the PR small and focused. 

As you wish. I don’t mind opportunistic refactoring. In fact I quite like it.

Issue #1343, opened 2026-08-25
Consider supporting `IAudioClient3::InitializeSharedAudioStream()` on WASAPI


LastExceed:
The key difference to `IAudioClient::Initialize()` is that instead of converting your stream to whatever the engine chose for the underlying device, you actually get to specify the device's format (like in exclusive mode). This is interesting for several reasons:
* The device might support formats that the engine's auto-converter doesn't
* It potentially saves a conversion step
* It allows for lower latency (by lowering the engine's periodicity)

Not sure how important this is, just throwing it out there

Marco-Farruggio:
Just wanted to mention, I checked the [Windows docs](https://learn.microsoft.com/en-gb/windows/win32/api/audioclient/nf-audioclient-iaudioclient3-initializesharedaudiostream) for this, and this would raise cpal's minimum windows requirement to Windows 10, unless we only used it if we could and instead used  `IAudioClient::Initialize()` on older versions if we didn't want to raise the minimum floor

LastExceed:
Funnily this *just* came up yesterday in another issue :D Turns out we already require Win10+ anyway: https://github.com/RustAudio/cpal/issues/1302#issuecomment-5496440590

Marco-Farruggio:
Oh that's interesting, I'll open a PR to update the docs then :)

roderickvd:
Sound interesting. I don't think that [wasapi-rs](https://github.com/HEnquist/wasapi-rs) does that (which CamillaDSP uses, and should have good exclusive mode support)?

PR #1351, opened 2026-09-02
Update documentation for Windows minimum version

Marco-Farruggio:
# Summary
Update the documented minimum Windows version to reflect the Windows versions supported by cpal's current MSRV. Originally mentioned here: https://github.com/RustAudio/cpal/issues/1302#issuecomment-5496440590

Brought to my attention by LastExceed while discussing https://github.com/RustAudio/cpal/issues/1343, I just wrote up the quick PR

# Changes
- Updated the minimum windows version in the README.md
- Added an entry in the changelog documenting when it was originally, actually raised

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -339,6 +339,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Bump overall MSRV to 1.78.
+- Bumped minimum Windows version to 10.
 - **ALSA**: Update `alsa` dependency to 0.11.
 - **ALSA**: Bump MSRV to 1.82.
 - **CoreAudio**: Update `core-audio-rs` dependency to 0.14.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Migrated to Rust 2024.
+- Bump minimum Windows version to 10 (since 0.17.2).
 - Bump `audio_thread_priority` dependency to 0.36.
 - `DeviceTrait` and `StreamTrait` now require `Send + Sync` as supertrait bounds.
 - `StreamTrait::play` is renamed to `start`.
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -339,6 +339,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 
 - Bump overall MSRV to 1.78.
+- Bumped minimum Windows version to 10.
```
roderickvd: Three points here:
1. We can't retroactively add entries to releases that already shipped. Please move it to "Unreleased".
2. If it already was Windows 10 before, it may be interesting to add that like "... (since x.y.z)" or similar.
3. Nitpick: "Bump" like the line before.

roderickvd (changes requested):
Appreciate the quick fix. Devil is in the details though 😅

CHANGELOG.md
Marco-Farruggio: No worries, the world would fall apart if code wasn't nit-picked properly 😆 

1) removed the entry from the old version (can't retroactively add entries)
2) I moved it to the current `Unreleased` section, and added since 0.17.2
3) Changed wording from "Bumped" -> "Bump" for consistency

CHANGELOG.md
```diff
@@ -27,6 +27,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **ALSA**: Update `alsa` dependency to 0.12.
 - **Linux**: `realtime` can now promote threads without requiring `realtime-dbus`.
 - **PipeWire**: Set `node.rate` property so that `default.clock.allowed-rates` PipeWire config works.
+- Bump minimum Windows version to 10 (since 0.17.2).
```
roderickvd: That's OK but please move it up, before the backend-specific entries start. This one would be good to float up for visibility, right next to the other bump that is already there.

roderickvd (changes requested):

CHANGELOG.md
Marco-Farruggio: No worries, I actually wanted to do that, but I didn’t incase it was in order of timeline, anyways should be all good now. I put it above the dep because I thought importance wise that made sense :)

PR #1367, opened 2026-09-12
Add fallback path for Pipewire

### LINKED ISSUE #1365: Fails to detect Pipewire without environment variables.
`cpal::platform::available_hosts()` does not list `pipewire` even if `/run/pipewire/pipewire-0` is available.

Workaround: `PIPEWIRE_RUNTIME_DIR=/run/pipewire`.

Other tools like `pw-cli` use `/run/pipewire/pipewire-0` as a fallback socket path, so CPAL should probably do that as well.

Marco-Farruggio:
Closes #1365

if you want me to add this to the changelog, and add a comment explaining why the fallback exists in the code, please do let me know, but I won't be able to touch this again until morning as sleep calls my name

### DIFF
--- a/src/host/pipewire/utils.rs
+++ b/src/host/pipewire/utils.rs
@@ -68,6 +68,10 @@ pub fn find_socket_path() -> Option<&'static PathBuf> {
                 }
             }
 
+            if let Some(p) = socket_in(Path::new("/run/pipewire")) {
+                return Some(p);
+            }
+
             None
         })
         .as_ref()


Marco-Farruggio:
Note: before this is merged I’d probably actually like to add  to the documentation somewhere about this

roderickvd:
Ah sorry :face_with_diagonal_mouth: had not seen this before I pushed the slightly simpler #1371.
I'll add you as co-author there.

PR #1368, opened 2026-09-13
feat(wasapi): add opt-in exclusive output support

teancom:
Hello! I'm one of the maintainers of the [sendspin-rs](https://github.com/Sendspin/sendspin-rs) library, an audio library used by the [Music Assistant desktop client](https://github.com/music-assistant/desktop-app) across Windows, Linux, and macOS.  We've been getting [requests](https://github.com/music-assistant/desktop-app/issues/167) to support exclusive mode and I saw that there had been [previous attempts ](https://github.com/RustAudio/cpal/pull/843)that were acceptable in theory, just needed to get through the PR process.

Anyway, this is that.  I created branches in sendspin-rs and the MA desktop client to test it end to end, using a Windows ARM VM and a Schiit USB DAC. If this approach is acceptable I'll follow-up with another PR for CoreAudio 'hog mode'. I haven't even looked at Linux yet but I presume there's something similar there, and I can do that as well. Also, this initial PR is purely output not input, which I can also follow-up with if this approach is acceptable for you folks.


teancom:
Happy to wait on it until the API is firmed up. And happy to help out where I can. We rely heavily on you folks and are grateful for the work you've done / are doing so we can concentrate on the other bits. ❤️

PR #1369, opened 2026-09-14
Fix null host

Marco-Farruggio:
Not sure if we will delete null host for now, but until then this patches up some things :)


src/platform/mod.rs
```diff
@@ -1008,19 +1004,28 @@ mod platform_impl {
...
+        #[cfg(feature = "custom")] Custom => CustomHost,
     );
 
     /// The default host for the current compilation target platform.
+    #[cfg(feature = "custom")]
+    pub fn default_host() -> Host {
+        CustomHost::new()
+            .expect("the default host should always be available")
+            .into()
+    }
```
LastExceed, line 1047: defaulting to custom host seems counter intuitive to me

src/host/null/mod.rs
```diff
@@ -121,23 +121,23 @@ impl HostTrait for Host {
...
     }
 
     fn now(&self) -> StreamInstant {
-        unimplemented!()
+        StreamInstant::ZERO
     }
 
     fn buffer_size(&self) -> Result<FrameCount, Error> {
-        unimplemented!()
+        Error::with_message(DeviceNotAvailable, "Null host in use")
```
LastExceed: How about adding a private field to `Device` and `Stream` so it is impossible to get an instance of them? then the functions could be left unimplemented

LastExceed (commented):
Quick glance, static only. Although I have a strong suspicion that this is all getting deleted anyway 😁

src/platform/mod.rs
Marco-Farruggio: this is only because no other hosts exist there, so its either custom or a panic

LastExceed: Hmm ok if it returned a result I would actually be inclined to an error, but panic is no-go of course

Marco-Farruggio:
I may close this PR, and make a new one deleting Null host, and making ::new() part of host trait

PR #1372, opened 2026-09-19
Add a pre-commit hook mirroring CI's fmt and clippy checks

ErwanLegrand:
It's all in the title. The aim is to increase the quality of submissions before CI even runs, obviously.


.githooks/pre-commit
```diff
@@ -0,0 +1,18 @@
+#!/bin/sh
+# Pre-commit hook: runs the same fmt and clippy checks as CI's Code Quality workflow
+# (`cargo fmt --all -- --check` and `cargo clippy --all -- -D warnings`), adapted to the
+# local platform. CI additionally lints each supported target with per-platform feature
+# sets (e.g. `--all-features` with the ASIO SDK on Windows), which a local hook cannot.
```
LastExceed: I'd remove this paragraph, as it seems a bit redundant. It does not matter here what CI does, and the hook itself seems simple enough to not require summary.

.githooks/pre-commit
```diff
@@ -0,0 +1,18 @@
...
+#
+# Install once per clone with:  git config core.hooksPath .githooks
+# Skip for a single commit with: git commit --no-verify
+
+set -e
+
+# CI skips documentation-only changes (paths-ignore: "**.md", "docs/**", "LICENSE*", ".gitignore").
+if ! git diff --cached --name-only --diff-filter=ACMR | grep -Eq '\.rs$|Cargo\.(toml|lock)$'; then
+    exit 0
+fi
```
LastExceed: This comment implies a black-list, but the filter actually uses a white-list, so listing all the excluded files is redunant (and prone to be incomplete).

I also think that the filter is redundant in general, since there is no harm in running `fmt` and `clippy` on an unchanged code base, and who knows what kind of files not fitting the filter `build.rs` might depend on in the future.

CONTRIBUTING.md
```diff
@@ -28,3 +28,15 @@ We place no restrictions on use of this codebase as AI training data beyond what
 ### Enforcement
 
 Enforcement actions may be taken at the sole discretion of the maintainers, immediately and without prior notice. Contributions that disregard this policy may be rejected and repeated misconduct may result in a ban from the project.
+
+## Development checks
+
+CI's Code Quality workflow runs `cargo fmt --all -- --check` and `cargo clippy --all -- -D warnings`
+(for each supported target/feature set). A pre-commit hook running the same commands on the local
+platform is provided: install it once with
```
LastExceed: Same as before, I'd remove the CI part.

Also, perhaps emphasize that this is just an optional tool, not a requirement. Keeping all commits 100% compliant is not always feasible, or even desirable, and having to explicitly use `--no-verify` might constitute unnecessary friction for some people, especially those who prefer to do cleanup at the end of each branch instead of each commit (which is a common workflow that IMO is perfectly fine).

LastExceed (changes requested):
I like this idea! There are a few things I'd change though:

CONTRIBUTING.md
ErwanLegrand: Will do!

ErwanLegrand:
I have made the changes as requested. Let me know if you find anything else.

PR #1373, opened 2026-09-19
Add AGENTS.md and CLAUDE.md

ErwanLegrand:
This adds minimal directives for agents and assumes #1372 is merged beforehand. Once again, the goal is to increase the quality of contributions to the repo.

CLAUDE.md is required since Claude Code is widely used and it does not read AGENTS.md unless told to do so.


ErwanLegrand:
As a side note, adding issue and PR templates to the repo could help structure the behavior of agents, too.

https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/about-issue-and-pull-request-templates

It could also make sense to check commit messages in git hooks and GitHub workflows.

I can look into this if it helps.

LastExceed:
I gotta be honest, this raises an eyebrow for me. None of these instructions are even remotely project-specific, so I feel like they do not belong here. Also, most of them seem blindingly obvious, to the point that I'd question why they need stating *at all*.

Our policy on AI contributions is that they are held to the exact same standards as any other contribution, and if someone needs to be told that they should "Always plan before making non-trivial changes", then I am really not sure if they are ready for significant open source projects, human or otherwise.

ErwanLegrand:
While I agree to some extent, this is the reality of mainstream coding agents as they ship nowadays. They know how to write code, but they know nothing about a sane software engineering / quality process. So, the sad truth is someone as to tell them how to make a decent job.
If low quality agent-generated PRs are not a problem at this point, what I suggest is to keep this on the side as a draft or close it. This is easy to restart if/when the problem arises.

LastExceed:
> the sad truth is someone as to tell them how to make a decent job.

Did you A/B test this? I'd be interested in seeing the results :) 

If it really was that easy to magically improve the performance of an LLM, then surely the vendors would have added these instructions to the built-in pre-prompts by now (right next to the paramount "make no mistakes").

> If low quality agent-generated PRs are not a problem at this point, what I suggest is to keep this on the side as a draft or close it. This is easy to restart if/when the problem arises.

Do you genuinely believe that simply adding these files in their current state would make any discernible difference in that regard? To me it seems more likely that it would only invite ***more*** vibe coders, if anything.

roderickvd:
I am going to close this one because:

1. What @LastExceed said.
2. Overall the RustAudio community is not in favor of LLM-usage, when these files signal the contrary.
3. Even an `/init` with Claude Code would generate a more useful framework.

So far we've not been hit by AI slop and I hope to keep it that way. All this coming from someone who is much more pro LLM's than most of the RustAudio community.

Issue #1370, opened 2026-09-15
ASIO backend panics with "attempt to add with overflow" in TimeBase::to_stream_instant after a few minutes (debug builds), small buffer size


ZackPedretti:
## Summary
Using the ASIO host with a small fixed buffer size (16 frames), the ASIO
input/output stream panics after a few minutes of continuous playback with
`attempt to add with overflow` inside `TimeBase::to_stream_instant`
(`src/host/asio/stream.rs:44`). Only reproduces in debug builds (overflow
checks enabled); release builds presumably wrap silently instead.

## Environment
- cpal: 0.18.2
- OS: Windows 11 Professional (10.0.26200 Build 26200)
- Interface: Focusrite USB ASIO (Focusrite Scarlett 2i2 3rd Gen)
- Rust toolchain: (`rustc 1.98.1 (48a229cea 2026-09-01)`)
- Profile: debug (`cargo run`)

Minimal repro, using a Focusrite interface over ASIO with a 16-frame fixed
buffer, 44.1kHz, 2ch, i32, simply copying input channel 0 to both output
channels via a ring buffer:

```rust
use std::time::Duration;
use anyhow::anyhow;
use cpal::{BufferSize, StreamConfig};
use cpal::traits::{DeviceTrait, HostTrait, StreamTrait};
use ringbuf::consumer::Consumer;
use ringbuf::HeapRb;
use ringbuf::producer::Producer;
use ringbuf::traits::Split;

fn main() -> anyhow::Result<()> {
    let host = cpal::host_from_id(cpal::HostId::Asio)?;
    let devices = host.devices()?;

    let focusrite = devices
        .filter_map(|device| {
            let description = device.description().ok()?;

            if description.to_string().contains("Focusrite USB ASIO") {
                Some(device)
            } else {
                None
            }
        })
        .next()
        .ok_or_else(|| anyhow!("Focusrite USB ASIO not found"))?;

    println!("Using device: {}", focusrite.description()?);

    let config = StreamConfig {
        channels: 2,
        sample_rate: 44_100,
        buffer_size: BufferSize::Fixed(16),
    };

    println!("Config: {config:?}");

    let ring = HeapRb::<i32>::new(16_384);

    let (mut producer, mut consumer) = ring.split();

    let input_stream = focusrite.build_input_stream(
        config,
        move |data: &[i32], _info| {
            for frame in data.chunks(2) {
                let input_1 = frame[0];
                let _ = producer.try_push(input_1);
            }
[...]

Issue #1356, opened 2026-09-07
[Windows+ASIO] Can't separately build input stream if output stream is already running


nik-r-at:
Context: Windows + ASIO
CPAL version: latest (commit id 4d61a25ff69a147eb7ba1194a8a783ba4273db71)
Tested with: Focusrite USB Asio and Yamaha/Steinberg USB ASIO

See below the following modification of the `beep` example:

- the output stream is started in a thread,
- then an input is started on the main thread 

I assume this should be a valid way to do things? Either way it fails with the following error:
```
     Running `target\debug\examples\beep.exe --device 'asio:Yamaha Steinberg USB ASIO'`
Output device: asio:Yamaha Steinberg USB ASIO
Default output config: SupportedStreamConfig { channels: 2, sample_rate: 48000, buffer_size: Range { min: 256, max: 256 }, sample_format: I32 }

thread 'main' (11312) panicked at examples\beep.rs:72:6:
failed to find input device
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
error: process didn't exit successfully: `target\debug\examples\beep.exe --device 'asio:Yamaha Steinberg USB ASIO'` (exit code: 101)
```

When digging around in asio-sys, I found that the comment states that ASIO only supports loading a single driver at a time. IF that driver is loaded, it should be returned in line 456.

But it seems like that's not happening, as seemingly the `LoadDriverError::DriverAlreadyExists` error is thrown. Changing that into `LoadDriverError::LoadDriverFailed` allows to run the example.

**MODIFIED BEEP EXAMPLE**
```
//! Plays a simple 440 Hz sine wave (beep) tone.
//!
//! This example demonstrates:
//! - Selecting audio hosts (with optional JACK support on Linux)
//! - Selecting devices by ID or using the default output device
//! - Querying the default output configuration
//! - Building and running an output stream with typed samples
//! - Generating audio data in the stream callback
//!
//! Run with: `cargo run --example beep`
//! With JACK (Linux): `cargo run --example beep --features jack -- --jack`
//! With specific device: `cargo run --example beep -- --device "wasapi:device_id"`

use clap::Parser;
use cpal::{
    CallbackInfo, Device, Error, ErrorKind, FromSample, HostId, I24, Sample, SampleFormat,
    SizedSample, StreamConfig,
    traits::{DeviceTrait, HostTrait, StreamTrait},
};

#[derive(Parser, Debug)]
[...]

PR #1374, opened 2026-09-19
fix(asio): paused streams, draining timeouts and in-use drivers

roderickvd:
@LastExceed could you review?

From a code review I am particular taken by the `pause()` mechanism. Did that never work before or what am I missing?

Fixes:
- #1356
- #1370


LastExceed:
> @LastExceed could you review?

Will do

CHANGELOG.md
```diff
@@ -49,6 +49,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **ASIO**: Fix a deadlock when dropping a `Stream` that owns another ASIO `Stream`.
 - **ASIO**: Fix loading a driver while a previous driver was still unloading.
 - **ASIO**: `Stream` no longer risks blocking or panicking in the driver callback while another stream is being created or destroyed.
+- **ASIO**: A paused output stream now goes silent instead of looping the last buffered audio.
```
LastExceed, line 52: > Did that never work before or what am I missing?

Apparently not, I just tested the latest `master`, and pausing is indeed broken there 😰 but your fix works fine.

CHANGELOG.md
```diff
@@ -49,6 +49,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
+- **ASIO**: Building a stream while another driver is loaded now reports `DeviceBusy` instead of `DeviceNotAvailable`.
```
LastExceed: Not sure about this one. If I understand correctly, then the situation is basically:

> this device is not available

because

> another device is busy

As a user, I'd intuitively assume the error to be about the device that I am accessing, not about another which happens to be in the way. In that mindset, `DeviceBusy` is misleading. Knowing the inside details, I understand that `DeviceBusy` better expresses the *cause* of the problem, but from an outside perspective, `DeviceNotAvailable` is more representative of the resulting situation.

LastExceed (commented):
Just 1 semantic technicality, rest LGTM

CHANGELOG.md
roderickvd: That is an interesting point, you could say that `DeviceBusy` is a specialisation of `DeviceNotAvailable`. But, from `src/error.rs`:

```rust
    /// The device is temporarily busy. This can happen when another application or stream
    /// is using the device. Retrying after a short delay may succeed.
    DeviceBusy,

    /// The requested audio device is not available.
    ///
    /// This can happen if the device has been disconnected while the program is running, or if
    /// the device identifier refers to a device that does not exist on this system.
    DeviceNotAvailable,
```

`DeviceBusy` seems to closer match the definition, wouldn't you agree?

LastExceed: No. The doc comment says "The device" twice. But *which* device? As a user I would naturally expect it to be the one I am trying to use. I might not even be aware that there is an alternative at all.

roderickvd: Hmm, you're right. I guess doing that still addresses #1356 sufficiently by having clarified the `asio-sys` error message.

Issue #1376, opened 2026-09-20
AUDCLNT_BUFFERFLAGS_SILENT not handled


ErwanLegrand:
Per Microsoft's specifications, it is the applications responsibility to zero the buffer when a driver returns AUDCLNT_BUFFERFLAGS_SILENT.

https://learn.microsoft.com/en-us/windows/win32/api/audioclient/ne-audioclient-_audclnt_bufferflags

I have a hardening PR for WASAPI in preparation which fixes this issue amongst other things.

roderickvd:
Good one. Reading the docs it actually sounds like we should just ignore the buffer and return all silence (which is slightly different than zeroing the buffer):

>  Treat all of the data in the packet as silence and ignore the actual data values

roderickvd:
I had several local WASAPI changes cooking, to which I added this: #1379

PR #1379, opened 2026-09-20
fix(wasapi): draining stops and silent capture buffers

roderickvd:
Following the discussion at #1357 I had a closer look at WASAPI and did see that stops are at risk of draining short. Though draining stops will never be perfect on WASAPI, this at least improves that.

While at it I also removed dead 64-bit format conversions, and quickly addressed #1376.


roderickvd:
@LastExceed

PR #1382, opened 2026-09-20
Fix/all targets

### LINKED ISSUE #1375: Hole in CI checks?
The clippy job that denies warnings runs only `--all-features` and without `--all-targets`.

So a lint-level regression that appears only under default features slips through and clippy lints in test and example bodies aren't checked.

I am not sure whether this is intended or not?

ErwanLegrand:
Fix CI so that clippy runs with all features enabled and with just the default set. Also, add `--all-targets`.

Some issues were found while running this. This PR also fixes these.

Fixes #1375


ErwanLegrand:
Let me fix those conflicts...

ErwanLegrand:
What do you think about checking both the default feature set and all features? This adds three additional jobs, but it catches issues which aren't caught otherwise.

roderickvd:
> What do you think about checking both the default feature set and all features? This adds three additional jobs, but it catches issues which aren't caught otherwise.

Yes, that's what we should normally do, ensuring we don't do anything double when we don't mean to.
Weren't we already doing that, where it made sense, and CI system libraries allow us?

ErwanLegrand:
As I remember, CI was running clippy only with all features enabled, never with the default set

I noticed that when working on a branch to add support for WASAPI exclusive mode, a while back. (I need this for realtime effects processing.)

CI was green, but just running `cargo clippy` returned a dead code error.

Possibly, this is the only class of bugs CI misses by not running clippy with the default feature set?

roderickvd:
I agree clippy should be OK for both default features and all features.

ErwanLegrand:
OK. This PR is an attempt at checking exactly this.

PR #1384, opened 2026-09-20
WASAPI hardening / fixing

### LINKED ISSUE #1376: AUDCLNT_BUFFERFLAGS_SILENT not handled

ErwanLegrand:
## Summary

The WASAPI backend trusts values that come from outside cpal: the buffer size `IAudioClient::GetBufferSize()` reports, the padding count `GetCurrentPadding()` reports, the frame count `IAudioCaptureClient::GetBuffer()` writes, the `cbSize` a driver declares for a format, and the flag bits on a capture packet. It also read a `WAVEFORMATEXTENSIBLE` through a pointer derived from the 18-byte `WAVEFORMATEX` prefix while handing the driver `cbSize = 22`, and the `GetBuffer`/`ReleaseBuffer` pairing on both the render and the capture path could be left open on an error return.

A misreporting or misbehaving driver turns them into a panic on the audio worker thread (which cpal cannot report to the user), an out-of-bounds read, a multi-gigabyte allocation, or a capture buffer served with stale contents. Two of them need no  misbehaviour at all: a wide interface (44+ channels), or a user-chosen 8-byte sample format at 22 channels, overflows the `WAVEFORMATEX` fields derived from the config and the probe path walks sample rates up to 24.576 MHz, so it overflows there first.

Every fix is a bound, a rejection, or a pairing fix at the point the untrusted value enters. No public API changes; the user-facing summary is in `CHANGELOG.md`.

### `src/host/wasapi/device.rs`

- `cbSize` checked before the extension is read.
- Overflow when deriving `WAVEFORMATEX`.
- Full-`WAVEFORMATEXTENSIBLE` provenance.
- `GetBufferSize()` bounded.

### `src/host/wasapi/stream.rs`

- Padding larger than the output buffer.
- Capture packet larger than the buffer.
- Empty and silent packets.
- Drain loop bounded per audio event.

### COM pairing

- Render packet released when the timestamp query fails.
- Capture packet released when processing fails.

### Docs

- The `DeviceTrait` timeout contract is now summarized in the trait docs, each host module documents what its `timeout` bounds and which backends ignore it, and the WASAPI activation path is annotated at the unbounded `IMMDevice::Activate` call.

### cfg gate

- The `frames_to_duration` cfg gate admitted all of Windows and any build with the `audioworklet` feature; on Windows only the ASIO and JACK backends call it.

## Tests

- `src/host/wasapi/device.rs`: pins the `checked_mul` overflow guard and the PCM/extensible field layout (only reachable through real misconfiguration) plus the values the i24 and PCM paths hand to the engine.
- `src/host/equilibrium.rs`: pins the per-format byte patterns `fill_equilibrium` writes (`0x80` for `U8`, zeros for signed/float/i24, `0x69` for DSD) and the typed path (`U16`/`U24`/`U32`/`U64`), the only one that casts the buffer to a wider pointer, including that its `chunks_exact` accounts for every byte.

## Verification

**Automated.** `cargo fmt --check`; `cargo clippy --all-targets --target x86_64-pc-windows-msvc -- -D warnings`;
`cargo clippy --all-targets --target i686-pc-windows-msvc -- -D warnings`; `cargo test --lib` (17 passed). Clean on Windows 11, MSVC toolchain linked with LLVM lld 23.1.1. Every commit in the series also builds with `cargo check --all-targets` for `x86_64-pc-windows-msvc`.

**Hardware, Windows 11 (build 26200), PreSonus AudioBox 22VSL and a VB-Audio Virtual Cable endpoint, 48 kHz,
a 440 Hz tone piped into it.** Capture was exercised end to end through the virtual and the native 24-bit
interfaces.

**Branches no healthy driver reaches**, forced under `lldb` on a live `record_wav` capture stream. 

- **`AUDCLNT_BUFFERFLAGS_SILENT` flag** forced before the branch.
- **Oversized packet** (`frames_available = 1057`, one past the buffer): the guard returned its
  `BackendError` to the user's error callback, and the packet was released.
- **I24 shift**: the callback was served the  scratch, and all eight sampled `i32` values equal `engine_i32 >> 8` exactly.

I used coding agents extensively while working on this. I don't expect regressions on other architectures, but I have not ran tests apart from the ones I ran on Windows 11.

Fixes #1376


roderickvd:
@LastExceed

LastExceed:
Pure AI slop. Not gonna waste my time writing an elaborate review

ErwanLegrand:
Would it help if I contributed fuzzers?

LastExceed:
Depends on what you plan to fuzz exactly, and how you'd go about evaluating the outcome.

roderickvd:
My initial thinking is that we need less, not more. I'm gonna go to bed now, but from a cursory look I was also like "when could this happen / which actual problem is this solving?"

ErwanLegrand:
OK. Let's try to structure this a little bit!

I'm thinking that, since this is a Rust project, some of what I have made could possibly be better addressed by using the type system to enforce correctness by construction. This is clearly the right direction from an engineering perspective, with the caveat that it will break API compatibility. I'll give it a little more thought and probably open issues, assuming this bears fruits, rather than make PRs.

Now that this is out of the way, do you agree that input validation should be performed at trust boundaries and in front of unsafe code? If you disagree with this, 90% of this PR is going to look useless to you.

When the above two points are handled, what remains is the question of runtime checks in APIs which can't be hardened using the Rust type system and do not sit either on a trust boundary or directly in front of unsafe Rust code. I guess that's about 20-30% of this PR and such checks can be seen as redundant, provided there is a clear API contract.

(I wish the Rust compiler checked operations on numerals, when possible. When it comes to numerals, Rust code is vulnerable to the same classes of issues as C code: integer overflows, unsafe casts and the like.)

Would the following look like a good plan to you?

I can open issues to track places where correctness by construction can be used to harden APIs according to my assessment, I can open issues to track where a clear API contract needs to be defined and then update this PR to keep only checks in front of unsafe code and on the driver-app boundary.

LastExceed:
> some of what I have made could possibly be better addressed by using the type system to enforce correctness by construction

Or just not at all, since it is comically overkill.

> do you agree that input validation should be performed at trust boundaries and in front of unsafe code?

Certainly not to the degree you seem to have in mind.

> I wish the Rust compiler checked operations on numerals, when possible

It literally does.

> Would the following look like a good plan to you?

I'm a bit sceptical, but will give you the benefit of doubt

ErwanLegrand:
> Or just not at all, since it is comically overkill.

Well, I've been working in software engineering and security engineering for close to three decades. This might explain a difference in perspective. Anyway, if you're not interested in what I have to offer, I'm not going to annoy you any longer.

> > I wish the Rust compiler checked operations on numerals, when possible
> 
> It literally does.

Are you kidding me? Debug builds have **runtime** checks. Production builds have no checks by default. `checked_add` and friends exist for a reason. `as` is never checked. Unless the developer explicitly require checks, there are none in production builds. This is no better than C.

LastExceed:
As you emphasized yourself, those are runtime checks, not compiler checks, so not sure why you're bringing this up now? I was referring to stuff like [`rustc(E0080)`](https://doc.rust-lang.org/error_codes/E0080.html) or [`rustc(arithmetic_overflow)`](https://doc.rust-lang.org/rustc/lints/listing/deny-by-default.html#arithmetic-overflow)

ErwanLegrand:
I misunderstood you. Sorry for that!

I wish Rust had ranged integers like Ada / SPARK. I wrote some code in Ada for Air Traffic Management applications near the end of the previous millennium. This isn't relevant to the current discussion, though.

Anyway, you might be right in thinking that audio processing does not require the same amount of input validation as other types of software and this is your project, not mine.

Let me know if there's anything in that area which could be useful in your eyes. If that isn't the case, or this just isn't the right time, let's just close this PR.

Issue #571, opened 2021-05-05
Support wasi?


YXL76:
Failed to run`wasmtime beep.wasm`. It shows

```
thread 'main' panicked at 'failed to find output device', examples/beep.rs:79:6
```

SeanOMik:
I get the same error

borsboom:
Is it possible at this time?  I've just been looking into it and it doesn't seem like there is any WASI API that supports audio yet.

PR #1386, opened 2026-09-21
Remove Null host

### LINKED ISSUE #1366: Null host shouldn't exist when a Custom host is present
Correct me if I'm wrong, but the whole idea of the Null host is that it should only exist if there are no other hosts for that compilation, so that each enum/type etc has atleast 1 member, the `NullHost`, however, the `Null` host can currently compile in, even if the Custom host is used, on a compilation target which cpal doesnt have a backend for, this seems counter-intuitive to me, and infact, in my mind, compiling for a target without any of cpals backends and then using a custom host to implement it themselves, seem to go hand in hand, this also could be quite an easy fix (I believe). Also, this means that it will always set the default host to `Null`, instead of `Custom`

This *seems* like a logical error to me, and a quick PR I can whip up.

Marco-Farruggio:
Remove the null host, add a compilation error on unsupported targets, and updated the documentation. To me expecting cpal to compile on a target with no audio capabilities is like expecting a GUI library to compile on a target with no display server, which should really be a build error. This also means its less clutter in the codebase which is always a plus.

Closes #1366


Marco-Farruggio:
Related: I think #571 can be closed because WASI doesn't support audio, and the proposed `wasi-media` isn't near starting

